PR #2, opened 2015-01-19
Feature/adempiere 257 update seed databases

mckayERP:
This pull request merges 380 with Feature/ADEMPIERE-257.  The important commit for this pull request is the last one which provides a possible solution to keeping Garden World up to date using a process rather than SQL.  I've added the process to the Migration Loader, like the role access update, so it will be run when the seed is generated.  The process is repeatable so users can run it as often as they like.  I'm providing it as a pull request for further testing and discussion.


e-Evolution:
Hi Mike I saw a lot of change only for 4 files I think this branch what create form a old line history, so I think that better that include in hot fix

PR #7, opened 2015-01-30
Fixing Local File Inclusion vulnerability

benben159:
Local File Inclusion vulnerability detected at the `Trace` variable on `/adempiereMonitor` page. I found it by misusing the variable by changing the value of `Trace` variable to any existing file outside the Adempiere home folder.
N.B. sorry for the `rename` and `merge` commit. It occured because I forgot to checkout the `master` branch.


e-Evolution:
hi Benben, please following of adempiere development procedure , you need create a new feature brach using git flow , master only can be commit with a release version. so this way I can adding your contribution.

armenrz:
Hi Benben, 

Thank you for your great findings!
Rather than to allow the whole Adempiere home, it is better to limit it to log folder only. Remember AdempiereEnv.properties contains your password unencrypted.

Armen

mckayERP:
Hi Benben,

I'm having issues with security and am receiving an error:  

05:49:31.243 Ini.loadProperties: C:\Adempiere\adempiere.properties #37 [1]
05:49:31.243 Secure.decrypt: Failed decrypting javax.crypto.BadPaddingException: Given final block not properly padded [1]

This is preventing me from connecting to the server.  I also can't work in eclipse without resetting the adempiere.properties file.  Could this be related to the changes in this pull request?

Thanks,

Mike

e-Evolution:
Hi Mike!

I can replicate of error, but it is caused because you are using Adempiere.properties old ,  Benben, changed of encrypt method from DES to AES

So to solve of issue only need delete your old Adempiere.properties

kind regards
Victor

benben159:
actually I haven't finish the fix. I'll continue the fix, but can I submit the fix under `develop` branch?

PR #12, opened 2015-03-07
Scroll Bar had been added to the process dialog against my ticket

Mbilalilyas:
ADEMPIERE-407


e-Evolution:
Hi we include your contribution from Mike pull request , please when you want send a pull request need send of a hotfix branch , the master branch only can be affect when we make a release.

you can review here of development procedure:

http://www.adempiere.com/Software_Development_Procedure

kind regards

trifonnt:
I would like to know more why such change is needed as ADempiere currently shows horizontal and vertical scroll bars when process parameter window has many parameters.

e-Evolution:
This was reported by @yamelsenih confirm that this problem was solve when migrate to MVC

PR #14, opened 2015-04-29
Adempiere 387 spin suite contribution

yamelsenih:
Add Spin Suite Contribution


e-Evolution:
Hola Yamel!

Can please create again of request pull using of ADEMPIERE-387-Spin-suite-Contribution branch , because master only should be affect by official releases.

kind regards

PR #19, opened 2015-04-30
Hotfix/3.8.0#002 adempiere 411 material receipt uom

mckayERP:
ADEMPIERE-411 - the UoM used on material receipts is incorrect.  It should be the UoM of the product but it uses the UoM entered on the document, which could be different.

The relevant changes are 4c3fcfc & 5a96707


mckayERP:
To clarify - I mean the UoM in the accounting facts generated by the MR.

mckayERP:
Based on discussions with the group, the fact UOM can follow the document UOM as long as the quantity is in the product UOM.  No change is required so this pull request can be closed.

PR #52, opened 2015-10-15
Fixed VButton to hide non active records.

trifonnt:
Fixed VButton to hide non active records.

P.S.
Complexity and branches kill productivity!


e-Evolution:
yes, but is necessary to create of release patches , I not found way to get , but I accept suggestion  :-)

PR #53, opened 2015-10-26
BF: Pedro's Financial Report is not working

armenrz:
Pedro's Financial Report should work in WEB UI. This patch make ProcessUtil.startJavaProcess to redirect to FinReport web version when using WEB UI


e-Evolution:
Dear Armen , Thank a lot for your request poll.

Based on development process http://www.adempiere.com/Software_Development_Procedure

of branch master only can be change from a release branch , so please would send again request poll from hotfix/3.8.0#002 branch , is here where we are collect all patch for next version.

Of idea is that all patches need be push into hotfix/3.8.0#002, this way I can collect all patches and create for old version patches.jar, so this reason because i need that pull request need come from hotfix/3.8.0#002 branch, here the procedure how to :

1.- make to backup next command cp adempiereSource adempiere.bak
2.- cd adempiereSource 
3.- git pull
4.- git checkout hotfix/3.8.0#002
5.- set here modify java files 
6.- git commit -m ' BF: Pedro's Financial Report is not working #53'
7.- git push 
8.- finality you can how create pull request for this commit


thank a lot

armenrz:
Hola Victor,
Thanks for directing me to the procedure. I will try to follow it this weekend.

Rgds

PR #54, opened 2015-10-29
Change the logic of matched to be within standard precision

mckayERP:
ADEMPIERE-450 Change the logic of matched to be within the standard precision of the product.  If the invoiced, PO and receipt quantities are decimals, the match of quantities should be within the limit of the standard precision.  It is hard to get decimal representations to match exactly.


mckayERP:
Base branch changed.  Will resubmit.

PR #76, opened 2015-11-03
webui statusbar message is fixed

syvasoft:
The message returned from Process will be shown in webui statusbar just like swing client.
Ref: http://adempiere.net/forums#!/thread/42516


e-Evolution:
Thank a lot syvasoft

syvasoft:
Welcome Victor. 
It's gratifying to see that my contribution has been accepted.

PR #77, opened 2015-11-03
Feature/admpiere 444 windows install scripts

mckayERP:
This adds the windows installer components to the main repository and makes a few necessary changes in the supporting utilities to make the installation simple and smooth.  Tested now over 300 times.  Suggest adding to 380LTS hotfix 2.


mckayERP:
Closed based on discussion to not include in core.  It will be added as a separate project.

Issue #102, opened 2015-11-18
The multiple selection not are work in ZK for select multiples items


e-Evolution:
In old version cloud  select multiples record to include multiples order line.

1.- Create a sales order
2.- Go sales order line , and click in product field
3.- Select multiples product lined and click ok
4.- The multiples line not is created

mckayERP:
Fixed.  See pull request #104.  WSearchEditor was not sending the array of values to the event listener in GridTabDataBinder where the multiple records are generated.

mckayERP:
Fixed with merge #104.

PR #104, opened 2015-11-18
Fix the multi-selection in WSearchEditor to match VLookup.  

mckayERP:
The array of values returned by the info windows needs to be passed to the event listener.  Only the first value was being returned in WSearchEditor.  Some code cleanup in both files to test for changes to the field where the possible array was tested rather than the single value.


mckayERP:
#102 fixed.

PR #120, opened 2015-11-24
Update 882_HR002_FR3015725_HRPaymentSelectionCheck.sql

ayazbaloch:
Fixed


e-Evolution:
please if you want send a pull request it should are into the current main branch in this case release/3.9.0 or before hotfix/3.8.0#002, please if you want including your fix , only send again from a fresh or last commits of origin branch

Issue #125, opened 2015-11-24
Set flag "isInvoiced" in an Order when an invoice has been completed


marcalwestf:
When in a Sales Order, the corresponding invoice has been completed, the  flag "isInvoiced" has to be set. 
Now this is not the case.

e-Evolution:
Hi Mario

This flag not is use current for system , but would be useful for indicate when sales order was invoiced , my recommendation is that implement this logic into completeIt method of Invoice this way when an invoice is completed validate that Order was invoiced

I not sure that should happen when is invoice parcial or exist an return with credit note , the flag should be set as N

yamelsenih:
Hi @marcalwestf @e-Evolution i think that it flag must be change to true just when the order is completely invoiced.

yamelsenih:
Hi, any update for it?

PR #126, opened 2015-11-24
Hotfix/3.8.0#002

marcalwestf:
Solving issue https://github.com/adempiere/adempiere/issues/125


e-Evolution:
currently Adempiere not implement noting functionality with of flag isInvoiced in sales order , but be useful to report , but I think that need more work to implement when an invoice is revert should be unmarked and not can see here this code.

PR #129, opened 2015-11-25
Feature/admpiere 444 windows install scripts

mckayERP:
The install scripts have been moved to a separate project.  These changes are to core project files to ensure the install process is seamless and silent.  Without these changes, the installer will not function smoothly and user intervention will be required at a number of points in the install process.

The changes are:

ini.java and RUN_Adempiere.bat - change the way the software searches for the adempiere.properties file.  The properties file is created in the ADEMPIERE_HOME directory by the first run of RUN_SilentSetup. The client software looks for the properties file in the user home directory but it doesn't exist, so a new properties file is created using the defaults which don't work.  This change allows the client to look in the ADEMPIERE_HOME directory as well.

RUN_SilentSetup.bat - Allow the batch file to be run from another directory, for example from C:\ as c:\adempiere\RUN_SilentSetup.bat.  Without this change, the .bat has to be called from ADEMPIERE_HOME as the classpath int he bat is relative to the calling directory.

A number of new icons are added along with the source graphic file to support window menu creation.

Build.properties is changed to include release migrations rather than 3.8.0.  The release name is changed to 3.8.0_2

RUN_ImportAdempiere.bat and RUN_MigrateXML were modified to allow for silent operation.

Some changes to improve the use of windows service and to allow the install scripts to install ADempiere as a windows service, start and stop the service and uninstall the service.


mckayERP:
Base branch changed.  Will resubmit.

Issue #134, opened 2015-11-27
Improve transactions in migration process


e-Evolution:
1.- Currently exist duplicate code
2.- Now you can call same process of different way 
3.-The transaction is idle when try load XML filed and apply from process
4.- No exist a good management of transactions , the idea is each step should a transaction and each migration script too.

e-Evolution:
working process ......

mckayERP:
As of 682ab5d44, 1 Dec 2015, the import of xml files from the 370seed is hanging on the first import of file \Adempiere\migration\370lts-380lts\xml\0000_120_Merge_Migration.xml, Step 30.

Its hanging when a column in AD_Migration is synced.  I think, column additions/changes that need to be synced with the database require exclusive locks of the table for ALTER TABLE statements.  So for columns in AD_Migration, these have to be done when the AD_Migration Table has no locks.  For Columns in AD_MigrationStep - when AD_MigrationStep has no locks.  For other tables, the column sync can be done as part of the step apply/rollback.

The function getSteps() in MMigration 

			for ( MMigrationStep step : getSteps(false) )
			{

locks the AD_MigrationStep table.  There is a similar lock in the AD_Migration table but I haven't found the specific line yet.

So a migration can't add and sync columns in the AD_Migration table or AD_MigrationStep table unless there is a way to release the locks on these tables.  For ease of use, we need to find a way to allow changes to these tables as part of the migration loading.

mckayERP:
This seems to fix it:

In MMigration.java, line 236

 //       MMigration mmigration = new Query(ctx, MMigration.Table_Name, where, trxName) // Locks migrationtable
        MMigration mmigration = new Query(ctx, MMigration.Table_Name, where, null)

In MigrationApply.java line 65

            public void run(String trxName) {
 //               MMigration migration = new MMigration(ctx ,migrationId , trxName);  // Locks migration table
                MMigration migration = new MMigration(ctx ,migrationId , null);        // Doesn't lock table 

Not sure yet how to allow changes to the migration record within the main transaction - for example, when updating the "applied" status.

e-Evolution:
Hi Mike I review of code and saw that only execute an query to know if this migration exist 

MMigration migration = new Query(ctx, MMigration.Table_Name, where, trxName)
		.setParameters(name, Integer.parseInt(seqNo), entityType, releaseNo).firstOnly();
		if ( migration != null && migration.getAD_Migration_ID() > 0) {
[...]

PR #168, opened 2015-12-11
Hotfix/3.8.0#002 refactor migration processes

mckayERP:
#134 Refactor the migration processes to use the process mechanisms and to improve the error reporting.  Tested for all xml migrations from 370 to 380#002 using RUN_MigrationXML, the Load migraitons from XML process in the gui, individual migration apply/rollback and step apply/rollback with and without errors introduced.

### DIFF
--- a/base/src/org/adempiere/process/MigrationLoader.java
+++ b/base/src/org/adempiere/process/MigrationLoader.java
@@ -17,6 +17,7 @@ import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 
+import org.adempiere.exceptions.AdempiereException;
 import org.apache.commons.io.FileUtils;
 import org.compiere.Adempiere;
 import org.compiere.model.I_AD_Migration;
@@ -30,7 +31,6 @@ import org.compiere.util.CLogger;
 import org.compiere.util.DB;
 import org.compiere.util.Env;
 import org.compiere.util.Trx;
-import org.compiere.util.TrxRunnable;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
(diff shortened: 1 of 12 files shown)


e-Evolution:
Hi Mike!

I think that this code in class Column, should be remove.


// For backwards compatibility with older XML, skip the validation of the lookup 
		// if a migration script is running
		if (Env.getContext(Env.getCtx(), "MigrationStepApplyInProgress").equals("Y") )
			return;

So if current system fail when an lookup table type is created because of data was incorrect we need fix XML script and not formalize the problem, I think that the error from 360 to 370 or 370 to 380 in reference table type are because of XML not is rightly and we need fix.

Is very simple we need add of value to AD_Reference_Value with right value , of any way the Table definition will fail in runtime.

e-Evolution:
Mike I found a code that not is a good practice  in class MEntityType https://github.com/adempiere/adempiere/blob/708bf740a9bd92fc822a4f8b1f3051ee755a64dc/base/src/org/compiere/model/MEntityType.java#L280


// Allow delete of entities for migration scripts
		if (Env.getContext(getCtx(), "MigrationStepRollbackInProgress").equals("Y") )
		{
			s_entityTypes = null;	//	reset
		return true;
		}
The reason is that you not should implement any code or behavior that try evaluate that are happen from other process , it not is general solution.

Imagine you should finish with a a lot of class with if and Env.getContext(getCtx(), "MigrationStepRollbackInProgress").equals("Y") , I know that work but not is good mix domain model class with external logic.

So I implement other approach https://github.com/adempiere/adempiere/commit/0a78f06eb2ec1cadf20874f0319660b3e10effc2

this way we are implement a general solution that can be useful other case.

e-Evolution:
About of MMigration class I see here a lot of Refactory and I have some objection.

!.- I think that not should deprecate Rollback process , no is good to read the code , now when look of code I not sure that logic is executed , this approach force that review of logic using debug code.

I think that we should separate of action of status, for my an apply method is an action not a status , and should may be create an Unapply for reverse an migration script or rollback, it will let read more easy of code.

Please no use  Boolean apply = false as status and action is very confused.

https://github.com/adempiere/adempiere/blob/708bf740a9bd92fc822a4f8b1f3051ee755a64dc/base/src/org/compiere/model/MMigration.java#L147

https://github.com/adempiere/adempiere/blob/708bf740a9bd92fc822a4f8b1f3051ee755a64dc/base/src/org/compiere/model/MMigration.java#L164

https://github.com/adempiere/adempiere/blob/708bf740a9bd92fc822a4f8b1f3051ee755a64dc/base/src/org/compiere/model/MMigration.java#L195

https://github.com/adempiere/adempiere/blob/708bf740a9bd92fc822a4f8b1f3051ee755a64dc/base/src/org/compiere/model/MMigration.java#L139

Please you implement an Unapply and rollback process is more readable.

Thank a lot

e-Evolution:
About of MigrationScriptBatchInProgress we should are sure that it variable can not overwrite while migration process is executed , because would cause inconsistency.
So I can imagine following case:

1. The Admin execute Migration Loader from terminal and for some reason the admin  execute of migration script from UI , What is behavior "MigrationScriptBatchInProgress" flag? , When is on and off ?  When is execute Migration Loader from terminal and sometime a script from migration window?

I think any progration execution should execute maintenance mode y evitar that the process is execute two or more time while other process is running.

I not sure that happen and more when use the current approach apply as action and status.

Issue #288, opened 2016-04-03
SyncTerminology can create duplicate system elements causing an error


mckayERP:
The SyncTerminology process finds existing centrally-maintained columns and process parameters which do not have system elements and creates the system elements for them.  The search uses the following query:

```
        sql="SELECT DISTINCT ColumnName, Name, Description, Help, EntityType "
            +" FROM AD_PROCESS_PARA p "
            +" WHERE NOT EXISTS "
            +" (SELECT 1 FROM AD_ELEMENT e "
            +" WHERE UPPER(p.ColumnName)=UPPER(e.ColumnName))"
            +" AND p.isCentrallyMaintained = 'Y'"
            +" AND p.isActive = 'Y'";
```

If two columns or process parameters use the same column name but slightly different descriptions, new system elements are created for each, causing an exception on the element save due to duplicate column names.

Solution is to order the query by name and then ignore duplicates of the name.  (The expression SELECT DISTINCT ON (ColumnName) ColumnName ... works in Postgres but is not standard.)

Later in the process, unused system elements are deleted based on columnName but this is done before the column names are updated.  The deletion should be based on system element ids instead.

yamelsenih:
Hi @mckayERP, i think that it process must be dynamic like the **AD_Language_Maintain**, if the tables of translation has a flag **_Trl**, then it can have a parent table for get changes in translation, example:

AD_Element -> AD_Column -> AD_Field
AD_Element -> AD_Process_Para
AD_Window -> AD_Menu

if in the table of translation will exists a column reference to parent table for translation then is possible get it dynamically

mckayERP:
Hi Yamel, Not sure this is the same as the translation as the process is limited to ensuring consistency of the system elements.  I was having this problem with the post migration processes because the system I was working on simply had different descriptions of a column name in the table and process parameters. The Sync Terminology routine would try to create two system elements with the same column name and crashed.

yamelsenih:
I think that in a future, it process will be changed, the query can be make from attributes in the table, now, for add a translation tab in the system, is necessary add to ***SyncTerminology*** a  support to new table.

Best regards

e-Evolution:
Solve in 4071a0c1b97f7a4cd416e069eb3b3bc3269a71d7

PR #289, opened 2016-04-04
#288 Prevent errors caused by creating duplicate system elements

mckayERP:
See issue #288. This correction prevents errors when running the SyncrhonizeTerminology process.


e-Evolution:
Mike , this validation would be a good , but I can see a potencial problem,  currently the AD_Element_ID not had as mandatory, so that if the validation only is using of Ids can be cause strange behavior

Please can check this scenario ?

mckayERP:
Will do.  Thanks.

Mike

e-Evolution:
Hi Mike , I can not merge any pull request that have conflict , please integrate or update of heads last commits and send again , So I can review

Issue #363, opened 2016-05-03
MProcess getUsingJavaClass does not set parameters properly.


mckayERP:
The query is defined but parameters are not set.  Also, the className variable should be set using .getCanonicalName() function directly as getClass().getCononicalName() will always return "java.lang.class".

yamelsenih:
Hi @mckayERP is nice to have a unique class for handle the process run, currently exists ProcessUtil, ProcessCtl, Procesa Builder and it is very hard when you want make a refactoring or add new functionality or validations.

mckayERP:
Agree.  This issue is just that I like being able to use the class name
rather than the process ID which is difficult to find and verify.

So instead of

    ProcessInfo processInfo = ProcessBuilder.create(ctx)
                .process(53173)...

It is much more readable to do this

    ProcessInfo processInfo = ProcessBuilder.create(context)
            .process(org.compiere.process.MigrationFromXML.class)...

Issue #325, opened 2016-04-15
SvrProcess must handle mandatory error on Process Parameters


yamelsenih:
Is nice to have that in SvrProcess the mandatory process parameters validation.
Also is good have a method for get parameter, example:

```java
  protected String doIt () throws Exception {
     boolean isSOTrx = getValueAsBoolean("IsSOTrx");
     int C_Invoice_ID = getValueAsInt("C_Invoice_ID");
  }

e-Evolution:
Hi Yamel I, think that is best remove of underscore getParameterToAsTimestamp(String parameterName)

This is should a best way when using the IDE, the other hand the java the Upper Case is used to separate world

kind regards
Victor

yamelsenih:
Hi @e-Evolution i put it like **getParameter_ToAsTimestamp(String parameterName)** because the parameter name also is ending in **_To**. But it is true, i will change the underscore in **To** parameter.

mckayERP:
Problem with the code to validate process parameters does not allow zero values in IDs.  For the Role Access Update, the AD_Client_ID for System is zero.  The current code does not allow the Role Access Update process to run for System.

yamelsenih:
Hi @mckayERP thanks you for it, I will see it for change

mckayERP:
Morning Yamel.  Perhaps the getParameterAsInt function can return -1 if the value is null rather than 0.  Then you can test for < 0 when its an ID/Lookup.  Cheers.

yamelsenih:
Hi @mckayERP the change has been, you can test it?
Thank you a lot for find it.

mckayERP:
Tested OK. Did not test null values - current code will return null BigDecimal parameters as zero which could be interpreted as valid data.  Modified the MigrationLoader to add mandatory parameters and switched to class names instead of Process IDs.  Depends on #363.

PR #364, opened 2016-05-03
Hotfix/3.8.0#002 #325 #363 process param

mckayERP:
Update MigrationLoader to use class names and correct getProcessByClass.  Set mandatory parameter for Role Access Update.


mckayERP:
Can this be reviewed quickly?  RUN_MigrationXML fails during role access update.

Issue #359, opened 2016-04-29
Field group created as label overflow the text with ellipsis when the field label is short


ishanuda:
Hi,

I created a field group with a group type as label.
Assigned some field(s) to the created group.

If all the field(s) labels are short of the group content, 
half of the group text label replaced with three ellipsis as in the screen shot.

If at least one group content field have a long text label the group label does not show this behavior.

Is this a bug or normal behavior? 
Can it be configured to display the full text?

![field-group-label-type-text-overflow-ellipsis](https://cloud.githubusercontent.com/assets/8893919/14913965/2578f8ae-0e24-11e6-8d4b-4cf13388c003.jpg)

yamelsenih:
Hi @ishanuda i see it issue. i will to change it.

Thak you for report it.

ishanuda:
Hi Yamel,

I am sorry to ask this, but how can i download the fix, or should I edit my source code and recompile to see the changes???

yamelsenih:
Hi @ishanuda do you have the current source code?
You must seek in [hotfix#002](https://github.com/adempiere/adempiere/tree/hotfix/3.8.0%23002) branch and do it pull for get news changes, after just run from eclipse or build your project

we hope your comments.

Best regard

yamelsenih:
Check the image bellow
![issue165](https://cloud.githubusercontent.com/assets/2333092/14954831/41dc52b2-103f-11e6-8594-21e6e0cf8877.png)

PR #370, opened 2016-05-05
Hotfix/3.8.0#002

ishanuda:
Testing the Field group created as label overflow the text with ellipsis when the field label is short #359


e-Evolution:
I close because this pull request have a lot the change , so can I review if you send any change from a fresh and last commits in branch

PR #375, opened 2016-05-07
Bug fix at ProcessParameterPanel 

tasosbull:
When initial client setup form is open i miss some top fields
like client, org etc
This is solved with a JScrollBar at ProcessParameterPanel .centerPanel


e-Evolution:
Hi I saw that re using a old version code , master not is update until that a new version is release , so you should test based on adempiere:hotfix/3.8.0#002

Issue #426, opened 2016-06-13
MPPOrderBOMLine release stock reservation BUG


bpali:
Say you have a draft Manufacturing Order document.

If you decrease the quantity on it, the components stock reservation should decrease also, but instead of decreasing proportionally with the order, they increase by the new amount required.

This bug is created by the condition imposed to the release of stock reservation (see below code)

_org.eevolution.model.MPPOrderBOMLine.beforeDelete()_
```java
protected boolean beforeDelete()
{
	// Release Reservation
	if(MPPOrder.DOCSTATUS_InProgress.equals(getParent().getDocStatus()) || 
	   MPPOrder.DOCSTATUS_Completed.equals(getParent().getDocStatus()))
	{	
		setQtyRequired(Env.ZERO);
		reserveStock();
	}			
	return true;
}
```

 which should simply be

```java
protected boolean beforeDelete()
{
	// Release Reservation
	setQtyRequired(Env.ZERO);
	reserveStock();
	return true;
}
```

because Manufacturing Order documents do stock reservation even when they are "Draft" status

e-Evolution:
The unique way to reserve stock is when the order had in process status or complete status , so may be that manufacturing order reserve with draft status , if this is this way then we have other issue , because the unique way to reserve stock is if manufacturing order have in process or complete status.

This reason because have the validation for any this document status.

Is the manufacturing order create reserved stock with draft status ?

bpali:
Manufacturing Order documents do stock reservation in Draft status. This is the only exception in Adempiere from what a Draft status means.

This kind of behaviour is required by MRP in the sense that when it runs, it creates Draft Manufacturing Orders which create demans by their own, and then goes further and creates other Manufacturing Orders generated by previous demands. So in order to keep things uniform within MRP records, stock reservation is done for any demand like the Draft Manufacturing Orders are. They are kep in draft status so that they can be deleted and recalculated when MRP runs again next time.

The above is what I understood from what I've seen and used in Adempiere.
It might not be the correct behaviour but what I do know for sure related to current issue is the following:

When there is no document status related condition at MPPOrderBOMLine creation, it should be the same when you delete it. Otherwise a disequilibrium is created as initially reported.

bpali:
Maybe @mckayERP can come with an opinion here

e-Evolution:
The functionality is similar that a sales order so that the sales order in draft status not reserve stock

[...]

PR #435, opened 2016-06-17
#426 Bugfix MPPOrderBOMLine stock reservation

bpali:
This pull request fixes issue #426 and establishes that only InProgress/Completed Manufacturing Order do stock reservation for their components


e-Evolution:
this pul request have a lot of change , I close , but you can send again if the patch or change come from a fresh or las commits in branch

Issue #297, opened 2016-04-05
Payment Selection must be like ADempiere Document


yamelsenih:
The **Payment Selection** currently is a standard table and it does not have **Document Action**, it cause many trouble when you try void it or when it is handle in ADempiere

e-Evolution:
Hi Yamel , THE Payment selection is necessary have control of approval process is ok to add all fields that allows to management a ADempiere document standard. So we need add fields.

C_DocType_ID
DocDate or DateTrx
DocAction
DocStatus 
ProcessedOn
Processing is currently 
Processed is currently 
Posted 

if now payselecion will be a document then is necessary create your respective workflow document

About of functionality would be nice add support the multiple currencies and prepayment 

So I think that can add the prepayment flag and the C_Currency_ID when the invoice have a different currency that not is same that bank account.

kind regards
Victor

e-Evolution:
I forgot, if this applies to C_PaymentSelection the same thing should be done to HR_PaymentSelection

yamelsenih:
Hi @e-Evolution i think that the **HR_PaySelection** can be supported in C_PaySelection

yamelsenih:
Wiki of new changes: [Spin Contribution: FR: Payment Selection Like Document](http://wiki.adempiere.net/Spin_Contribution:_FR:_Payment_Selection_Like_Document)

e-Evolution:
The pay selection manual not should be deprecated , a lot of user use this to print check in short , use new window simple selection invoice and pay is a large process.

e-Evolution:
I revert the last commit , @yamelsenih  will add a new smart browser with same functionality  

Is important that the new smart browser emulate the same functionality of payment selection manual

![captura de pantalla 2016-08-17 a las 7 50 42 p m](https://cloud.githubusercontent.com/assets/786968/17758335/09f12a02-64b4-11e6-9c85-022bdabf36df.png)

yamelsenih:
Hi everybody, the change is done.

Best Regards

PR #462, opened 2016-07-26
Feature/adempiere 297 improvement payment selection

yamelsenih:
Hi, it pull request have the changes related to #297 
The new payment selection handle it like document with document action, also is added support to payment request.


e-Evolution:
Hi Yamel , I reviewed your pull request changes, but exist an problem here, the org.eevolution.hr_and_payroll was removed to move all clases moved into of base.

So this no is a good idea if we want improve uncouple of code for each  bounded context of the business functionality.

The idea in future should be separate the functionality to a specific package class by bounded context  , I am planning to begin the code  uncouple after the release 382 and change the dependences management and leave of old approach Ant script.

A big problem of our legacy code is that the entity model class is attached and join with persistence engine , so the persistence engine should be a service that is implement from a interface , the classes  that describe the entities and business logic should be agnostic of the infrastructure layer.

So can you restore the org.eevolution.hr_and_payroll and resolve of dependence using the X_ and I_ class that this ara now in base.

some disventage that can cause have a monlolitic code are:

1.- if need change some business logic of bounded context in this case human resource and payroll ,I only need to regenerate liberoHR.jar and liberoHRzk.jar , if all org.eevolution.hr_and_payroll code is move to base , then now I need deployment all system not only a simple jar file.

2.- If now in future want go the new architecture schema based in Microservice, then  We need  begin to uncouple the base directory in smaller pieces.

If you need my help to get that this new funcionality work, you can contact me.

yamelsenih:
Hi @e-Evolution thank you for your recommendation, i will to change it

yamelsenih:
Hi @e-Evolution with it change the refactory project is done
Thank you

e-Evolution:
Hi Yamel , now I reviewing your contribution.

Will set here my objections and suggestion that should be solve to proceded to completa integration.

 Objection 1 The Document No  should be updated 

captura de pantalla 2016-07-28 a las 8 43 32 p m

The new window not allows entry info in document as before , so the behavior that other document i.e. <1000000> if document not is modify then take original sequence.

captura de pantalla 2016-07-28 a las 8 47 08 p m

Objection 2 The Name not should be removed , I agree that now not should be mandatory but the field should continue for legacy compatibility , the previos  installation need this reference , the other hand how you are think migrate from old DB that have all references in name.

e-Evolution:
Objection 3 The Sales Transaction not should be include here , so that is payment selection it can be confuse why should exist AR Invoice in a Pay selection , this should be removed and always filter only AP Invoice .

captura de pantalla 2016-07-28 a las 9 14 10 p m

Suggestion 1 -> two new filter should be added  , the first is currency , I think that is nice filter the invoices to a specific currency , the second should be added the Order Reference , in import product all invoice have same Order Reference or import code , so will nice enable filter by order reference to payment a import document.

captura de pantalla 2016-07-28 a las 9 22 15 p m

e-Evolution:
Suggestion 2 -> The column of Pay Amt should editable , this way the user can set here of mount that want pay of this is parcial payment on fly .

captura de pantalla 2016-07-28 a las 9 33 00 p m

e-Evolution:
Objection 4 -> Review pay selection line for payment rule legacy payment selection show 5 payment rule.

captura de pantalla 2016-07-29 a las 9 37 58 a m

Now in new approach only are show 4 payment rule , so the Direct Debit was removed but not should be so.

captura de pantalla 2016-07-29 a las 9 43 28 a m

because currently exist business logic for this payment rule.

captura de pantalla 2016-07-29 a las 9 47 43 a m

yamelsenih:
> Objection 4 -> Review pay selection line for payment rule legacy payment selection show 5 payment rule.
> 
> captura de pantalla 2016-07-29 a las 9 37 58 a m
> 
> Now in new approach only are show 4 payment rule , so the Direct Debit was removed but not should be so.
> 
> captura de pantalla 2016-07-29 a las 9 43 28 a m
> 
> because currently exist business logic for this payment rule.
> 
> captura de pantalla 2016-07-29 a las 9 47 43 a m

Hi @e-Evolution the direct debit is used for sales transaction the payment selection is for PO, if the payment rule **Direct Debit** it generate a Receibe instead a payment.

yamelsenih:
> Suggestion 2 -> The column of Pay Amt should editable , this way the user can set here of mount that want pay of this is parcial payment on fly .
> 
> captura de pantalla 2016-07-28 a las 9 33 00 p m

The Payment Amount is updateable, are you sure of test it?

See the image: 
issue235

yamelsenih:
> Objection 3 The Sales Transaction not should be include here , so that is payment selection it can be confuse why should exist AR Invoice in a Pay selection , this should be removed and always filter only AP Invoice .
> 
> captura de pantalla 2016-07-28 a las 9 14 10 p m
> 
> Suggestion 1 -> two new filter should be added  , the first is currency and currency type, I think that is nice filter the invoices to a specific currency , the second should be added the Order Reference , in import product all invoice have same Order Reference or import code , so will nice enable filter by order reference to payment a import document.
> 
> captura de pantalla 2016-07-28 a las 9 22 15 p m
> 

The Sales transaction exists in the Payment Selection Line, it is a legend column, i think that is better create a new widow for process sales transaction, tha buiness logic for it is old in ADempiere and it behavior is like a receibe.

yamelsenih:
> Hi Yamel , now I reviewing your contribution.
> 
> Will set here my objections and suggestion that should be solve to proceded to completa integration.
> 
>  Objection 1 The Document No  should be updated 
> 
> captura de pantalla 2016-07-28 a las 8 43 32 p m
> 
> The new window not allows entry info in document as before , so the behavior that other document i.e. <1000000> if document not is modify then take original sequence.
> 
> captura de pantalla 2016-07-28 a las 8 47 08 p m
> 
> Objection 2 The Name not should be removed , I agree that now not should be mandatory but the field should continue for legacy compatibility , the previos  installation need this reference , the other hand how you are think migrate from old DB that have all references in name.
> 

I thik that it must be asked to FT:

1. All generated document by company are a document number and one sequence, if it number is changed then why use sequence?
2. The document number can be updateable when the document is write from a external document, example: a AP Invoice or a AP Credit Memo, a Collect...

yamelsenih:
> Hi @yamelsenih nop , I made coverage test and is used current in a no sales transaction , you can check of class [MPaySelectionCheck.java](https://github.com/adempiere/adempiere/blob/master/base/src/org/compiere/model/MPaySelectionCheck.java#L406)  , other way that can check search of used reference for X_C_Order.PAYMENTRULE_DirectDebit static property.

Hi @e-Evolution if you generate the payment, the result is that it is generated like a Receipt and if you see the class [MPayment](https://github.com/adempiere/adempiere/blob/master/base/src/org/compiere/model/MPayment.java#L275)
The follow Images show the test:
1. I make a Payment Selection:
issue240

2. Add a Line (Nothe the payment rule is **Direct Debit**).
issue238
3. See the Prepared Payment.
issue239
4. See the Generated Payment like **Receipt**.
issue236

e-Evolution:
Suggestion 3 -> my recommendation is that we are management two behavior with same documents 

The request payment and pay selection , the different between first with second is only the the first have not bank account.

But I think that of workflow should be different for a request payment of a payment selection.

the request payment should only have 4 document status (Draft , In Process , Not Approved , Close), So a request payment not approved or pending to assign to pay selection should have in process status , when the request payment is assign to pay selection then of status of request payment should be close. the reasons is because the complete action for request payment have not business logic and in ADempiere always we hope some business logic to completed a document.

the pay selection should have 5 documents status (Draft , Not Approved , Complete , Reversed , Close)

a pay selection when is completed execute the business logic , this case generate the payment selection check , it is nice because we hope a behavior when a document action as complete it is executed.

If a pay selection is based in request payment and pay selection is completed the request payment should change the document status of in process to closed , this way the request payment not can be take for other pay selection.

Other advantage is that if you create a report for pay selection window , is very simple filter of records with doc status in process this way I can know o get the request payment pending of be assign to pay selection. on the other way I have not a way to separate of payment request of pay selection , I would but create a complex filter as (Bank account should null , and document status should be complete , and some way indicate that the record not is assign to pay selection) more work more complex.

What do you think ?

yamelsenih:
> Suggestion 3 -> my recommendation is that we are management two behavior with same documents
> 
> The request payment and pay selection , the different between first with second is only the the first have not bank account.
> 
> But I think that of workflow should be different for a request payment of a payment selection.
> 
> the request payment should only have 4 document status (Draft , In Process , Not Approved , Close), So a request payment not approved or pending to assign to pay selection should have in process status , when the request payment is assign to pay selection then of status of request payment should be close. the reasons is because the complete action for request payment have not business logic and in ADempiere always we hope some business logic to completed a document.
> 
> the pay selection should have 5 documents status (Draft , Not Approved , Complete , Reversed , Close)
> 
> a pay selection when is completed execute the business logic , this case generate the payment selection check , it is nice because we hope a behavior when a document action as complete it is executed.
> 
> If a pay selection is based in request payment and pay selection is completed the request payment should change the document status of in process to closed , this way the request payment not can be take for other pay selection.
> 
> Other advantage is that if you create a report for pay selection window , is very simple filter of records with doc status in process this way I can know o get the request payment pending of be assign to pay selection. on the other way I have not a way to separate of payment request of pay selection , I would but create a complex filter as (Bank account should null , and document status should be complete , and some way indicate that the record not is assign to pay selection) more work more complex.
> 
> What do you think ?

I thin that is nice view the follow BPM:
payment selection
The Process is created own it.

Best Regard.

yamelsenih:
Wiki of new changes: [Spin Contribution: FR: Payment Selection Like Document](http://wiki.adempiere.net/Spin_Contribution:_FR:_Payment_Selection_Like_Document)

yamelsenih:
> @yamelsenih 
> The standard functionality in ADempiere when you not entry a value into Document No , if  document type not was defined as document is number control or not exist document sequence , then the record id is used to set Document No value.
> 
> but the more strong argument is that in some countries the payment request first is create in physical paper when we collect all signer,  then the request payment is entry in system, with current approach not is possible put  the  sequence preprint of the physical document and it not is good, this reason because should be updatable when is new record.
> 
> captura de pantalla 2016-07-29 a las 11 22 45 a m
> 
Hi @e-Evolution i ask: why create a payment request in a physical paper if it can be generated from ADempiere and print? the generated document have a sequence and the payment request functionality allow create in ADempiere the document that is create manual currently. Please see the wiki for view how to work the payment request: Wiki of new changes: [Spin Contribution: FR: Payment Selection Like Document](http://wiki.adempiere.net/Spin_Contribution:_FR:_Payment_Selection_Like_Document)

Best regards.

SusanneCalderon:
To Objection # 1 from e-Evolution: We tested after running the latest xml. The field "name" is still in the window. The column "name" of table c_payselection is not mandatory in the AD, but we had to run manually the following sql in postgres, because "Synchronize Column" did not alter the not null constraint in the DB.
ALTER TABLE C_PaySelection ALTER COLUMN Name DROP NOT NULL;
After running this command we could save without defining a name.

SusanneCalderon:
To objection # 1 from e-Evolution. 
The document no was updated following the document type sequence rules.
This objection is solved.

SusanneCalderon:
To objection #3 from e-Evolution.
The document status of a payment request should not automatically set to close. It is posible to leave selection lines of a payment request for a later payment. 
So the status should change to close only when all lines have been assigned to a pay selection.
When a pay selection is reversed the assigned payment request should change the status back to in process.

SusanneCalderon:
To Suggestion 2 from e-Evolution. -> _The column of Pay Amt should editable , this way the user can set here of mount that want pay of this is parcial payment on fly_ .
This suggestion is ok, it should be enhanced by a callout to adjust the openamt, because it changes with a different payamt.

SusanneCalderon:
To objection #3 from e-Evolution: _The Sales Transaction not should be include here , so that is payment selection it can be confuse why should exist AR Invoice in a Pay selection , this should be removed and always filter only AP Invoice ._
I agree with this objection, in my opinion payselection should be used for payments, not for receipts.

SusanneCalderon:
To objection #4 from e-Evolution. _Review pay selection line for payment rule legacy payment selection show 5 payment rule. Now in new approach only are show 4 payment rule , so the Direct Debit was removed but not should be so._
I have not tested the payment rule direct debit, so I have no opinion about it.

mckayERP:
A few of observations (in swing only):

1. The search criteria don't seem to be applied.  If I want to only pay AP Invoices without any matching, the list still shows ALL document types.
2. The Create From Invoice list should be the payment schedule, not the invoices.  I only want to pay the part of the invoice that is due, not the whole invoice.  Its not clear if the "Payment amount" is the remaining open amount or the due portion of the payment schedule.  The previous Payment Selection window showed payment number/total payments for example 3/12.
3. The header of the Create From Invoice table has lost the checkbox in the selection column so the user cannot select all.
4. There needs to be a way to see the bank cash available and the total of the selected items.  Without this info, its difficult to deal with limited cash situations.
5. You can't sort the columns in the create from search results by clicking on the header. 

I'm just starting to test it, so I might have a few more comments.

e-Evolution:
@mckayERP 

> 3 The header of the Create From Invoice table has lost the checkbox in the selection column so the user cannot select all.

Mike you can the icon select all, this have the same approach that previos version.

mckayERP:
I thought the header approach was more effective - kept the controls in the table and was intuitive.  This icon isn't.

Re the search criteria don't seem to be applied and the values aren't remembered.

I think this has to do with the handling of the vetoable change events in the criteria fields.  This is a bit of a messy design in adempiere as there is the field (display), field editor, and gridField which all have to share the same value.  If the user changes the field editor, it fires a vetoable change event **but does not set its own value**. The vetoable change listeners can veto the change but mostly sync the girdField with the new value (See the GridController).  The vetoable change in the GridController also handles multiple selection.  The gridField value change triggers a property change which the field listens to and uses to set its value and to set the display of the new value.

In the VBrowserSearch, the vetoable change listener does not sync the gridField so the model does not get the new value.  This means that the criteria is not applied to the search and the value is not set in the field when the criteria are expanded again.

I made a change to add a property change listener function in the VBrowserSearch class and moved the current veto code there and then copied some of the GridController vetoable change code into the vetoable change listener in VBroswerSearch. This last sets the value in the gridField which was the missing piece.  The search criteria now affect the search and the criteria field remember their values.

A similar change was made to the new ProcessParameterPanel.  

I haven't tested ZK yet but a similar solution would likely work there too.

mckayERP:
Also, the gridField setValue method updates the context, so the BrowserSearch processNewValue method does not need to do so.

yamelsenih:
Hi @mckayERP 

> A few of observations (in swing only):
> 
> 1. The search criteria don't seem to be applied.  If I want to only pay AP Invoices without any matching, the list still shows ALL document types.

I will test it.

> 2. The Create From Invoice list should be the payment schedule, not the invoices.  I only want to pay the part of the invoice that is due, not the whole invoice.  Its not clear if the "Payment amount" is the remaining open amount or the due portion of the payment schedule.  The previous Payment Selection window showed payment number/total payments for example 3/12.

We was testing it and is worked for payment schedule, i don't know if exists a error in it, see the follow images:
Making the invoice:
issue2011
Watching the payment schedule:
issue2012
See the payment schedule by the same document:
issue2013

I will add the payment count 1/2, 2/2.

> 3. The header of the Create From Invoice table has lost the checkbox in the selection column so the user cannot select all.

I agre with you, i think that the buttons for select should not be in the top. It can be like column selection, is more clear.

> 4. There needs to be a way to see the bank cash available and the total of the selected items.  Without this info, its difficult to deal with limited cash situations.

Yes i agree, i will add it on search.

> 5. You can't sort the columns in the create from search results by clicking on the header. 

It is a bug of Smart Browser.

> I'm just starting to test it, so I might have a few more comments.

Best Regards,

yamelsenih:
Hi everybody, the change is done.

Best Regards

e-Evolution:
More unstable functionality generate for pull request 

captura de pantalla 2016-10-18 a la s 18 49 30

Issue #508, opened 2016-08-17
Buttons on VLookup fields in Find window don't work.


mckayERP:
When opening a window where a search/find window appears before the window appears, if a VLookup (Search) field is used as a parameter, the button in the field will cause an NPE in the Info.java code as line 350:

`if (frameClass == AWindow.class)
        {
            //  Activated from a window - find the active tab to limit the context
            p_TabNo = ((AWindow) frame).getAPanel().getCurrentTab().getTabNo();
        }
`
The problem is that the call to getCurrentTab() is null as the variable m_curTab in APanel.java has not been initialized by the APanel code.  

Solution is to initialize the m_curTab to the first tab as the APanel is being initialized.

yamelsenih:
Hi @mckayERP i think that is better if is validate the getCurrentTab() with null

yamelsenih:
Hi, it is solved.

PR #509, opened 2016-08-17
#508 Buttons on VLookup fields in Find window don't work. 

mckayERP:
https://github.com/adempiere/adempiere/issues/508.

Initialized m_curTab as tab 0 when the window is being initialized.

Also moved the initialization of m_curWinTab up a few lines to be closer to the window section where it fits with the logical flow.
Also cleaned up the header section acknowledgments to use a consistent format.

### DIFF
--- a/client/src/org/compiere/apps/APanel.java
+++ b/client/src/org/compiere/apps/APanel.java
@@ -121,8 +121,8 @@ import org.eevolution.form.VBrowser;
  * 	@version 	$Id: APanel.java,v 1.4 2006/07/30 00:51:27 jjanke Exp $
  *
  *  Colin Rooney 2007/03/20 RFE#1670185 & related BUG#1684142 - Extend Sec to Info Queries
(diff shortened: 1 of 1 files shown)


e-Evolution:
Hi Mike 

Why modify of credit autor in of class?

mckayERP:
Just to make it consistent in format since there are lots of contributors to this file.  The `@see` moves the link down to the See Also section of the java doc which can make it difficult to relate the link to the change.  If you make the comment like <li><a href="link">text</a>  you get a better output.  This is what it looks like with the change.  No credits where changed, just combined and formatted.

apanerl_doc

e-Evolution:
Nop problem , I see the [change files ](https://github.com/adempiere/adempiere/pull/509/files) and show as if modification was delete lines , but I check and is ok

But you have a good point should we continue with this practice , so that now had github and git history.

I think we not should set here all changes into of class ,  What do you think ?

mckayERP:
OK.  I'll just add my note then and resubmit.

PR #512, opened 2016-08-17
#508 Buttons on VLookup fields in Find window don't work.

mckayERP:
#508.  Initialized m_curTab as tab 0 when the window is being initialized to prevent a NPE.

Also moved the initialization of m_curWinTab up a few lines to be closer to the window section where it fits with the logical flow.


e-Evolution:
Hi @mckayERP , can please send again this pull request from a fresh 390 code , please validate that your pull request have not conflict , this way we can integrate

Issue #598, opened 2016-10-07
Complete Integration of Production Light in ADempiere


marcalwestf:
Several important improvements and enhancements in order to integrate Production Light into ADempiere's infrastructure, processes and workflow.

1. Production becomes an ADempiere document (i.e. with docActions etc.)
2. Add a corresponding document type which among other things reserves material for production
3. Implement Partial Production
4. Integrate _PPProductBOM_ into Production Light, substituting _MProductBOM_
5. Integration of Production Light into Cost Engine
6. Misc improvements developed by Adaxa such as preparation for _Mini MRP_, material movements to production location, etc.

PR #601, opened 2016-10-10
Complete Integration of Production Light in ADempiere

marcalwestf:
#598: Complete Integration of Production Light in ADempiere

Implementation of issue https://github.com/adempiere/adempiere/issues/598 .


e-Evolution:
Mario let solve conflict directly into source fork

PR #749, opened 2017-02-04
Sandbox/4.0.0#001

sramazzina:
Hi All,

I just sent a minor pull request containing things related to overall project's cleanup. I also added some useful entries to .gitignore file. In my opinion would a good idea to merge all these changes in the main project


e-Evolution:
in process when we release 390 we can work with develop to included this restructure

Issue #734, opened 2017-01-25
Error when try posting  a payroll process document


e-Evolution:
Che try posting a payroll process document , the problem is that is incorrect sql for dimension

PR #811, opened 2017-03-06
fix #810 Accounting of hr process doesn't filter by account schema

### LINKED ISSUE #810: Accounting of hr process doesn't filter by account schema
If you have 2 or more account schemas by client the accounting result is duplicated.

jdaison:
This PR fix issue #810


e-Evolution:
Hi John!

I saw your code , that the patch not is complete because should be filter in consult by account schema id

e-Evolution:
was integrate with https://github.com/adempiere/adempiere/issues/734

PR #1342, opened 2017-09-26
Feature/gradle.001

sramazzina:
This pull request contains a script that applies a first set of changes to the first 4 directories (base, client, org.adempiere.assets, zkwebui) so that the directories layout will be changed to the Maven pattern. _PLEASE do not consider my README.md in the pull request. Consider only the last two commits_.

The script is located in a directory called _gradle.001_ which has a subdirectory _etc_ that contains updated _build.xml_ files that are related to the projects that are directly or indirectly impacted to layout changes of the four directories we reworked. I ask to merge this pull request because it has not impact of course on the project but your free in case to launch the script from the directory where it is located and look at the effect. 

After applying the script, the project fully recompiles the usual way with the newest project structure for these 4 sub-projects. The script contains also some statements to cleanup things that are old or they don't need to stay in repository's a project's structure but only local to the developer's environment. Feel free to test it and let me know. In the meanwhile I will proceed to rework on my codebase the rest of the project the same way.


yamelsenih:
This pull request have conflicts

PR #1327, opened 2017-09-10
build Release/3.9.0 in eclipse OXYGEN with java 1.8 from scratch

homebeaver:
to compile base+client I have to change some settings in `.classpath` and `.project`

PR #1353, opened 2017-10-04
eliminate duplicate keys in AccountingDE_SKR03

homebeaver:
accounts no: 1900 and 8705

As I wrote [here](https://github.com/adempiere/adempiere/pull/1327#issuecomment-332138692) this pull does not depend to the developer's environment


e-Evolution:
@homebeaver thank a lot for your contribution

Issue #1451, opened 2017-11-03
Add additional condition to an Standard Request 


sbouissa:
To solve this necessity, we think in adding a new column in R_RequestStandardType table where we can code differents conditions:

-Condition for the object (PO) that is consider to create the requests (an AND condition to de select of the table object) ---> When the condition start whit "SQL=" 
Example : SQL= M_Product_ID IN (Select p.M_Product_ID from M_Product p where p.M_Product_Group_Id = XXXX) 

-Condition in where we can get values of the object that is being condider to create or not the requests. ---> When the condition start with @SQL=
Example:
@SQL= SELECT a.M_Product_ID FROM M_Product a 
JOIN M_Product_Category b ON a.M_Product_Category_ID = b.M_Product_Category_ID 
WHERE  a.AD_Client_ID = @AD_Client_ID@ AND b.M_Product_Category_ID = 105 AND a.M_Product_ID = @M_Product_ID@ AND a.name IS NOT NULL

-Condition for data of the objet that is beeing consider --> When isn't an @sql or sql
Example: AD_Org_ID = XXXX & AD_Client_ID = XXXX

sbouissa:
![issue1451](https://user-images.githubusercontent.com/4325868/32395051-49cb8462-c0be-11e7-8c0e-c9dd223c648c.png)

sbouissa:
I will add the column "WhereClause" on the table R_RequestStandardType if all of you are agree. If some think another element fit bether please tell me. 
Thankyou

e-Evolution:
Related #1330

e-Evolution:
@sbouissa 

Need review of implementation of where clause , this should work of same way that Alert Rule or Report View, you can use any context variable using @@ so not is necessary to use @SQL=

PR #1484, opened 2017-11-13
Add additional condition to an Standard Request #1451

sbouissa:


org.adempiere.request/src/main/java/org/compiere/model/RequestModelValidator.java
```diff
@@ -17,20 +17,27 @@
...
 /**
  * Request Model Validator
+ *
+ *  @author OpenUp Solutions Sylvie Bouissa, sylvie.bouissa@openupsolutions.com, http://www.openupsolutions.com
+ *      <li>#1451 Add additional condition to an Standard Request
+ *      <li>Reference to issue https://github.com/adempiere/adempiere/issues/1451
  */
 public class RequestModelValidator implements ModelValidator {
 
     Timestamp currentDate;
```
yamelsenih: Is nice if it can be private declaration by default

PR #1569, opened 2018-02-09
a forgotten translation for "MM Vendor Return"

homebeaver:


e-Evolution:
@homebeaver Thank a lot for contribution

Issue #1370, opened 2017-10-12
Enhance the process AllocationAuto with allocation of prepayments


SusanneCalderon:
This process can be used as a process in the workflow Process_Invoice in order to allocate a prepayment of an order with the newly generated invoice.

e-Evolution:
Hi Susanne @SusanneCalderon , can you more detail about the process that you want execute?

SusanneCalderon:
Hello, here is the method which es called at the end of complete().

```
private String allocatePrepayments()
	{
		if (!get_ValueAsBoolean("IsAllocateImmediate"))
			return "";
		if (getC_Order_ID() == 0)
			return "";
		if (getC_Order_ID() != 0 && getC_Order().getC_POS_ID() != 0)
			return "";
		BiFunction<BigDecimal, BigDecimal, BigDecimal> getMaxPayAmt     = (t1, t2)  -> {
			return t1.abs().compareTo(t2.abs()) <=0? t1:t2;
		};
		BiFunction<BigDecimal, BigDecimal, BigDecimal> getOverUnderAmt    = ( payamt,  openamt)  -> {
			return payamt.abs().compareTo(openamt.abs()) >=0? Env.ZERO:openamt.subtract(payamt);};

		BiFunction<BigDecimal, Boolean, BigDecimal> getAmtIsreceipt    = ( amt,  isReceipt)  -> {
			return isReceipt?amt:amt.negate();};

		String whereClause = "C_ORDER_ID=? and docstatus in ('CO','CL') and isallocated = 'N'";
		List<MPayment> prepayments = new Query(getCtx(), MPayment.Table_Name, whereClause, get_TrxName())
				.setParameters(getC_Order_ID())
				.list();
		BigDecimal openAmt = getOpenAmt();
		for (MPayment pay:prepayments)
		{
			MAllocationHdr alloc = new MAllocationHdr (Env.getCtx(), true,	//	manual
					getDateInvoiced()	, getC_Currency_ID(), Env.getContext(Env.getCtx(), "#AD_User_Name"),
					get_TrxName());
			alloc.setAD_Org_ID(getAD_Org_ID());
			alloc.saveEx();
			BigDecimal aLineAmount = getMaxPayAmt.apply(openAmt,pay.getPayAmt());
			BigDecimal overUnderAmt = getOverUnderAmt.apply(aLineAmount, openAmt);
			//	Allocation Line
[...]

PR #1596, opened 2018-03-19
adempiere:#1370 Enhance the process AllocationAuto with allocation of…

SusanneCalderon:
… prepayments

https://github.com/adempiere/adempiere/issues/1370
The automatic allocation must be omitted if the document is a reversion document (reversal_ID is not null)


yamelsenih:
Hi @SusanneCalderon it is ok, i will to merge

Issue #1499, opened 2017-11-22
Defining two Standard Request Type associated to the same table, event model validator and different where clause, the requests defined in each one duplicates if the conditions are ok.


sbouissa:
We are trying to create differents requests when a c_projectline is created.
First: we defined a standard request type for the table c_projectline "after insert" (with 13 requests)
Second: we defined a standard request type for the table c_projectline "after insert" and the conddition that the product defined into the c_projectline have to be of 'Example' group.(with1 request)

We create 2 c_projectline (N.1) one with a product associated to the group "Example" and other (N.2) with the product associated to the group "NO"
At the end there are 28 (13+13+1+1) requests associated to the c_projectline (N.1) and 26 (13+13) requests associated to the c_projectline (N.2)

SusanneCalderon:
I merged after some changes so I am cleaning my branch

PR #1602, opened 2018-03-21
#1499 Defining two Standard Request Type associated to the same table…

e-Evolution:
…, event model validator and different where clause, the requests defined in each one duplicates if the conditions are ok.  https://github.com/adempiere/adempiere/issues/1499

- Refactor the methods move to model class and implement other approach to add listener model validator

Related #1499 and #1451


org.adempiere.request/src/main/java/org/compiere/model/MStandardRequestType.java
```diff
@@ -176,4 +175,120 @@ else if (entity.get_ColumnIndex(MProject.COLUMNNAME_DateStart) >  0)
...
+     * @param entity
+     * @param standardRequestIsSOTrx
+     * @return
+     */
+    private Boolean isValidSOTrx(PO entity , String standardRequestIsSOTrx)
+    {
+        if (standardRequestIsSOTrx == null)
+            return true;
+
+        Boolean isSoTrx;
```
yamelsenih: It can be resolved with a native var like boolean instead Boolean

nsarlabos:
The tests are performed and it works correctly!

e-Evolution:
fixes

Issue #1592, opened 2018-03-12
Add support for set percentage of the sales vs forecast to commissions definition


RJPC:
Currently you can configure the commission definition by collection days, example:

0 to 10 days you pay 30% on sales
11 to 20 days you pay 20% on sales
21 to 30 days you pay 10% on sales

It is proposed to add percentages for sales compliance

75% to 80% of sales compliance is paid 20% on sales
81 to 90% sales compliance is paid 30% on sales
91 to 100% sales compliance is paid 40% on sales

![pr1](https://user-images.githubusercontent.com/5696550/37468236-51fbe11e-285a-11e8-9d4b-d7e10f3e1225.png)

RJPC:
When the Maximum Percentage field located in the Commission Line is equal to null, there is no calculation of commission on that Line (product)

![linea de comision](https://user-images.githubusercontent.com/5696550/38060482-7ad4efd8-32da-11e8-9067-991af767cdc5.png)

marcalwestf:
it would be great if there were data in Garden World to test it...

marcalwestf:
@RJPC can you please explain what is "Collection" and "Activation" ?

RJPC:
Hi @marcalwestf , 

**Collection** is when the commission is based on receipt, for example:

**% commission:** 4%
**Invoiced:** 10.000 $
**Receipt:** 8.000 $
**Commission:** 8.000 * 4%
**Total Commission:** 320 $

**Activation** is when the commission is based on the client's portfolio is own sales representative, for example:

**Conditions for commission**
invoiced the 70 % of the clients

**% commission:** 1%
**portfolio client's:** 100
**invoiced clients:** 70
**amount invoiced:** 10.000 $
**commission:** 10.000 * 1%
**Total Commission:** 100 $

yamelsenih:
This was added yet

PR #1617, opened 2018-03-26
Feature/#1592 commission sales vs forecast

yamelsenih:
Hi Everybody, add it enhancement for commission, it allows create commission based in:

- Forecast vs Invoice: Get Quantity 
- Forecast vs Order

It make reference with #1592


marcalwestf:
loking at it ....

Issue #1613, opened 2018-03-26
Add support to the definition of commissions (by collection)


RJPC:
Validation must be added so that when calculating commissions do not take into account the reverse charges

Images regarding the configuration, reverse charge and calculation of commission

**Configuration**
![comision por cobranza](https://user-images.githubusercontent.com/5696550/37914935-e9fae54a-3107-11e8-9314-38f63cbb389d.png)
![linea de comision por cobranza](https://user-images.githubusercontent.com/5696550/37914951-eee19c20-3107-11e8-9b7d-d6061c5fe335.png)

**Charge reverse**
![Uploading Cobro Reversado.png…]()

**Caculation of commission**
![detalle de comision](https://user-images.githubusercontent.com/5696550/37915027-151d3cb4-3108-11e8-9329-e8e507e88348.png)

Open balances (where you can see that the bill for which you are commissioning has not yet been charged)
![saldos abierto](https://user-images.githubusercontent.com/5696550/37915250-9bfe0204-3108-11e8-8d62-dce606696a32.png)

marcalwestf:
hi @RJPC . Could you please describe with an example the new functionality (or is it a bug correction)?
This way we could figure out better the situation now and the proposed new behavior.
Thanks in advance.

RJPC:
hi @marcalwestf , It is a correction, currently the process generates commission for reverted charges, if the charge is reversed it should not generate commission for this charge since it is a zero charge.

yamelsenih:
It is caused by:
<pre>
h.DocStatus IN ('CL','CO','RE')
</pre>

yamelsenih:
line 683 of MCommissionRun

marcalwestf:
Code reviewed and indeed, the  the documents to be found (Payments and Invoices) must not have the docStatus='RE'.

yamelsenih:
This was added

PR #1618, opened 2018-03-26
#1613, fixed error with reversed document for commission, reference to

yamelsenih:
issue:https://github.com/adempiere/adempiere/issues/1613

Hi @e-Evolution @marcalwestf can you see it?

I add a patch for get only completed and closed document for commission, it make reference to issue: #1613 
Best Regard


marcalwestf:
looking at it...

marcalwestf:
As documented in https://github.com/adempiere/adempiere/issues/1613 , the issue is valid and the fix OK.

Issue #1619, opened 2018-03-26
Bug when reactivate a calculate of commission


RJPC:
when proceeding to reactivate a calculation of commissions the following message is shown

attached image and log

![reactivar](https://user-images.githubusercontent.com/5696550/37932120-37d03a70-3137-11e8-8f49-8e9ec77d4cf1.png)

[trace.log](https://github.com/adempiere/adempiere/files/1849498/trace.log)

marcalwestf:
looking at the code, it seems that the error can be produced when _MCommissioRun.reActivateIt()_ is called. In this case, the method _deleteMovements()_ is called.

The variable _m_comissionLog_ is not initialized, producing the error.

I completed a commission, then called to re-activate it.
The program sprang to
![20180327_commission_01](https://user-images.githubusercontent.com/1789408/38064382-f1d35fb8-32ba-11e8-909b-d5becc1d113b.png)

..and as expected, the variable _m_comissionLog_ is null.
With the following initialization, the method is executed correctly.

yamelsenih:
Hi Mario it is solved yet in the pull request https://github.com/adempiere/adempiere/issues/1621

PR #1621, opened 2018-03-26
#1619, fixed bug with NPE for reactivate commission, reference to issue:

yamelsenih:
https://github.com/adempiere/adempiere/issues/1619

Hi @all it is a pull request for resolved NPE for issue: #1619 

Best Regard


marcalwestf:
looking into it...

marcalwestf:
as documented in https://github.com/adempiere/adempiere/issues/1619 , the code is safe and solves the issue described.

PR #1626, opened 2018-03-27
Price entries

SusanneCalderon:
A Smartbrowser to generate or edit prices in the window Pricelist

### DIFF
--- a/base/src/org/adempiere/process/CreateProductEntriesInPriceList.java
+++ b/base/src/org/adempiere/process/CreateProductEntriesInPriceList.java
@@ -1,0 +1,56 @@
+/******************************************************************************
+ * Product: ADempiere ERP & CRM Smart Business Solution                       *
+ * Copyright (C) 2006-2017 ADempiere Foundation, All Rights Reserved.         *
+ * This program is free software, you can redistribute it and/or modify it    *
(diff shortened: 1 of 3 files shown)


e-Evolution:
@SusanneCalderon please no use more null reference in your code, you use for Optional so you can avoid finding found possible future NPE.

On the other hand, when adding a new feature to the dictionary for Adempiere core you should use Entity Type ' D', that means Dictionary the C or A never was used.

[Java 8 Optional](http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html)

i.e.

`Optional<MProductPrice>  productPriceOptional  = Optional.ofNullable(MProductPrice.get(getCtx(), getRecord_ID(), key, get_TrxName()) ); 
        MProductPrice productPrice = productPriceOptional.orElse(MProductPrice(getCtx(), getRecord_ID(), key, get_TrxName()));
`

SusanneCalderon:
OK, Victor, I will change the code, thank you for your observation.

wdtz:
**Test results:**
1.- The base operation is correct

**Improvements:**
1.- Product Group could be part of the filters
2.- Business Partner (Vendor) could be part of the filters

SusanneCalderon:
Thank you Waditza, I will update the smart browser.

Issue #1237, opened 2017-07-10
Smart Browser: Editable fields are not saved correctly in ordered tables


SusanneCalderon:
When you order the minitable of a smartbrowser and select 2 or more lines which are not contiguous the saved values are not correct.
How to reproduce:  a smart browser with editable fields.
Example: Smartbrowser "Create bankstatementlines from payment".
I ordered the minitable by documentno and selected the first and third line. (screenshot1)
The generated bankstatementlines are assigned to the correct payment, but the second statement amount is $-8.63, which was the amount of the 4. line in the ordered table.
In the DB the saved converted amount is $-8.63 instead of $-23.65 of the selected payment. 
**1.screenshot, the ordered minitable**
![imagen](https://user-images.githubusercontent.com/15349639/28002715-e0a35976-64f4-11e7-8ceb-a81b336c6e34.png)
**2. screenshot: the generated lines**
![imagen](https://user-images.githubusercontent.com/15349639/28002729-fe1d0b28-64f4-11e7-90cf-88349a4b1d77.png)
**3. screenshot, table t_selection_browse**
![imagen](https://user-images.githubusercontent.com/15349639/28002737-0e316784-64f5-11e7-8324-2bc54a8c626c.png)

PR #1629, opened 2018-03-27
#1237 Smart Browser Sorting and Browsing

SusanneCalderon:
The sort funcionality is omitted when you deactivate the field isAllowSort in the smart browser. The default value is Y, so the behaviour is only changed when you edit this field.


e-Evolution:
@SusanneCalderon 

I not understand the  Bugfix/#1237 what mean this prefix ?

I think not is necessary adding other flag IsAllowSort when exist Order By and Order Sequence.

So you would use a logic if the Order Sequence is >  0 then is allow sort.

captura de pantalla 2018-03-27 a la s 17 01 39

captura de pantalla 2018-03-27 a la s 17 00 23

SusanneCalderon:
@e-Evolution , Hi Victor, if you use only sort order, this order by  is not updatable. Working with a smart browser sometimes it is necessary to order by other criterias, as amount, business partner....

Example: The smart browser for the reconciliation. When I open the smart browser the result is ordered by ID, maybe I want to order it by date acct or by amount, so I can compare better with the statement, generated by the bank. The payamt in the code is taken from the browsefield, the list in the code is not updated after the order, so I get false results in the reconciliation line.

e-Evolution:
@SusanneCalderon , but currently you can defined how want sort the record for a Smart Browser,

@yamelsenih  implemented this functionality , may be can you ask he before the implement anything?

SusanneCalderon:
Hello Victor, the order by clause is defined as SystemAdministrator in System. Sometimes the user needs to order the result of the search in a smart browser. So it is convenient to use the sort funcionality of CTable.
But it only works if you have no editable fields in the smart browser and use just the key in the corresponding function. 
Every user nowadays expects a grid to be sorted and it is considered as an error if he cannot sort. So with this flag you can provide sortable smart browser tables if you have no editable fields, which is an improvement to the non sortable smart browser. 
Generally iy would be a great improvement to provide the sort functionality for every smart browser.

beruskajimenez:
Hello @SusanneCalderon , allows you to sort even when the AllowSort field is disabled, that is, there is no difference when it is active or inactive

SusanneCalderon:
Hello @beruskajimenez . I will test it now.

SusanneCalderon:
Hello @beruskajimenez . The omission of the sorting does only work for swing. In webui the problem is different, you can sort always, but when you sort and try to run the process, nothing happens.
So in webUI the problem is different in swing, where the process is run with incorrect data.

I upload a video with the smartbrowser create from payment in the window bank statement.

First I start the smartbrowser and select the 2 first lines. The bankstatement lines are generated correctly.
Then I start the smartbrowser again, order by amount and select a line. The bankstatement line is not generated.
[SmartBrowserSort.zip](https://github.com/adempiere/adempiere/files/2095357/SmartBrowserSort.zip)



See the uploaded video.

beruskajimenez:
Hello @SusanneCalderon , I did the tests in the swing and it works correctly, It could be considered an improvement if the change is applied to the web environment. Thank you!

wdtz:
**Result of the Test:**
Successful

yamelsenih:
Hi @SusanneCalderon i don't sure if it change can be util for current problem. I seem that you make a button for disable sorting instead fix the bug referenced on #1237

yamelsenih:
Hi, i think that is unnecessary disable sorting for SB.

Is better fix the original bug for sorting and values on SB

SusanneCalderon:
Hi @yamelsenih. I created this issue some time ago. I consider this only a temporary solution until the original bug is resolved.
So you can use the order function when you do not use editable fields. 
I tried to find a solution in the smart browser enviroment, but I do not know how to reorder the hashmap created by the entries in T_Selection_Browse.

marcalwestf:
As I understand , the best would be for the moment to programmatically disallow the sorting of the browser table if a single editable field is encountered.

This solution would be better than spending a field/column which later will be eliminated anyway. Also, the users would not be confronted with the task of taking care of disabling the sorting when a field would be defined as editable.

This way smart browsers would be consistent meanwhile the sorting error is found and fixed; once this has been done, the programmatically disabling of sorting will become obsolete.

SusanneCalderon:
Hello @yamelsenih, I omited the field _isAllowSort_ as you suggested

yamelsenih:
Hi @SusanneCalderon I will test it.
Are you sure that it bug not exist in ZK UI?

yamelsenih:
Hi @SusanneCalderon it change apply for VBrowseTable and it is for Swing UI, exist the same bug for ZK UI?

SusanneCalderon:
Hi @yamelsenih 
In ZK the behaviour is different, you can always sort.
But the process unerlying the smart browser does nothing after sorting the browse table.
Here is a video with the 2 behaviours in zk.

[SmartBrowserSort.zip](https://github.com/adempiere/adempiere/files/2095357/SmartBrowserSort.zip)

Issue #1609, opened 2018-03-23
Enhancement Move Scala Testing for Test Scala Directory


e-Evolution:

e-Evolution:
implement with 0d1c855e7d5160045cca402d10ea6872e3360b37

Issue #1595, opened 2018-03-15
Enable migration script for ADempiere 3.9.0.a


e-Evolution:
Now the ADempiere separate the XML migration script for previos version

Issue #1580, opened 2018-02-24
The enhancement allows defining the payment document type into the document type for pay selection. 


e-Evolution:
The enhancement allows defining the payment document type into the document type for pay selection. 

When pay selection is completed the payment will be generated with the payment document type defined in document type of the pay selection.

SusanneCalderon:
@e-Evolution . Hello Victor, comparing I_C_Doctype of your branch with the class in adempiere/develop I see that the column isReversedWithOriginalAcct is omitted. Did you delete this funcionality?

e-Evolution:
@SusanneCalderon thank a lot for your suggestion , this was fixed into e8a9a94

SusanneCalderon:
@e-Evolution Hi Victor, I looked at the changes for this isssue in the pull request 1631 and the methods for the columns revertwithoriginal acct are deleted. I do not know if this will give you troubles with this enhancement in 3.9.1.
are omitted.
![imagen](https://user-images.githubusercontent.com/15349639/38748932-0d4cf0e8-3f0d-11e8-9649-55aec3cf7f12.png)

e-Evolution:
@SusanneCalderon this change was included again into e8a9a94 , I regenerate the model class and fix this problem.

Issue #1564, opened 2018-02-06
Add Project Model Validator to Catch multiples Events


e-Evolution:
This Project Model Validator check if a payment is linked to project and payment is a receipt then adding the payment amount to project balance and if payment is linked to project and is not receipt then subtract payment amount of project balance

PR #1631, opened 2018-03-28
Synchronize e-Evolution develop brach with ADempiere

e-Evolution:
#1608 Enhancement Update IntelliJ and Scala setting
#1609 Enhancement Move Scala Testing for Test Scala Directory
#1553 HHRR: New SmartBrowser to generate invoices for accumulated
-Fix Script Migration
#1560 Commission for complete payments: commissions from past months
-Fix Script Migration
#1564 Add Project Model Validator to Catch multiples Events
#1580 The enhancement allows defining the payment document type into the document type for pay selection.
#1595 Enable migration script for ADempiere 3.9.0.a
#1415 Barcode Scanner Products Not is Functionality whit Attribute
#1294 Add to support ZK Web debugging using IntelliJ
#1499 Defining two Standard Request Type associated


yamelsenih:
Hi @e-Evolution i see a changes for XML that already exist, this changes just remove some steps of XML, i want to know if the removed steps is created in other XML for 3.9.1 or exist other way for resolve it.

Best Regard

e-Evolution:
Hi @yamelsenih, you are in rightly, I removed this columns for column view that no are showed and used , I suppose that was generate when the smart browser was created, but not should exist dependences between improves given the should be agnostic, So 2 columns that was added in other XML 03630_1552_HHRR_FlagPostIndividually.xml was infected into of 03640_1553_SB_HHRR_CreateInvoice_View.xml , as good practice any improves not should infect with changes from other improves , each improve should be agnostic the other.

yamelsenih:
I think that it is ok, but i need know if it will be added in other XML or exist a method for keep the original changes in a upgrade

e-Evolution:
@yamelsenih the two the view columns that were deleted do not affect the operation of the new smart browser since these columns are not displayed in the smart browser, so the change does not represent a risk.

marcalwestf:
Reviewed changes for  issu https://github.com/adempiere/adempiere/issues/1560. No objection this issue concerned.

PR #1639, opened 2018-04-04
Process complete and Payment correction

SusanneCalderon:
New files: 
   InvoiceCompletePayAndPrintAbstract
   InvoiceCompletePayAndPrint
Changed file
   MPaySelectionCheck (aggregate paymentrule cash for documentprint)

### DIFF
--- a/base/src/org/adempiere/process/InvoiceCompleteAndPay.java
+++ b/base/src/org/adempiere/process/InvoiceCompleteAndPay.java
@@ -1,0 +1,92 @@
+/******************************************************************************
+ * Product: ADempiere ERP & CRM Smart Business Solution                       *
+ * Copyright (C) 2006-2017 ADempiere Foundation, All Rights Reserved.         *
+ * This program is free software, you can redistribute it and/or modify it    *
+ * under the terms version 2 of the GNU General Public License as published   *
+ * or (at your option) any later version.										*
+ * by the Free Software Foundation. This program is distributed in the hope   *
+ * that it will be useful, but WITHOUT ANY WARRANTY, without even the implied *
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.           *
+ * See the GNU General Public License for more details.                       *
+ * You should have received a copy of the GNU General Public License along    *
+ * with this program, if not, write to the Free Software Foundation, Inc.,    *
+ * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.                     *
+ * For the text or an alternative of this public license, you may reach us    *
+ * or via info@adempiere.net or http://www.adempiere.net/license.html         *
+ *****************************************************************************/
+
+package org.adempiere.process;
+
+import org.compiere.model.*;
+import org.compiere.print.ReportCtl;
+
+/** Generated Process for (Invoice Complete And Pay)
+ *  @author ADempiere (generated) 
+ *  @version Release 3.9.1
+ */
+public class InvoiceCompleteAndPay extends InvoiceCompleteAndPayAbstract
+
+{
+	@Override
+	protected void prepare()
+	{
+		super.prepare();
+		if (getCheckNo().length()<=0){
+            X_C_BankAccountDoc cBankAccountDoc = new X_C_BankAccountDoc(getCtx(),getBankAccountDocId(), get_TrxName());
+            Integer checkNo = cBankAccountDoc.getCurrentNext();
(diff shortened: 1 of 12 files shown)


SusanneCalderon:
Funcionality of the process
 There exist invoices which are paid inmediately. This process combines complete, pay and pcheck print.
So the user does not have to change windows to pay the invoice and print the check.

**Parameters**:
- Bankaccount
- PaymentRule
- Checkno
- isprinted
- isdirectprint

**Steps**
1. Generate an invoice
2. Complete the invoice 
3. Create a Payment with the same amount and reference to the invoice
4. Complete the payment
5. Print the check.

wdtz:
**Test results:**
**Observation  1**
The currency is copied from the invoice, causes an error when the invoice is in a currency different from the bank account with which it will be canceled:
**Case:**

- An invoice of EUR 20
- A Payment from a USD account

**Expected Result:**
I had to make the payment in USD and convert the amount.

**The result:**
Generated a payment with EUR currency from a USD bank account

**Observation  2**
The sequence of the bank account document is not updated.

**Improvement:**

- The partial payment of the invoice should be considered.
- The following parameters should be considered:
     - Document type
     - Description

-

SusanneCalderon:
Hi @wdtz , thank you. I will work on your comments now.

SusanneCalderon:
Modifications:
- New Parameters:

- - Documenttype
- - Conversion Type
- - Payamount
- - Description

- Restrict the process to invoices without discount and/or payment schedule.
- Permit parcial payment

yamelsenih:
Hi @SusanneCalderon i think that complete process can be used own workflow and button of window.

The user can complete document and after complete it, can run process for generate payment.

1. Complete using **DocAction** button.
2. Run Process for generate Payment.

yamelsenih:
Hi @SusanneCalderon please take my comments for a proposal of changes

SusanneCalderon:
Hi @yamelsenih. 
We combined complete and pay in one botton because our clients were complaining about so many steps and clicks for simple invoices.
I suggest the following changes
**In code line 57**
`
        //if (!(invoice.getDocStatus().equals("DR") || invoice.getDocStatus().equals("IP")))
		//	return "";
        if ((invoice.getDocStatus().equals("DR") || invoice.getDocStatus().equals("IP"))){
            if (!invoice.processIt("CO"))
                return "Could not complete Invoice";
            invoice.saveEx();
        }
`
**In the definition of the field**
By default the button is obly visible for completed or closed documents. You can change the visiblity in certain installations to _Drafted_ or _In progess_ if the user wants to simplify the invoice process.

e-Evolution:
@SusanneCalderon 
I think this functionality is not necessary since there is one that does exactly the same SB **Generate payment selection (from invoice)**, create something similar is only duplicating code and is not efficient.

I recommend adding a process that calls the SB Generate payment selection (from invoice) including the smart browser with same name and attached to the table C_Invoice, modify the smart browser field of C_Invoice_ID and add the default as @ C_Invoice_ID @ to when calling the SB the parameter is filled with the invoice that is in context, after that you only have to select invoice and put the type of document and process the payment.

tip: you can save a step if set the Smart Browser as Is selected by default in yes, or create a copy of original smart browser, also include the [patch for default for field marked as is key](https://github.com/adempiere/adempiere/pull/1865) .  other good improve may be adding a new flag for pay selection document type to generate payment without print check avoid so the print of checks and generate the payment.



Here can see the result:

payselection

yamelsenih:
I agree with @e-Evolution, but is nice if a flag is added for Payment selection document type, i flag can be used for generate payments after complete payment selection, then the process step by step is:

- Complete invoice
- Run SB from process for generate payment selection (when it is generated, automatically is generated payments)

SusanneCalderon:
Hi @yamelsenih , @e-Evolution 
The process is ment for small invoices which are paid inmediately and the user doesn't want to use another button. If you use the smart browser which creates the pay selection you still have to leave the window and print the payselection with the special form.

yamelsenih:
> Hi @SusanneCalderon i think that complete process can be used own workflow and button of window.
> 
> The user can complete document and after complete it, can run process for generate payment.
> 
> 1. Complete using **DocAction** button.
> 2. Run Process for generate Payment.

Please see it.

I think that workflow should be kept.

yamelsenih:
Hi @SusanneCalderon can you resolve this conflict witho trunk? I think that can be add with some corrections but I need a complete environment without conflicts.

Best regards

yamelsenih:
I will close this meanwhile is correct

Issue #1648, opened 2018-04-09
Smart Browser Create from Invoice in pay selection does not consider the cash discount


SusanneCalderon:
The payamt in the smartbrowser is the open amount and does not consider a discount like _2% discount if paid in 10 days_
![captura de pantalla 162](https://user-images.githubusercontent.com/15349639/38508360-3d9dba4a-3bdc-11e8-90db-b1e0ab69c5dd.png)

Solution: 
replace the view column payamt _currencyConvert(invoiceOpen(inv.C_Invoice_ID, inv.C_InvoicePaySchedule_ID), inv.C_Currency_ID, @INV_C_Currency_ID@, TO_DATE('@PayDate@','YYYY-MM-DD'), inv.C_ConversionType_ID, inv.AD_Client_ID, inv.AD_Org_ID)_
with 
_currencyConvert(invoiceOpen(inv.C_Invoice_ID, inv.C_InvoicePaySchedule_ID), inv.C_Currency_ID, @INV_C_Currency_ID@, TO_DATE('@PayDate@','YYYY-MM-DD'), inv.C_ConversionType_ID, inv.AD_Client_ID, inv.AD_Org_ID) - 
currencyConvert(paymentTermDiscount(inv.GrandTotal, inv.C_Currency_ID, inv.C_PaymentTerm_ID, inv.DateInvoiced, TO_DATE('@PayDate@','YYYY-MM-DD')), inv.C_Currency_ID, @INV_C_Currency_ID@, TO_DATE('@PayDate@','YYYY-MM-DD'), inv.C_ConversionType_ID, inv.AD_Client_ID, inv.AD_Org_ID)_

yamelsenih:
Please continue discussione here: https://github.com/adempiere/adempiere/pull/1650

PR #1650, opened 2018-04-10
adempiere: #1648 Smart Browser Create from Invoice in pay selection d…

SusanneCalderon:
…oes not consider the cash discount

https://github.com/adempiere/adempiere/issues/1648
SeqNo for xml: 3800

### DIFF
--- a/migration/390lts-391/03800_1648_CorrectionSBCreatePaySelectionFromInvoice.xml
+++ b/migration/390lts-391/03800_1648_CorrectionSBCreatePaySelectionFromInvoice.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<Migrations>
+  <Migration EntityType="U" Name="1648_CorrectionSBCreatePayselectionFromInvoice" ReleaseNo="3.9.1" SeqNo="3800">
+    <Comments>https://github.com/adempiere/adempiere/issues/1648</Comments>
+    <Step SeqNo="10" StepType="AD">
+      <PO AD_Table_ID="53232" Action="U" Record_ID="54229" Table="AD_View_Column">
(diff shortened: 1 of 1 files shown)


beruskajimenez:
Hi @SusanneCalderon , with regard to the 1650, could you tell me exactly what is the name of the Smart Browser? and if it is possible to better define how the process works. Can you put a video of the process here?Thank you!

SusanneCalderon:
Hi @beruskajimenez . 
Here is an upload with the error.
[Payselection.zip](https://github.com/adempiere/adempiere/files/2100126/Payselection.zip)
The amount of the invoice is $5000, the payment amount is $4900 due to a 2% discount for payment in the first 10 days. 
The name of the SB is PaySelectionCreateFromInvoice, the corrected field is payamt.

SusanneCalderon:
Hi @beruskajimenez , the problem is that you get an save error when payamt > openamt.
That was the reason to consider it as bug.

yamelsenih:
Hi, any result?

marcalwestf:
hi,
reviewed the issue and reproduced the events that Susanne recorded in the attached video.

Without the changes, an error is generated and the pay selection process is aborted; with the changes, the error is not generated and the pay selection completes.

@beruskajimenez acknowleged the issue, and suggests the changes to be considered an improvement.  Thus it seems to be a question of semantics if the issue is labeled or not as "bug". Nevertheless, the changes solve the case.

SusanneCalderon:
imagen

SusanneCalderon:
On the status bar you can see the error _Payment amount>Invoice Open Amount_

e-Evolution:
@beruskajimenez 

about of  

**however, could be considered an improvement that the total Open Amount column is the subtraction between the Grand Total minus the Discount Amount**

The Open amount is rightly because the discount for prompt payment is only granted if the payment is made, so the open amount is correct and depend of pay .

Issue #1113, opened 2017-06-13
Bugfix the ZK not work jasper report included in print format.


e-Evolution:
1.- Created a print report process based on jasper report
2.- Create a new print format and link with report process based on jasper
3.- The Jasper report use Record_ID as parameter to print of report.
4.- Select the report of print format list

The current issue is that the process dialog is show and not execute of report.

e-Evolution:
The process dialog is show when no exist a parameter and if is executed not get the report

![i431 pimgpsh_fullsize_distr](https://user-images.githubusercontent.com/786968/27096250-02f22f30-5036-11e7-82d6-55d9e12eeaec.png)

e-Evolution:
Issue related Process Dialog is not MVC #571

marcalwestf:
The class _org.adempiere.webui.apps.WProcessCtl_ is implemented now in such a way that a Jasper Report gets **never** executed. See code:
![20180412_bug_01](https://user-images.githubusercontent.com/1789408/38659474-ff63d900-3de6-11e8-93cb-191084379d82.png)

As can easily be seen, the execution of the process (via calling _processModalDialog.runProcess()_ ) is never reached.

This bug is solved by eliminating the _else_ part:
```
ProcessModalDialog processModalDialog = new ProcessModalDialog(asyncProcess, windowNo, pi);
		if (processModalDialog.isValidDialog())
		{
			processModalDialog.setWidth("500px");
			processModalDialog.setVisible(true);
			processModalDialog.setPosition("center");
			processModalDialog.setAttribute(Window.MODE_KEY, Window.MODE_MODAL);
			AEnv.showWindow(processModalDialog);
			if (!processModalDialog.isOK()) {
				return;
			}
		}
		processModalDialog.runProcess();
		
		return;
```

PR #1653, opened 2018-04-12
 #1113 : The ZK not work jasper report included in print format.

marcalwestf:
See issue https://github.com/adempiere/adempiere/issues/1113
The reports were never executed; with the change, they are.

### DIFF
--- a/zkwebui/WEB-INF/src/org/adempiere/webui/apps/WProcessCtl.java
+++ b/zkwebui/WEB-INF/src/org/adempiere/webui/apps/WProcessCtl.java
@@ -104,8 +104,7 @@ public class WProcessCtl {
 				return;
 			}
 		}
-		else
-			processModalDialog.runProcess();
+		processModalDialog.runProcess();
 		return;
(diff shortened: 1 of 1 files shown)


yamelsenih:
Hi @e-Evolution can you be more specific for it?

I see issue #1113 but i don't know what is the step by step. Can you please put here a gif or image of problem?

Best regards

e-Evolution:
@yamelsenih , I set in #1653 the step to replicate , this functionality is used when use a print format based on jasper report linking to process defined as report and attachment jasper report.

Here an example.
captura de pantalla 2018-06-13 a la s 15 05 05

yamelsenih:
Thanks @e-Evolution, have you a print shot from user window calling report?

e-Evolution:
@yamelsenih , yes the called is from icon Report , that allow generate a report in short dynamically .

SusanneCalderon:
Please ignore my commits, they are from a branch which is already deleted.

Issue #1656, opened 2018-04-13
New Asset Split Document 


e-Evolution:
The new functionality Fixed Asset Split document allows split a Fixed Asset by type (Amount , Percentage and Quantity), this new document will create accountants facts  and depreciation for each new Fixed Asset

e-Evolution:
New window Fixed Asset Split:

![splitfixedasset](https://user-images.githubusercontent.com/786968/38830425-df8b7ffe-4181-11e8-9c15-e6aedc241ea2.gif)

e-Evolution:
I am creating the functionality for the split of fixed assets, that is, a fixed asset that is consolidated and that is to be individually split.

Even with this, I am considering being able to split based on percentage, Quantity or Amount, Thus I have created a smart browser to be able to select any fixed asset that can be split with the possibility that you can split when selecting the fixed asset a document called split fixed assets will be created, this document has all the information to split the fixed asset, as quantity, percentage or the amount, my doubt now is,
How to proceed when completing document?
The first option is the disposal of the original fixed asset and creates additions of new assets, this would imply that the document of split fixed assets generates other documents of fixed assets such as the fixed asset disposal and fixed asset addition. So I am not sure what best approach since workflow for the document could not be completed until the others are complete which is somewhat complex to control, on the other hand, the accounting part is transparent because when disposal the fixed asset and the accounting part is clearly determined. given that the amounts and amounts would be taken from the asset allocation document and the history of the original fixed asset
The other approach is not to create additions or disposal, allowing the split fixed asset document to generate the new fixed assets and deactivate the consolidated asset, so the fixed asset split document itself will have its own accounting rules, which will allow implementing the behavior required.
The only problem in the second approach is that a fixed asset is not created from an addition as is the normal flow, and the decline of the consolidated asset in a fixed asset disposition document.

What do you think that should be of the right approach to be implemented?

adaxa-steven:
I would process as...
dr new asset 1
cr depreciation reserve new asset 1
cr transfer suspense

db new asset 2
cr depreciation reserve new asset 2
cr transfer suspense
 
db new asset 3
cr depreciation reserve new asset 3
cr transfer suspense

Cr old asset 
db depreciation reserve old asset
dr transfer suspense

it would probably require a new system account 'transfer suspense' .. this would be preferable to posting to the disposal account and probably confusing the values needed in the cash flow statement 

regards...

marcalwestf:
[...]

PR #1673, opened 2018-04-30
Feature/1656 asset split document

e-Evolution:
Asset Split Document see detail #1656


yamelsenih:
I test it applying XML and running build and is OK

PR #1716, opened 2018-06-05
adempiere:  #1370 Enhance the process AllocationAuto with allocation …

SusanneCalderon:
…of prepayments

https://github.com/adempiere/adempiere/issues/1370
Correction of the XML 03690_1370


wdtz:
**Test results:**

- When making an invoice with the total quantities of the purchase order the result is successful (the assignment is made)
- When making an invoice with the partial quantities of the purchase order the result is unsuccessful (The assignment is not made)

**Improvement:**

- The  Flag Auto Assignment should go in the type of document

Issue #1730, opened 2018-06-08
Process as silent run show parameters


yamelsenih:
When a process is flagged as silent run, itshow parameter when ios called from a window, also the busy dialog is not showed. But if is called from menu the behavior is standard. See definition Step by Step:
 - Create Process with parameters
![screenshot_20180608_144146](https://user-images.githubusercontent.com/2333092/41175350-65c27d24-6b22-11e8-97ed-3f115344606a.png)
 - Add Process to Table
![screenshot_20180608_144308](https://user-images.githubusercontent.com/2333092/41175387-87505394-6b22-11e8-8891-a020e030484c.png)
 - Open window
![screenshot_20180608_144340](https://user-images.githubusercontent.com/2333092/41175433-a7576cb8-6b22-11e8-9770-dc6d7893cb0d.png)
 - Run Process
![screenshot_20180608_144404](https://user-images.githubusercontent.com/2333092/41175441-ad93b2da-6b22-11e8-9007-3655e8f7a8dc.png)

yamelsenih:
It is referenced on pull request #1731

wdtz:
**Result Test:**

Successful

mckayERP:
Develop branch.

The ZK process dialog is still showing when there are no parameter or "don't show help" or "run silently" are selected.  The fix from swing needs to be ported to ZK.

Specifically, the lines from the client org.compiere.apps.APanel.java 2702 to 2707 need to be ported to across to org.adempiere.webui.panel.AbstractADWindowPanel.java

APanel in swing (running a process from a button click - starting 2696

		else {
			ProcessModalDialog dialog = new ProcessModalDialog(ctx, Env.getWindow(m_curWindowNo), Env.getHeader(ctx, m_curWindowNo),
					this, m_curWindowNo, vButton.getProcess_ID(), table_ID,
					record_ID, startWOasking);
			//	FR [ 265 ]
			//	BR [ 323 ]
			if(!startWOasking
					&& !dialog.isAutoStart()) {
				dialog.validate();
				dialog.pack();
				AEnv.showCenterWindow(Env.getWindow(m_curWindowNo), dialog);
			}


The equivalent lines in ZK org.adempiere.webui.panel.AbstractADWindowPanel.java starting 2264 are missing the if statement.  As a result, the parameter panel is always displayed.

		else
		{
			ProcessModalDialog dialog = new ProcessModalDialog(this, curWindowNo,
[...]

PR #1731, opened 2018-06-08
#1730, Fixed running as silent process with modal dialog for process

yamelsenih:
clled from window, reference to issue:
https://github.com/adempiere/adempiere/issues/1730

Please review it pull request, currently exist a bug when a process is executed from a window and it is configured as silent process

Best Regard


e-Evolution:
@yamelsenih , Hi Yamel I only saw that swing class was modify , so than have no problem for ZK implementation?

yamelsenih:
Hi @e-Evolution you are right, ZK is the same error, i will to change it and make commit.

yamelsenih:
screenshot_20180608_181436

yamelsenih:
> @yamelsenih , Hi Yamel I only saw that swing class was modify , so than have no problem for ZK implementation?

It is solved yet

Issue #1603, opened 2018-03-21
Bugfix when an active is generate as expense the active not is created


e-Evolution:
When an invoice line for expende fixed assed is selected, not allow create of fixed asset and the fixed assed addition is incorrect.

Steps to business case:

1.- Create new Asset Group 
<img width="1628" alt="captura de pantalla 2018-06-20 a la s 17 24 56" src="https://user-images.githubusercontent.com/786968/41882431-704e17fa-78ae-11e8-8d99-2d5af0e1ec28.png">

2.- Define Asset Group Account 
<img width="1621" alt="captura de pantalla 2018-06-20 a la s 17 28 50" src="https://user-images.githubusercontent.com/786968/41882455-9501f990-78ae-11e8-8de1-df053f0aa2b3.png">

3.- Asset Group uncheck of One Asset Per UOM and Depreciate
<img width="1621" alt="captura de pantalla 2018-06-20 a la s 17 29 36" src="https://user-images.githubusercontent.com/786968/41882498-d5e65bd6-78ae-11e8-9772-b69499aa0c2b.png">

4.- Create a new product category and link with Asset Group
<img width="1620" alt="captura de pantalla 2018-06-20 a la s 17 31 48" src="https://user-images.githubusercontent.com/786968/41882522-fc935ab8-78ae-11e8-98df-aee591cd3337.png">

5.- Create a new Asset Product and set new Product Category 
<img width="1619" alt="captura de pantalla 2018-06-20 a la s 17 34 39" src="https://user-images.githubusercontent.com/786968/41882544-1e27c86c-78af-11e8-910c-e44ff48457f3.png">

6.-Define the product price 
<img width="1609" alt="captura de pantalla 2018-06-20 a la s 19 46 53" src="https://user-images.githubusercontent.com/786968/41882682-ed0d1e16-78af-11e8-954e-7e16818ce9f9.png">

7.- Create a new Purchase Order and Purchase Order Line with new Product Asset
<img width="1621" alt="captura de pantalla 2018-06-20 a la s 19 42 49" src="https://user-images.githubusercontent.com/786968/41882636-a77c7c7a-78af-11e8-897e-53b5dddd5995.png">

<img width="1620" alt="captura de pantalla 2018-06-20 a la s 19 43 09" src="https://user-images.githubusercontent.com/786968/41882656-c92f8d58-78af-11e8-85fa-9b13d1f6039f.png">

8.- Create a new Material Receipt and completed based on  Purchase Order
[...]

PR #1746, opened 2018-06-12
 #1603 Bugfix when an active is generate as expense the active not is…

e-Evolution:
… created  https://github.com/adempiere/adempiere/issues/1603

- Now when an invoice have a fixed asset line as expense is posible adding the asset only as control and tracking

(cherry picked from commit b6de1ce39748f8b26940a06b685c8a36485d1345)

### DIFF
--- a/org.adempiere.asset/src/main/java/base/org/compiere/FA/model/ModelValidator.java
+++ b/org.adempiere.asset/src/main/java/base/org/compiere/FA/model/ModelValidator.java
@@ -467,10 +467,6 @@ implements org.compiere.model.ModelValidator, org.compiere.model.FactsValidator
 			}
 		
 			//Expense Asset_ID check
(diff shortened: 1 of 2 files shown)


RJPC:
Hi, the pull request is already under review

wdtz:
Hello @e-Evolution, I think it convenient to make a change, the definition of the following points of interest for functional tests:

- Benefit of Change
- Previous operation procedure
- New operation procedure (Change)

In this way, it will be easier to understand the proposed change, thus giving an opportunity to give an opinion and evaluate its optimal functioning.

e-Evolution:
Hi @wdtz Waditza, 

- The benefit of Change:  allows generated an addition of an asset fixed that not is depreciated to have a control of any fixed asset.

- Previous operation procedure: not allow create any Fixed Asset not depreciated and Fixed Asset Addition.

-  New operation procedure (Change): see the steps list for issue #1603

e-Evolution:
fixes

Issue #1690, opened 2018-05-15
Bug with database record lock when enabling accounting immediate is enabled.


e-Evolution:
The problem is caused by get an accounting valid combination without db transaction when should use the original db transaction of a document when is completed, this not is problem when the accounting is not immediately, given that  accounting create a new transaction in this moment.

PR #1756, opened 2018-06-12
1690 Bug with database record lock when enabling accounting immediate is enabled.

e-Evolution:
Bug with database record lock when enabling accounting immediate is enabled. #1690 https://github.com/adempiere/adempiere/issues/1690

### DIFF
--- a/base/src/org/compiere/acct/Doc.java
+++ b/base/src/org/compiere/acct/Doc.java
@@ -1315,17 +1315,17 @@ public abstract class Doc
 
 	/**
 	 *	Get the Valid Combination id for Accounting Schema
-	 *  @param AcctType see ACCTTYPE_*
-	 *  @param as accounting schema
+	 *  @param acctType see ACCTTYPE_*
+	 *  @param acctSchema accounting schema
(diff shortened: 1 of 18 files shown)


yamelsenih:
It have syntax error:
screenshot_20180711_214946

marcalwestf:
I would strongly reccomend not to mix naming refactoring changes with functional changes in the same issue.

Example of naming refactoring:  https://github.com/adempiere/adempiere/pull/1756/commits/2eeb8f5a225a056c61ac4b23523a197a3903a056

It would be better if there is one issue for refactorting names and another for functional changes, even if they apply to the same class.

When a large class is involved, this improves checking and testing.

marcalwestf:
Moreover, I would be grateful if you could explain step by step how to reproduce the error, this means under which circumstances an immediate accounting locks a table.

I didn't understand it well by the explanation given in https://github.com/adempiere/adempiere/issues/1690 .
Thanks

yamelsenih:
Hi Víctor, exist a path for solve it?

Issue #1708, opened 2018-05-30
Enhancement Payroll causing lock record when execute prepare Payroll process


e-Evolution:
When a Payroll process is completed this executing the workflow step fro this document, first prepare it document and continue to complete, when the complete step is executed, the record never is released.

yamelsenih:
Hi @e-Evolution it is a bug introduced with 3.9.0 release

e-Evolution:
@yamelsenih , I am reviewing all code of Payroll and unfortunately there are many poorly defined methods given that a static method must always receive as parameter the context and name of the transaction, so change it as improvement and fix the theme of the methods based on the best practice even if it is a bit painful, I will use this improvement to solve this issue.

PR #1762, opened 2018-06-12
#1708 Enhancement Payroll causing lock record when execute prepare Payroll process

e-Evolution:
Enhancement Payroll causing lock record when execute prepare Payroll process #1708 https://github.com/adempiere/adempiere/issues/1708

### DIFF
--- a/base/src/org/compiere/model/MActivity.java
+++ b/base/src/org/compiere/model/MActivity.java
@@ -52,19 +52,21 @@ public class MActivity extends X_C_Activity
 	 * @param activityId
 	 * @return
 	 */
+	/*
 	@Deprecated
 	public static MActivity get(Properties ctx, int activityId)
 	{
-		return getById(ctx, activityId);
-	}
+		return getById(ctx, activityId, null);
+	}*/
 
 	/**
 	 * Get/Load Activity [CACHED]
 	 * @param ctx context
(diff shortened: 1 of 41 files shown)


yamelsenih:
Hi
i will review it pull request.

Some info: it bug was added on 3.9.0 release

yamelsenih:
Hi @e-Evolution i see that exist many changes with method without trxName, have you old compatibility? exist many implementation of MActivity and other class for custom.

yamelsenih:
Old implementation
<pre>
	@Deprecated
	public static MActivity get(Properties ctx, int activityId)
	{
		return getById(ctx, activityId);
	}
</pre>
It is removed
<pre>
	/*
	@Deprecated
	public static MActivity get(Properties ctx, int activityId)
	{
		return getById(ctx, activityId, null);
	}*/
</pre>

It should be keep back

e-Evolution:
@yamelsenih I fixed the back compatibility method, as this is an enhancement, unfortunately, multiple java files are affected to support completely the transaction management,   fortunately, this fixed the transaction locking for payroll process

Please try compile our old customization to check back compatibility, if you found any problem please notice

yamelsenih:
hi @e-Evolution i will test it.

Best regard

Issue #1722, opened 2018-06-07
Enhancement Payroll Send Mail when trying to send the document payroll payment voucher


e-Evolution:
The process Payroll Send Mail should find a payment for the employee , generate from a payroll process  and obtain the document of the receipt of payment and send by email to the employee.

yamelsenih:
Please add a Step by Step. Also is nice to have screenshots with new functionality

PR #1768, opened 2018-06-12
#1722 Bug Payroll Send Mail when trying to send the document payroll …

e-Evolution:
…payment voucher https://github.com/adempiere/adempiere/issues/1722

- Rename the process  with the best name
- Now using the payment as entity base to send the payroll payment voucher report.

(cherry picked from commit d76a3b4a80e1f89daa725610caa30ee46be0a7e0)


SusanneCalderon:
Hi @e-Evolution , we use the former process payrollViaEmail in an instalation as an advanced email report to the employees, before completing the HR_Process. So they have 5 days to control the forthcoming payroll.
Can we use this new process the same way?

- The payroll is in status In process
- There are no payments realized

SusanneCalderon:
Hi @e-Evolution , we examined the code and it seems to be a entirely new funcionality. The process PayrollViaEmail, which is deleted in this PR, has as a starting point the employees, the new process has as a starting point a payment.
So this pull request violates backward compatibility.

The (old) process PayrollViaEmail can generate a pro forma report of payroll process which are not paid.
It can be used as a general report for the employees to control the composition of the payroll or as an anouncement of a forthcoming payroll. This works well. It creates an individual mail to every employee who is included in the selected HR_Process with a individual report as attachment.

So as the focus of the new process is entirely different I suggest to add this as a new process and leave the process PayrollViaEmail as it is.

yamelsenih:
Hi @e-Evolution can you reply @SusanneCalderon comment?

Best Regard

e-Evolution:
@yamelsenih , yes I think that is best create other process, I will create new process

yamelsenih:
Any update?

yamelsenih:
I will reject it pull request, @e-Evolution please when you have new change for new process make a new pull request.

Best regard.

Issue #1741, opened 2018-06-12
Deleted recent items causes NPE (only in Swing)


marcalwestf:
When for one reason or another, there is an entry in Recent Items (table AD_RecentItem) which points to a non-existing record, an NPE is generated when during the construction of the Recent Items Tree.

Example: in the recent items table there is a Record (ID==1000044) for table M_Product (ID==208).
![20180612_quicksheet_01](https://user-images.githubusercontent.com/1789408/41282388-2db48f04-6df1-11e8-939b-c79d1989b27f.png)

Unluckily, in the M_Product table, there's no record with this ID (1000044)
![20180612_quicksheet_02](https://user-images.githubusercontent.com/1789408/41282411-3c49e30c-6df1-11e8-84d4-5e459c385728.png)

This leads to a NPE:

> ADempiere starting: org.compiere.apps.AMenu - java.lang.NullPointerException
> java.lang.NullPointerException
>         at org.compiere.model.MRecentItem.lambda$getLabel$0(MRecentItem.java:410)
>         at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:174)
>         at java.util.stream.SortedOps$SizedRefSortingSink.end(SortedOps.java:352)
>         at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
>         at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
>         at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
>         at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
>         at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
>         at java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:418)
>         at org.compiere.model.MRecentItem.getLabel(MRecentItem.java:412)
>         at org.compiere.grid.tree.VTreePanel.addToBar(VTreePanel.java:1051)
>         at org.compiere.grid.tree.VTreePanel.initTree(VTreePanel.java:349)
>         at org.compiere.apps.AMenu.<init>(AMenu.java:164)

**The only way to solve it right now is by deleting the Recent item entry, which is first difficult to find out. Moreover, Adempiere doesn't even start!**

The reason: in the class MRecentItem, the instantiation of the MProduct with the ID yields -as expected- a null object:
[...]

Issue #1426, opened 2017-10-28
Null Pointer Exception at Adempiere start


mbar49:
I imported the products, then deleted the i_product table with the Delete Import function.
At Adempiere restart before menu window I got:

09:54:55.167 DB.isDatabaseOK: DB_Version=2017-08-01 [1]
09:54:55.579 MRole.get: AD_Role_ID=102, AD_User_ID=100, reload=false [1]
09:54:55.850 WFPanel.<init>: RW=false [1]
09:54:59.074 Env.getImageIcon2: GIF/PNG Not found: ShowAllWindow16 [1]
09:54:59.075 Env.getImageIcon2: GIF/PNG Not found: ShowAllWindow24 [1]
09:54:59.180 MTree.<init>: AD_Tree_ID=10, AD_User_ID=100, Editable=false, OnClient=true [1]
-----------> X_I_Product.load: NO Data found for I_Product_ID=1001787 [1]
ADempiere starting: org.compiere.apps.AMenu - java.lang.NullPointerException
java.lang.NullPointerException
	at org.compiere.model.MRecentItem.lambda$getLabel$0(MRecentItem.java:410)
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:174)
	at java.util.stream.SortedOps$SizedRefSortingSink.end(SortedOps.java:352)_

yamelsenih:
Hi @mbar49 is possible that the reference to products was deleted and recent item have a broker link, I will fix it.

Thanks you for report it.

yamelsenih:
Hi @mbar49 it was solved yet, i test it easy:

1. Open ADempiere.
2. Create a new Payment Term by example (It generate a ID 1000000).
3. After it i run it query: <pre>DELETE FROM C_PaymentTerm WHERE C_PaymentTerm_ID=1000000;</pre>
4. I Open newsly ADempiere and it throw a NPE.

PR #1778, opened 2018-06-14
#1741: Deleted recent items causes NPE (only in Swing), reference to

marcalwestf:
#1741: Deleted recent items causes NPE (only in Swing), reference to https://github.com/adempiere/adempiere/issues/1741

Improvement from
#1426: Null Pointer Exception at Adempiere start, reference to https://github.com/adempiere/adempiere/issues/1426

### DIFF
--- a/base/src/org/compiere/model/MRecentItem.java
+++ b/base/src/org/compiere/model/MRecentItem.java
@@ -405,7 +405,7 @@ public class MRecentItem extends X_AD_RecentItem
 			PO po = table.getPO(getRecord_ID(), null);
 			//	Validate PO
 			if(po == null) {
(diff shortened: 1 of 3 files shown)


e-Evolution:
@marcalwestf reviewed your code and found next suggestion:

if you modify a swing class [VTreePanel.java](https://github.com/adempiere/adempiere/pull/1778/files#diff-64a71b1bf96843e40383773191e652bd) then I think that same bug are into WTreePanel.java or ZK implementation.

yamelsenih:
Hi @e-Evolution VTreePanel is used for wing UI but Recent Item implementation for ZK UI is from DRecentItem dashboard

Issue #1790, opened 2018-06-28
Quick Info for Fields


yamelsenih:
Is nice to have a quick info about a field added to help, it can help to user to know very usefull info about a table or field.

For use it is required use functionality for context info, see it: #1784

yamelsenih:
A little change for ER-Diagram and it is worked very fine, see:
![dynamic-statusbar](https://user-images.githubusercontent.com/2333092/42057828-25c789fe-7ae4-11e8-9526-760ed2a2d3f8.png)

yamelsenih:
A simple Step by Step for configure it:
1- **Create new Context info value don forget that for create a new context info you must create a message before it**:
![screenshot_20180628_150232](https://user-images.githubusercontent.com/2333092/42057900-5ebbd42c-7ae4-11e8-9a3c-b44ad7feffa2.png)
2- **Link the contecxt info to a field**:
![screenshot_20180628_160524](https://user-images.githubusercontent.com/2333092/42058067-cb544fa6-7ae4-11e8-88cb-ee3a9d50643e.png)
3- **See result on your window and field**:
![screenshot_20180628_155002](https://user-images.githubusercontent.com/2333092/42058094-e8edb002-7ae4-11e8-8599-038a866ced15.png)

Niceee

yamelsenih:
Hi, it already exists in https://github.com/adempiere/adempiere/pull/1791 pull

Continue discussion on pull

Best Regard

Issue #1784, opened 2018-06-20
Add dynamic status bar info


yamelsenih:
Currently ADempiere has a Status Baf for all windows, it can be used for show information about process status like Save, New, Processing and other.

For windows like:
  - Sales Order
  - Purchase Order
  - Sales Invoice
  - Invoice Vendor
  - Bank Statement
  - Delivere
  - Receipt
  - RMA Customer
  - RMA Vendor

The information is more specific, however, for all window the information is very half.
![screenshot_20180620_164034](https://user-images.githubusercontent.com/2333092/41684283-a3b338f6-74a2-11e8-9137-5aa007f22195.png)

What happen if i want add more information for my custom window? :-1: 
It is not possible.

Here a example of hardcode :-1: 
![screenshot_20180620_171126](https://user-images.githubusercontent.com/2333092/41684942-a22e74f8-74a4-11e8-8890-6e19fa65ec5b.png)

yamelsenih:
I add a ER-Diagram of feature
![dynamic-statusbar](https://user-images.githubusercontent.com/2333092/41828317-32a4980c-77fa-11e8-8658-ba9dd73ff3e3.png)

yamelsenih:
Here a Step by Step:

- Now you can see that already exists a window named **Status Bar**
![screenshot_20180625_112011](https://user-images.githubusercontent.com/2333092/41859789-9487ddb4-7862-11e8-8c01-22bdc8d6529f.png)
- See a example with Order, now you can define status bar info for: **Table**, **Tab** and **Window**
![screenshot_20180625_112044](https://user-images.githubusercontent.com/2333092/41859885-d25afb4e-7862-11e8-85f7-7d3804272785.png)
Note that current status bar appy for **C_Order** Table, **Sales Order** and **Purchase Order** window
![screenshot_20180625_112058](https://user-images.githubusercontent.com/2333092/41859955-03090772-7863-11e8-8c71-f404abfa7ad6.png)
- After define a Status bar, you can link it to Table:
![screenshot_20180625_112130](https://user-images.githubusercontent.com/2333092/41860005-1da13a96-7863-11e8-96b0-ef612b531392.png)
- Also it is available for Window:
![screenshot_20180625_112151](https://user-images.githubusercontent.com/2333092/41860020-2c6daf8c-7863-11e8-8be6-9f3db344ee57.png)
- It each tab has distinct Status Bar then you can define it for **Tab**
[...]

PR #1791, opened 2018-06-28
Feature/#1790 quick info for fields

yamelsenih:
Hi guys, it is a nice change and allows create easy info for any field, and Status Bar.

Please take a time for it.

It include: #1790 and #1784 issues

Best Regard


marcalwestf:
Tested and works OK.
Remarks:
- Status works well on both Swing and ZK
- Quick info for fields works well on ZK, but it doesn't work on Swing
- Quick info for fields is displayed when hovering directly over the field (compared with tooltip help which is displayed when hovering over the caption)

PR #1822, opened 2018-07-18
patch in ProcessUtil.startScriptProcess when error msg startsWith "@Error@"

homebeaver:
this conforms to ad-processes written in java. Error messages have to start with "@Error@".


yamelsenih:
Thank you

Issue #1782, opened 2018-06-20
Cannot complete Planned Manufacturing Order with Movement Date in the future


homebeaver:
in our Accounting schema `Future Days` is defined to 0. This is OK.

I create a new `Planned Manufacturing Order` and set `Movement Date` to `DatePromised` which is in the future. Also this is OK in my opinion.

Now I cannot complete the Document!

![grafik](https://user-images.githubusercontent.com/3493126/41656004-ea8e33c0-748e-11e8-926e-7592babd89bb.png)

You can reproduce it in swing or in a programm:
```java
		MProductionBatch mProductionBatch = new MProductionBatch(getCtx(), 0, get_TrxName());
		
		mProductionBatch.setC_DocType_ID(DOCBASETYPE_MPO_ID); 
		mProductionBatch.setM_Product_ID(mProduct.getM_Product_ID());
		
		mProductionBatch.setMovementDate(this.mOrder.getDatePromised());
...
		mProductionBatch.completeIt();	// <<====== PeriodClosedException
```

e-Evolution:
@homebeaver please , in future use the our convention ion to commits messages:

#Issue Description of Issue URL of issue 

<img width="1263" alt="image" src="https://user-images.githubusercontent.com/786968/44948580-c93af380-ade5-11e8-9fb4-8ec27b14920d.png">

PR #1826, opened 2018-07-19
bugfix for #1782 Movement Date in MProduction

homebeaver:
see https://github.com/adempiere/adempiere/issues/1782

prepareIt: check only if MovementDate is in the past 
by the way I added toString-Method

### DIFF
--- a/base/src/org/compiere/model/MProduction.java
+++ b/base/src/org/compiere/model/MProduction.java
@@ -182,7 +182,7 @@ public class MProduction extends X_M_Production implements DocAction , DocumentR
 				String whereClause = "M_Product_ID=? and M_Locator_ID=? and Movementdate<=? ";
 				ArrayList <Object> params = new ArrayList<>();
 				params.add(product.getM_Product_ID());
-        params.add(line.getM_Locator_ID());
+				params.add(line.getM_Locator_ID());
 				params.add(getMovementDate());
 				if (line.getM_AttributeSetInstance_ID()!=0){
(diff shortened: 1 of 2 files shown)


yamelsenih:
Hi @homebeaver i think that all documents with workflow must be validate with period, it is for:

1. Period control.
2. Fact Account changes.

marcalwestf:
@yamelsenih does your comment mean that -if the commit for this pull request is OK- the actual pull request can be accepted?

Or doest it mean that similar changes must also be done for period control and fact account changes before accepting this pull request?

homebeaver:
Hi @yamelsenih @marcalwestf 
yes you are right. But this is a plan, see screenshot here #1782 - and the planned movement date must be in the future.
When the production is completed, days ore weeks later, the movement date then has to be within the accounting period. 
Do I understand the Manufacturing planned order process wrong?
Regards E.

marcalwestf:
@homebeaver you're right. I will test the pull request as of your explanation.

yamelsenih:
Hi @homebeaver itest it but i find follow:
 - Validation for period for production Order (**M_ProductionBatch**) is removed when it is to future (Ok)
 - Validation for period for production (**M_Production**) is removed when it is to future (It should be keep like previous version, the production change quantity and make account acct).

For code revision i find some improvements:
Please change identation, see images:
screenshot_20180813_142445
screenshot_20180813_142515
screenshot_20180813_142533

yamelsenih:
Please fiux it for merge

yamelsenih:
Testing...

Issue #1088, opened 2017-06-11
Add Adaxa's "Quicksheet" Functionality for massive data input


marcalwestf:
Many customers complain that ADempiere doesn't offer a tool for massive entry. 
Though it is possible to reduce the fields displayed in a window and set the order to navigate through them, the windows are a compromise between displaying and editing.

@adaxa-steven has kindly offered Adaxa's _Quicksheet_ functionality, which overcomes this problem, as it is designed just to allow for (as the name says) quick and massive entry.

The Quicksheet can be configured to be used in any window.

Here are snapshots of Adaxa's document explaining the usage of Quicksheet:
![20170608_quickseheet_01](https://user-images.githubusercontent.com/1789408/27011113-5b945c1e-4e72-11e7-9709-b3346545fb7d.png)
![20170608_quickseheet_02](https://user-images.githubusercontent.com/1789408/27011112-5b940692-4e72-11e7-9a66-a220b2d30856.png)

marcalwestf:
Here a link to the documentation provided by @adaxa-steven : [02-HowTo - Quick Entry Set-up.pdf](https://github.com/adempiere/adempiere/files/1066429/02-HowTo.-.Quick.Entry.Set-up.pdf)

ayazbaloch:
Thanks @marcalwestf  for providing document. Nice functionality.

yamelsenih:
Hi, i add a branch for integrate it: https://github.com/adempiere/adempiere/tree/feature/%231088_Quichsheet_Functionality

Raul-mz:
[#1088 Add keyboard events functionality for grid when is multi record](https://github.com/adempiere/adempiere/commit/b5859b168809063f6fdbf07ec6cfdaa795d8712d)

marcalwestf:
For anyone who's interested, the ongoing integration is on branch feature/#1088_Quichsheet_Functionality

marcalwestf:
.. as of now, the development is almost finished; FYI you can download the code from https://github.com/erpcya/adempiere/commits/feature/%231088_Quichsheet_Functionality

The Quicksheet form is the same as the tab grid, which makes the look and feel more real.

Here a snapshot of the Quicksheet form:
![20180512_quicksheet_03](https://user-images.githubusercontent.com/1789408/41081480-3a899434-69e8-11e8-93d5-a0f54bfb934e.png)

yamelsenih:
Continue it in: https://github.com/adempiere/adempiere/pull/1828

[...]

PR #1828, opened 2018-07-19
feature/#1088_Quick_Sheet

yamelsenih:
See: https://github.com/adempiere/adempiere/issues/1088
Hi everybody, this is a new fuctionality for handle quick sheet.

Please take a time for review it.

Best regard.


marcalwestf:
Hi, I've been ivolved in the testing of the requesting branch.
Now, I'm goin to test the pull request.....

marcalwestf:
The file migration/**380lts-390lts/03000**_1088_Add_Column_IsQuickEntry.xml has to be renamed to migration/**390lts-391/03920**_1088_Add_Column_IsQuickEntry.xml.

03920 is the actual sequence; depending on whether there has been another pull request, it should be eventually changed.

marcalwestf:
... comparing ADTabPanel.java from the requesting repository with Adempiere's repository, the merging conflicts can IMO solved by just adopting the differences.

marcalwestf:
.... accotdingly, the header of _....Add_Column_IsQuickEntry.xml_ file should be changed to
```
<Migration EntityType="D" Name="QuickEntry" ReleaseNo="3.9.1" SeqNo="3920">
 <Comments>See: https://github.com/adempiere/adempiere/issues/1088</Comments>
```

marcalwestf:
... still in the XML file, there must be entries for translations

marcalwestf:
... moreover in the XML file, the View _ad_field_v_ is changed.
It has to be changed also for Oracle and the changes also documented in _ddlutils/../views_.

marcalwestf:
.. again in the XML, the definition of column _isQuickEntry_ in table _AD_Field_ must such that the default value equals **N**.
Right now it is not defined, so it defaults to NULL. 

Additionally, the already existing fields must -via SQL- be set to N (for example in Postgres):
`update adempiere.ad_field set isquickentry='N'`
Otherwise, there is an error when saving.

marcalwestf:
thanks @Raul-mz I will do the whole testing again....

marcalwestf:
Checked out branch _feature/develop_QuickSheet_ and executed all XMLs.

After having finished the XML execution, I proceeded to configure the Quick Sheet for Sales Order Lines.
When trying to save the first field, an error was prompted:
20180726_quicksheet_02



The error on the console when trying to save is
_-----------> GridTable.saveWarning: SaveErrorDataChanged - isQuickEntry= false(java.lang.Boolean) != DB: null -> New: true(java.lang.Boolean)_ 

When executing the SQL
`update adempiere.ad_field set isquickentry='N'`
the error when saving disappears.

marcalwestf:
... actually this seems to be an error from Adempiere.
I don't understand why if there is a null value and you want to set it to true (in the database =N) it is considered an error.

Perhaps there is a deeper meaning behind it?

yamelsenih:
Hi @marcalwestf for now we can change default value for it

marcalwestf:
ok. I'd say then do please change the default value and I'd proceed to accept the pull request.

yamelsenih:
Hi @marcalwestf it is solved yet

marcalwestf:
...testing ....

marcalwestf:
... checked out and XMLs executed.
Now it is possible to configure the Quick Sheet for fields.
I will start testing the functionality.

marcalwestf:
... perhaps as a small improvement: the icon for launching the Quick Sheet in the icon bar (the one at the outmost right) is depicted in the "not available" modus. 

Users do expect a full drawn icon when Quick Sheet is available and the present one when not available.

See screenshot:
20180730_quicksheet_01

marcalwestf:
A helper for testing without much clicking:

How to open Quick Sheet Dialog

- Open dialog with only one new record
- Open dialog with all existing records



Buttons
- F10: Open Quick Entry Dialog
- F2: New record
- ENTER at row end: New record
- ENTER inside a row: edit column
- F3: Delete record
- F4: Save record
- F5: Refresh
- UP, DOWN, LEFT, RIGHT arrows: Function depends on context

- Shitf + ENTER: Navigate to prior column
- ESC: leave edit mode
- Ctrl + Q:	Close Quick Entry Dialog
- Ctrl + ENTER: Close Quick Entry Dialog

(Positive) Side effects: 
- Standard Grid can be navigated with arrows
- Fields of Standard Grid  can be edited with ENTER
- "Normal" Grid and Quick Entry dialog in sync 

**Preferrable to use newest browsers**

marcalwestf:
**Functional test passed!!!**

Merge can be done right now, but if (just for completness sake) the icon issue I commented a couple of minutes before is to be changed, I can wait untill it is solved prior doing the merge of this pull request.

Raul-mz:
@marcalwestf this is ready to try.

marcalwestf:
thanks @Raul-mz ; test follow suit.....

marcalwestf:
I suppose the commit has been done, but the push is missing.
The reason: if I checkout **remotes/origin/feature/develop_QuickSheet** and **feature/develop_QuickSheet**, I see differences, though it should be none.

Here the result after `git checkout feature/develop_QuickSheet`:
20180806_quicksheet_02
This is the result on the console:
```
git checkout feature/develop_QuickSheet
Switched to branch 'feature/develop_QuickSheet'
Your branch is behind 'origin/feature/develop_QuickSheet' by 3 commits, and can be fast-forwarded.
```



Wheras after `git checkout remotes/origin/feature/develop_QuickSheet`:
20180806_quicksheet_01

This is the result on the console:
```
Note: checking out 'remotes/origin/feature/develop_QuickSheet'.
You are in 'detached HEAD' state. You can look around, make experimental
```

marcalwestf:
As I see that the repository was synchronized, I proceeded to testing.
The test was done on a seed database, parting from the latest XML (03910).

marcalwestf:
When executing the file _03920_1088_Add_Column_IsQuickEntry.xml_ as it is on a seed up to XML 0390, there is an error at step 40:
20180807_quicksheet_01

The error is generated because in step 40 the field _isQuickEntry_ should be created for the tab _Field_ in the Application Dictionary, but in the database the column _isQuickEntry_ of table _AD_Field_ -needed for this step- doesn't exist.
And it doesn't exist because in step 30 the column isQuickEntry was created for the AD table _AD_Field_, **but not synchronized with the database**.

Evidently, the migration process synchronizes columns with the database at the end of the migration process.
Unluckily in this case, the synchronization is needed directly after step 30, but it dosen't occur then.
So the column is not synchronized and step 40 tries to create en entry for AD_Field which generates the error.

**The error can be avoided by splitting the XML file into two: the first containing steps 10 to 30; the second containing the other steps.**
Executing the so separated XML file in two chunks, it works.

marcalwestf:
.... thanks @Raul-mz , now the XMLs are applied flawlessly and the translations are executed as well.
Now a short funtional test....

marcalwestf:
all tests passed!

Issue #1744, opened 2018-06-12
Dashboard Role Access


Raul-mz:
Currently all users can see all dashboards configured in adempiere, restrict access to dashboards by role is a solution.

yamelsenih:
Please continue discussion here: https://github.com/adempiere/adempiere/pull/1849

PR #1849, opened 2018-07-31
Feature/#1744 dashboard role access

yamelsenih:
Hi everybody, here a new functionality for dashboard role access, it is from https://github.com/adempiere/adempiere/issues/1744

please take a time for review it

### DIFF
--- a/base/src/org/compiere/model/MRole.java
+++ b/base/src/org/compiere/model/MRole.java
@@ -649,6 +649,8 @@ public final class MRole extends X_AD_Role
 	private HashMap<Integer,Boolean>	m_browseAccess = null;
 	/**	Info Windows			*/
 	private HashMap<Integer, Boolean>	m_infoAccess = null;
+	/**	DashBoard Browse Access				*/
+	private HashMap<Integer,Boolean>	m_dashboardAccess = null;
 	
 	/**
 	 * 	Set Logged in user
@@ -688,6 +690,7 @@ public final class MRole extends X_AD_Role
 			m_workflowAccess = null;
 			m_formAccess = null;
 			m_browseAccess = null;
+			m_dashboardAccess = null;
 		}
 
 		loadIncludedRoles(reload); // Load/Reload included roles - metas-2009_0021_AP1_G94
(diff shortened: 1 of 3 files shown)


marcalwestf:
When applying the file _3940_1744_AddDashboardAccesTab.xml_, an error is generated: 

20180803_dashboard_role_access_01

..and a funny error in the console:
> org.postgresql.util.PSQLException: ERROR: column "ad_dashboard_access_id" is in a primary key
>   Where: SQL statement "alter table ad_dashboard_access alter column ad_dashboard_access_id drop not null"
> PL/pgSQL function altercolumn(name,name,name,character varying,character varying) line 97 at EXECUTE statement
> 	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2412)
> 	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2125)


After some research, I found that two tables were created in the database **ad_dasboard acess** and **ad_dahboard_access** !
20180803_dashboard_role_access_02


Looking then at the XML file, I found, the original definition was **ad_dasboard acess**, which later was changed to **ad_dahboard_access** . This seems not to work when running the XML later.

So I changed all **ad_dasboard acess** to **ad_dahboard_access** and deleted all updates in the XML file:
20180803_dashboard_role_access_03

After doing this, I could load and apply the XML:
20180803_dashboard_role_access_04


This means, it is better not to update a misspelled table name; it is better to set the name correctly from the beginning.

marcalwestf:
Summary, please do replace  all **ad_dasboard_acess** to **ad_dahboard_access** and deleted all updates in the XML file.

marcalwestf:
I have some questions this regarding @yamelsenih : 
Q1: does it only work for ZK? I have only seen changes in class _webui/desktop/DefaultDesktop.java_.

Q2: how has the configuration to be done as to get the expected results?
I ask because I created a single entry in the _Role_ window, _Dashboard Access_ tab, for the _GardenWorld Admin_ role and the Dashboard Content _Calendar_ like in the following image:
20180803_dashboard_role_access_05

I'd expect that for the role _GardenWorld Admin_ only the defined Dashboard Content could be seen.This is not the case as you can see in the following image:
20180803_dashboard_role_access_06

Instead I see -as before- all Dashboard Contents , including the Performance Meters.

Thus my question about configuring; I suppose I've missed a step.

marcalwestf:
functional tests passed.

marcalwestf:
..... and the tab translation is missing:
20180808_role_access_02

marcalwestf:
tested again (applied XML on new seed)

Translation OK. 
**Tab still not attached well to parent tab.**

marcalwestf:
now it's OK

mckayERP:
After applying the latest from develop, there are no desktop access rights applied.  Were these included in the XML?

e-Evolution:
@mckayERP , I had that adding manually for Garden World 

@Raul-mz @yamelsenih  may be is good include other enhancement when you execute System Admin -> General Rules -> Security -> Role Access Update create the full access as other application dictionary components.

yamelsenih:
Hi @mckayERP @marcalwestf @e-Evolution  i fix it here but it is not merged yet https://github.com/adempiere/adempiere/pull/1893

Issue #1354, opened 2017-10-05
Adempiere Smart Browser zoom functionality should be override High Volume marked table.


ayazbaloch:
@e-Evolution and @yamelsenih 
High Volume marked table when zoomed from the Smart Browser should not be prompted for parameters. Zoom functionality already carrying a ID.

In my case a window has high volume data, I made Browser and select the record and click zoom in Smart Browser, it opens a desire window with parameters.

ayazbaloch:
Illustration:-
Login as System----->Find "RV_PP_MRP_Supply" into Table and Column window.
Mark it as "High Volume"
![image](https://user-images.githubusercontent.com/10656336/31574873-50c3b108-b0f2-11e7-947b-3c89349b8946.png)

Now login as GardenAdmin
Run Cache Reset process.
Open "MRP Supplies Browse" browser.  Extract all records.

![image](https://user-images.githubusercontent.com/10656336/31574896-356f37aa-b0f3-11e7-8197-e88a3be37c1c.png)

Select any one record then click zoom button. It will as you for parameter value, leave it blank and click OK button.

![image](https://user-images.githubusercontent.com/10656336/31574907-814aff74-b0f3-11e7-92c6-f9761a93f1d9.png)

Now "MRP Supplies View" window showed all records.

![image](https://user-images.githubusercontent.com/10656336/31574918-d0ef057a-b0f3-11e7-8bed-2dcb5d40925b.png)

yamelsenih:
Hi @ayazbaloch le me view it.

yamelsenih:
Hi @ayazbaloch i will fix it

yamelsenih:
Please continue discussion and testing here: https://github.com/adempiere/adempiere/pull/1855

ayazbaloch:
ok, @yamelsenih  let me check and test.

ayazbaloch:
Hi @yamelsenih  I checked it ok for select one record it shows properly without prompting parameters.
But Selecting more than one records it shows 1st record, where you already mentioned in the following lilnk.
https://github.com/adempiere/adempiere/issues/1458

PR #1855, opened 2018-08-02
#1354, fixed error with bad zooming from Smart Browser, reference to

yamelsenih:
issue: https://github.com/adempiere/adempiere/issues/1354

Hi giy, please test it, in Swing UI is worked but ZK is not worked.

Now is ok.

Best regard


marcalwestf:
Tested as described in issue https://github.com/adempiere/adempiere/issues/1354 : 
Defined table as _High Volume_:
20180808_bad_zoom_01

Then opened Smart Browser _Process Orders_  and selected + zoomed several records. It retrieves exactly the records selected:
20180808_bad_zoom_02

20180808_bad_zoom_03

...selecting a single record yields also the expected result:
20180808_bad_zoom_04
20180808_bad_zoom_05


BTW, it is irrelevant to select _High Volume_: the results are the same.

PR #1862, opened 2018-08-07
#1861 Add support to direct load flag in packin import, reference to

### LINKED ISSUE #1861: Dicrect Load for Packin is ussupported
When a Packing is imported from Package module on Application Dictionary, it can be used for create XML  migration directly from import.

The bug is the follow:

For PO class exists a flag named isDirectLoad, it allows know when a object is created from import or not, example:

- Without direct load flag:
  - Create Table: It run a method after save for create default columns
- With direct load enabled:
  - Create Table: It not run method for create default columns because all columns will be created for import.

Currently it functionality is unsupported for packin

yamelsenih:
issue:https://github.com/adempiere/adempiere/issues/1861

Hi guys it is a bugfix of ADempiere packing.

fixes #1861


yamelsenih:
Hi @marcalwestf here https://github.com/adempiere/adempiere/issues/1861#issue-348362374

marcalwestf:
Results of test
1.- checked out bugfix/#1861-Direct-Load-For-Packin
2.- buit + installed , no XML executed
3.- run Adempiere on a seed up to XML 03910
4.- created a new table, synched it with DB
5.- created a packout and exported the packout
6.- restored the DB to the seed up to XML 03910
7.- created a packin, referring to the package created before; imported the packin
An error is generated:
20180810_packin_10

The corresponding log:
```
===========> PackIn.importXML: importXML: [72]
java.lang.IllegalArgumentException: AD_Original_ID is mandatory.
        at org.compiere.model.PO.set_Value(PO.java:791)
        at org.compiere.model.PO.set_Value(PO.java:740)
        at org.compiere.model.X_AD_Package_Imp_Detail.setAD_Original_ID(X_AD_Package_Imp_Detail.java:99)
        at org.adempiere.pipo.AbstractElementHandler.record_log(AbstractElementHandler.java:107)
        at org.adempiere.pipo.handler.TableElementHandler.startElement(TableElementHandler.java:164)
        at org.adempiere.pipo.PackInHandler.startElement(PackInHandler.java:437)
```

This is caused because a null value is tried to be set to a column (AD_Original_ID) of table ad_package_imp_detail, which is defined as NOT NULL:
20180810_packin_12


Same happens when exporting a newly created window, which refers to a existing table.

yamelsenih:
Hi @marcalwestf please test it, i fixed problem.

Best regard

marcalwestf:
Hi Yamel, thanks; I'm about to start testing.
One question; on your last commit (https://github.com/adempiere/adempiere/pull/1862/commits/2053ca1c815240ec4d4d8eed77242b51a102f37b) you changed sevelal files, mostly on _pipo/handler_.
What are the effects? What should I take care of?

Thanks

Issue #1863, opened 2018-08-07
The parameter for the field of Smart Browser does not take the default value defined when this field is marked as a key.


e-Evolution:
Currently, any field that is marked as a key does not allow to establish a default value in a field for a parameter, window field or smart browser field.

This can be correct in a window since the keys field is usually generated by the internal id of a sequence, but for a parameter of a process or field of a smart browser it is necessary to allow a value to be set by default so that it works in a filter of the call of an Smart Browser

yamelsenih:
Merged

PR #1865, opened 2018-08-07
#1863 The parameter for the field of Smart Browser does not take the …

e-Evolution:
…default value defined when this field is marked as a key https://github.com/adempiere/adempiere/issues/1863

(cherry picked from commit 0de17bf93da2a8040ef948f615cc90b4d5abdb37)


yamelsenih:
I tested it and is ok

Issue #1649, opened 2018-04-09
OrderPOCreate - Generate PO from Sales Order creates POs from Drafted sales order


homebeaver:
this is a bug. The process has to generate POs only if the DocStatus of the sales order is `Completed` (Status `InProgress` seems for me ok too), but not on `Drafted` as the example shows:

![grafik](https://user-images.githubusercontent.com/3493126/38510644-f14792f2-3c25-11e8-9cc6-01d1b5ae8b89.png)

homebeaver:
the solution for this Bug is simple. Add the following in private method `OrderPOCreate.createPOFromSO`
 
```java
...
if(DocAction.STATUS_InProgress.equals(so.getDocStatus()) || DocAction.STATUS_Completed.equals(so.getDocStatus())) {
	// OK - continue 
} else {
	log.warning("DocStatus is " + so.getDocStatus());
	return 0;
}
...
```
But there are some other shortcomings:

- products which are stocked and onHand should not be purchased
- products with `isPurchased='N'` should not be purchased
- BOM products `isBOM='Y'` should not be purchased

yamelsenih:
Hi @homebeaver have you a pull request for it? i can test for you

Best regard

homebeaver:
Hi @yamelsenih - I'm working on it.

yamelsenih:
Hi @homebeaver y have a change for it, i seem mani commits from your change.

Want you get my change and adde your changes?

Best regard

homebeaver:
thank you very much @yamelsenih  
I'II have no time this year to track the pull request #1866 . Sorry :-(
Mybe you can bring this PR to an end in 3.9.1

regards EUGen

yamelsenih:
Yes @homebeaver i will to review your PR

yamelsenih:
Hi @homebeaver I added a validation for document status here https://github.com/adempiere/adempiere/pull/2201

Best regard

PR #1651, opened 2018-04-10
patch OrderPOCreate process, see issue #1649

homebeaver:
Hi @yamelsenih, this is the pull request as promised today.
You can test it.

A remark on BOM products. I explode the BOM but I did not implement the purchasing of components of the BOM-product. 

regards

PR #1866, opened 2018-08-08
#1649 OrderPOCreate process patch for bug 

homebeaver:
Hi @yamelsenih , @e-Evolution ,

this is a replacement for PR https://github.com/adempiere/adempiere/pull/1651 

It fixes bug https://github.com/adempiere/adempiere/issues/1649


base/src/org/compiere/process/OrderPOCreate.java
```diff
@@ -38,54 +44,50 @@
...
-	/**	Vendor				*/
-	private int			p_Vendor_ID;
-	/**	Sales Order			*/
-	private int			p_C_Order_ID;
-	/** Drop Ship			*/
-	private boolean		p_IsDropShip = false;
 	
+	private int m_explosion_level = 0;
+	private Set<MPPProductBOMLine> bomSet = new HashSet<MPPProductBOMLine>(); 
+	private int explosion(MPPProductBOM bom) throws AdempiereUserError {
```
yamelsenih: Hi @homebeaver i don't know what is the reason for add BOM process here, BOM process is created when you complete a Sales Order, see it here: https://github.com/adempiere/adempiere/blob/develop/base/src/org/compiere/model/MOrder.java#L1239

All changes is ok but in don't agree with add a separed process for make **exprosion** is it is part of business model for MOrder

Best regard

homebeaver: @yamelsenih , let me explain:

if you have a BOM-product in your sales order and you create the purchase order from it, so you cannot purchase the BOM-product itsself, You have to purchase all the components and produce the product (make-to-order). This was not considered in `OrderPOCreate` .
I did not implement purchase components, nor make-to-order. This is heavy work. 
But at least I send a warning:
```
   // TODO purchase component and prepare BOM production
  log.warning(mProduct.toString() + " is BOM product, will not be purchased (not implemented)");
```
It's up to you (or us) to decide what is better
- send a warning on not implemeted feature
- do nothing , remove the explosion and do not inform the user about the hole

This have some interfaces to production.
What do you think? What is the opinion of Mario @marcalwestf  ?
regards

marcalwestf:
giving it a look....

marcalwestf:
hi Eugen,
I agree with you on the status issue.

Here my comments about BOM explosion + PO.

You wrote
> if you have a BOM-product in your sales order and you create the purchase order from it, so you cannot purchase the BOM-product itsself

My comments: we have here to remember that it is also possible that you can buy a product with a BOM. This would be the case when you mark the product as _isPurchased_ and it has also a BOM. This is not uncommon.

Then you stated
> You have to purchase all the components and produce the product (make-to-order). 

My comments: the correct procedure would be to create first a _PPOrder_ (Libero) or a _M_ProductionBatch_ (Manufacturing Light). This two documents generate demand, which would be satisfied with Purchase Orders or Productions. 

Also, different demands are usually **bundled** into a single Purchase Order for example using MRP or Replenishment reports. It is uncommon to directly satisfy a demand for a single Sales Order. This would disrupt the practice implemented.

Thus IMHO I would agree **not to include the BOM explosion and Purchase Order for the exploded products**.

I'd suggest to leave only the changes concerning the status and not include the explosion.

base/src/org/compiere/process/OrderPOCreate.java
marcalwestf: see my answer below

homebeaver:
OK. Thanks Mario, @marcalwestf for the explanation

marcalwestf:
Eugen, would you do the suggested changes? I'd accept the pull request then.
(it's a pity because of the nice and complex code)

homebeaver:
Mario @marcalwestf I am on holidays for the next 3 weeks.
Can you or @yamelsenih please delete or comment out the unnecessary lines in `OrderPOCreate.java` tkx+regards

homebeaver:
Hi Mario, you 've looked to a wrong branch. The correct is
https://github.com/klst-de/adempiere/commits/develop+1649 (with a plus) Try tihis 👍 
regards E.

marcalwestf:
hi Eugen @homebeaver , I looked again at the pull request.

I compared the changes and they look OK.
I like this PR because you have also improved the code and updated the process.

Only with the following snippet I have some doubts:
```
if(purchaseProduct && mProduct.isStocked()) { // check whether the product is onHand
  MStorage[] mStorages = MStorage.getOfProduct(getCtx(), M_Product_ID, get_TrxName());
  for(int s = 0; s < mStorages.length; s++) {
	log.config("mStorage="+mStorages[s]);
	if(mStorages[s].getQtyOnHand().compareTo(soLines[i].getQtyOrdered()) > -1 ) {
		log.warning(mProduct.toString() + " is onHand, will not be purchased");
		purchaseProduct = false;
	}
  }
}
```
1.- This would conflict with backward compatibility: up to now, the process worked **without** warehouse check. What happens with existing implementations that rely on the fact that the Purchase Order is generated "as a whole" out of the Sales Order?  
With the the pull request as it it, this would not be possible because of the _quantity on hand_ check.
This could be overcome with an additional parameter in the process.

2.- When checking storage (for example using the parameter suggested above), then you should first add _quantity on hand_ for all warehouses and only then compare, instead of just comparing just one warehouse:
```
//	Qty On Hand
BigDecimal qtyOnHand = Env.ZERO;
MStorage[] storages = MStorage.getWarehouse (getCtx(), 0,
    mProduct.getM_Product_ID(), soLines[i].getM_AttributeSetInstance_ID(),
	null, MClient.MMPOLICY_FiFo.equals(MMPolicy), true, soLines[i].getM_Locator_ID(), get_TrxName());

if(storages != null
		&& storages.length > 0) {
	for(MStorage storage : storages) {
		qtyOnHand = qtyOnHand.add(storage.getQtyOnHand());
	}
}

if(qtyOnHand.compareTo(soLines[i].getQtyOrdered()) > -1 ) {
	log.warning(mProduct.toString() + " is onHand, will not be purchased");
	purchaseProduct = false;
	}
```

If you think this is too much a trouble, you can comment out the snippets lines and we can discuss this matter of checking _on hand quantity_ later.

Sorry to bother you again, but I think backward compatibility is an important matter.
Thanks for your understanding.

marcalwestf:
hi Eugen @homebeaver as we're accepting the last pull requests prior to the 3.9.1 release, I wonder if you can do something regarding this pull request?
Thanks

homebeaver:
> 
> 
@marcalwestf wrote
> ... I wonder if you can do something regarding this pull request?
Hi Mario,
let us switch it after 3.9.1 - no hasty reaction at finish line.
PS: and thanks for the comments
bye EUGen

yamelsenih:
Replaced by https://github.com/adempiere/adempiere/pull/2201

Thanks @homebeaver

Issue #1873, opened 2018-08-11
Add helper method for TimeUtil class


yamelsenih:
Currently not exist a helper method for get duration between two dates, i want to add follow methods to TimeUtil class:

- Get Duration from two dates and validate if a date is null
<pre>public static long getMillisecondsBetween(Timestamp dateFrom, Timestamp dateTo)</pre>
- Get Time Between two dates expressed in durationUnit
<pre>public static double getTimeBetween(Timestamp dateFrom, Timestamp dateTo, String durationUnit)</pre>
- Get Hours between two dates
<pre>public static int getHoursBetween(Timestamp dateFrom, Timestamp dateTo)</pre>
- Get Minutes between two dates
<pre>public static int getMinutesBetween(Timestamp dateFrom, Timestamp dateTo)</pre>
- Get Seconds between two dates
<pre>public static int getSecondsBetween(Timestamp dateFrom, Timestamp dateTo)</pre>

yamelsenih:
Please continue discussione here: https://github.com/adempiere/adempiere/pull/1874

PR #1874, opened 2018-08-12
Feature/#1873 add time util helper method

yamelsenih:
https://github.com/adempiere/adempiere/issues/1873

Hi everibody, it is a little enhencement for TimeUtil class, just add helper methods.

Best regard


marcalwestf:
..testing would be just to see if it compiles

Issue #1852, opened 2018-07-31
cannot RUN_MigrateXML.sh : 390lts-391 - A circular Problem.


homebeaver:
on 13 Dec 2017  there was a [commit](https://github.com/adempiere/adempiere/commit/a510d03185c8c172155525d820bed648ba4d0b15) which changes the mandatory parameter name
from oldValue="FileName" to FilePathOrName in the migration process

Now I have a bootstrap problem ( headless setup ). I get
```
*** 2018-07-31 16:11:19.481 Adempiere Log (CLogFile) ***
16:11:19.481 Adempiere.startup: ADempiere(r) Release 3.9.0_2017-08-01 -Smart Suite ERP,CRM and SCM- (c) 1999-2017 ADempiere(r); Implementation: ${ADEMPIERE_VERSION} 20180731-0956 - ${ADEMPIERE_VENDOR} [1]
16:11:19.482 Ini.loadProperties: /home/eugen/adempiere-develop-1567/adempiere/Adempiere/Adempiere.properties #29 [1]
16:11:20.034   CLogMgt.setLevel: CONFIG [1]
16:11:20.079   GridField.loadLookup: (ReportType) [1]
16:11:20.254   MWorkflow.getDocValue: #1 [1]
16:11:20.386 Msg.initMsg: Records=1462 - en_US [1]
16:11:20.396   MIssue.create: Fill mandatory fields: File Name [1]
16:11:20.404   GridField.loadLookup: (SystemStatus) [1]
===========> MigrationFromXML.process: Fill mandatory fields: File Name [1]
org.adempiere.exceptions.AdempiereException: Fill mandatory fields: File Name
        at org.compiere.process.SvrProcess.validateParameter(SvrProcess.java:241)
        at org.compiere.process.SvrProcess.process(SvrProcess.java:172)
        at org.compiere.process.SvrProcess.startProcess(SvrProcess.java:128)
        at org.adempiere.util.ProcessUtil.startJavaProcess(ProcessUtil.java:160)
        at org.adempiere.util.ProcessUtil.startJavaProcess(ProcessUtil.java:105)
        at org.compiere.process.ServerProcessCtl.startProcess(ServerProcessCtl.java:415)
        at org.compiere.process.ServerProcessCtl.run(ServerProcessCtl.java:246)
        at org.eevolution.service.dsl.ProcessBuilder.run(ProcessBuilder.java:216)
[...]

PR #1886, opened 2018-08-13
#1852 cannot RUN_MigrateXML.sh : 390lts-391 - A circular Problem. htt…

e-Evolution:
…ps://github.com/adempiere/adempiere/issues/1852

- The problem is that the FileName not exist how core element, so when the synchronization is executed the system change column name by new element name , this change enable back compatibility

(cherry picked from commit c70572c991e4fa436e4d1aca1845899428850f27)

### DIFF
--- a/base/src/org/compiere/process/MigrationFromXMLAbstract.java
+++ b/base/src/org/compiere/process/MigrationFromXMLAbstract.java
@@ -46,6 +46,8 @@ public abstract class MigrationFromXMLAbstract extends SvrProcess {
 	@Override
 	protected void prepare() {
 		filePathOrName = getParameterAsString(FILEPATHORNAME);
+		if (filePathOrName == null)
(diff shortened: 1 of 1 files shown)


base/src/org/compiere/process/MigrationFromXMLAbstract.java
```diff
@@ -46,6 +46,8 @@
 	@Override
 	protected void prepare() {
 		filePathOrName = getParameterAsString(FILEPATHORNAME);
+		if (filePathOrName == null)
```
yamelsenih: Hi @e-Evolution it is good for work but bad for future changes, this class is volatile and any developer can regenerate from Application Dictionary

yamelsenih:
fixes #1852

e-Evolution:
fixes

PR #1920, opened 2018-08-21
Bugfix/invoice reverse accrual costdetail match inv

SusanneCalderon:
**Description of the pull request:**
With the implementation of the new method _reverseAccrual_ the handling of the documents match inv and match PO was changed.
Before this implementation the _match_inv_ and _match_PO_ were deleted when you cancelled a purchase invoice or a material receipt. The problem of this procedure was that it was necessary to delete from the accounting. Above all using costing method _average Invoice_ it was necessary to recalculate manually the product costs.
Implementing _reverseAccrual_ this procedure was changed, the documents match_Inv and match_PO were cancelled creating a reversal document which created the corresponding m_Costdetail registers.

**Errors using costing method average Invoice/average PO**

- the method createReversalCostDetail() in class AbstractCostingMethod does not find the original costdetail of a cancelled match_Inv/match_PO. So the product cost are never updated.

The following screenshot shows a transaction value report with a cancelled purchase invoice, the costs are not updated to the value of the purchase order.
kardexmatchinv

- The new costdetail, generated with the values of the original document does not negate the values of costamt, costamtLL, costadjustment, costadjustmentLL because the movement quantity of a costdetail of matchInv/matchPO is zero.
- Using revertaccrual with the currentdate as dateacct does not necessarily find the material receipt as last costdetail, but costdetails of formally cancelled match inv or costdetails of landed cost allocations.

**Solution:**

- Generate the fields M_MatchInv_ID and M_MatchPO_ID in table M_Costdetail and search the original entries with these references.
- Negate the values in the new cost detail.
- Use the costdetail of the material receipt for a correct calculation of the new purchase costs.
- Insert the new costdetail at the end in case of reverseaccrual.

With these changes the product costs are updated correctly, using reverseCorrect and reverseAccrual. 
The base of the class AbstractCostingMethod was the modified class of pull request https://github.com/adempiere/adempiere/pull/1917, which corrects the wrong sequence number in cost detail for reverseAccual.

As attachment I send the results of the tests of reverseCorrectIt and reverseAccrual for costing method average Invoice.
[TestAverageInvoice.pdf](https://github.com/adempiere/adempiere/files/2308135/TestAverageInvoice.pdf)


e-Evolution:
@marcalwestf

Review the change code can see some class was modify completly , please check your formaring IDE or character end line 

The change only should the lines that was modified 


https://github.com/adempiere/adempiere/pull/1920/files#diff-1

SusanneCalderon:
Hi @e-Evolution , I changed the line seperator

e-Evolution:
fixes

Issue #1925, opened 2018-08-24
Migration 3605 Step 890 fails but shows applied.


mckayERP:
Develop branch.  Ref #1532.  Migration 3605 step 890 fails (trace below) but the step status shows "Applied".  Clicking the record ID button open the Table/Column window with a null record.

The record adds a translation for "RequestProcessor" with AD_Column_ID=86204.  Likely the column was deleted but the translation wasn't.

A couple of questions I haven't taken the time to look at
- I though translations weren't included/captured in migration scripts?
- If a migration steps fails, shouldn't the step fail and the migration process stop?

Quick solution - delete the step.

08:00:41.142   MMigrationStep.apply: Applying migration step: Migration 3605 - Project Processor Window #1532 - 3.9.1 (D) ; Step: 890; Type: AD.  [1]
08:00:41.636 MTable.getPO: Using GenericPO for AD_Column_Trl [1]
-----------> GenericPO.set_ValueNoCheck: AD_Language - Class invalid: class java.lang.Integer, Should be class java.lang.String: 0 [1]
08:00:41.762   MIssue.create: POSave_4bc3afb4-706b-463a-8f8f-acd76d14b61a [1]
08:00:41.919   GridField.loadLookup: (SystemStatus) [1]
===========> Trx.commit: POSave_4bc3afb4-706b-463a-8f8f-acd76d14b61a [1]
org.postgresql.util.PSQLException: ERROR: insert or update on table "ad_column_trl" violates foreign key constraint "adcolumn_adcolumntrl"
  Detail: Key (ad_column_id)=(86204) is not present in table "ad_column".; State=23503; ErrorCode=0
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2412)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2125)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:297)
	at org.postgresql.jdbc.PgConnection.executeTransactionCommand(PgConnection.java:739)
	at org.postgresql.jdbc.PgConnection.commit(PgConnection.java:760)
	at com.mchange.v2.c3p0.impl.NewProxyConnection.commit(NewProxyConnection.java:803)
	at org.compiere.util.Trx.commit(Trx.java:336)
	at org.compiere.util.Trx.commit(Trx.java:364)
[...]

PR #1927, opened 2018-08-24
#1925  Deleted migration step 890 from 03605_1532...xml

mckayERP:
…cessorWindow.xml

### DIFF
--- a/migration/390lts-391/03605_1532_ProjectProcessorWindow.xml
+++ b/migration/390lts-391/03605_1532_ProjectProcessorWindow.xml
@@ -2721,22 +2721,6 @@ There are two reasons for de-activating and not deleting records:
         <Data AD_Column_ID="84310" Column="UUID" isNewNull="true"/>
       </PO>
     </Step>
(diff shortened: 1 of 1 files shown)


marcalwestf:
Seems to refer to issue https://github.com/adempiere/adempiere/issues/1564

One question @mckayERP : what is the reason for deleting it?

mckayERP:
Deleting the step will prevent the error from appearing in the logs.  Application of the XML suggests a review of the logs following the application to determine if there were any errors. See step 10 in http://wiki.adempiere.net/Creating_a_New_Seed_Database.

marcalwestf:
Hi,
as I see, the error in step 350 is produced because there is reference to `AD_Column_ID=86204` which doesn't exist and also is not created in file _03605_1532_ProjectProcessorWindow.xml_.
I too agree that when deleting the step, the error doesn't appear.

Perhaps should better @e-Evolution  look at it and create the missing column if it is missing.

e-Evolution:
fixes

Issue #1928, opened 2018-08-24
Eclipse project Web Deployment Descriptor missing Jasper Report libraries.


mckayERP:
Develop branch at d016e58f... Eclipse project web deployment assembly is missing jasper report libraries.  These were updated in the project properties class path but not the web deployment assembly.  Without the libraries in the web deployment assembly, running the adempiere project in a server in eclipse fails.  

[INFO] Initializing c3p0 pool... com.mchange.v2.c3p0.ComboPooledDataSource [ acquireIncrement -> 3, acquireRetryAttempts -> 2, acquireRetryDelay -> 1000, autoCommitOnClose -> false, automaticTestTable -> null, breakAfterAcquireFailure -> false, checkoutTimeout -> 0, connectionCustomizerClassName -> null, connectionTesterClassName -> com.mchange.v2.c3p0.impl.DefaultConnectionTester, dataSourceName -> AdempiereDS, debugUnreturnedConnectionStackTraces -> false, description -> null, driverClass -> org.postgresql.Driver, factoryClassLocation -> null, forceIgnoreUnresolvedTransactions -> false, identityToken -> 1b60o0p9xug6gd81t9k82h|64030b91, idleConnectionTestPeriod -> 1200, initialPoolSize -> 10, jdbcUrl -> jdbc:postgresql://localhost:5432/adempiere?encoding=UNICODE, maxAdministrativeTaskTime -> 0, maxConnectionAge -> 0, maxIdleTime -> 1200, maxIdleTimeExcessConnections -> 1200, maxPoolSize -> 150, maxStatements -> 0, maxStatementsPerConnection -> 0, minPoolSize -> 5, numHelperThreads -> 3, numThreadsAwaitingCheckoutDefaultUser -> 0, preferredTestQuery -> SELECT Version FROM AD_System, properties -> {user=******, password=******}, propertyCycle -> 0, testConnectionOnCheckin -> false, testConnectionOnCheckout -> false, unreturnedConnectionTimeout -> 0, usesTraditionalReflectiveProxies -> false ]
===========> [/webui].log: StandardWrapper.Throwable [1]
java.lang.NoClassDefFoundError: net/sf/jasperreports/engine/JRException
	at org.adempiere.webui.session.WebUIServlet.init(WebUIServlet.java:83)
	at org.apache.catalina.core.StandardWrapper.loadServlet(StandardWrapper.java:1213)...

PR #1929, opened 2018-08-24
#1928 Add missing jasper report libraries to the eclipse web deployme…

mckayERP:
…nt descriptor.  Only relevant to eclipse users but helps when running adempiere on a server in eclipse.


marcalwestf:
hi @mckayERP , could you explain what the error is and what it solves ?

I see on Eclipse (Proton) that the change doesn't affect the compilation nor the building of ADempiere, but as I don't know the scope of file _org.eclipse.wst.common.component_ nor the syntax of the changes, I cannot judge if there is any side-effect.
Thanks

mckayERP:
@marcalwestf 
Hi Mario,

Its the web deployment assembly which defines the packaging structure for the web application project.  Its missing the jasper libraries and this only causes an issue in eclipse when launching the webui from a server.

To test, use a clean checkout of the develop branch and try to launch the server in Eclipse.

The image below shows the jasper libraries are missing.

image

PR #1941, opened 2018-08-27
JDK10

marekmosiewicz:
This is pull request to make compile with JDK10. There are three functionalities missing:
 LDAP authentication
 Theming Metal PLAF
 Old html gui which is probably no more used.
I switched build.xml to JDK10 target so it will not compile with previous JDK. Probably build.xml should be without target so it can be compiled with previous versions of JDK
There is also work to do to have setup run with JDK10


e-Evolution:
@marekmosiewicz Thanks a lot for your contribution, I reviewed but found two big problems, first you remove LDAP support that is important for a customer that uses LDAP Server, other remove the User Interface Theme it is not good because some customer want to have your settings

So need solve this to can accept into the ADempiere core

marcalwestf:
hi Marek; there are some conflicts.
Could you please solve them so we can look at the PR?
Thanks

e-Evolution:
implementing with #3479  Thank you a lot @marekmosiewicz for your contribution.

Issue #1936, opened 2018-08-26
System Admin role - Migration window shows lookup with no identifying records.


mckayERP:
When first opened, the Migration window shows a blank lookup window. 

![image](https://user-images.githubusercontent.com/5507616/44633288-3a4a5900-a957-11e8-9759-c7f616b99763.png)

PR #1942, opened 2018-08-27
#1936 Adding selection records to AD_Migration

mckayERP:
Added selection records version, sequence number and name.


marcalwestf:
...after loading and applying _04010_1936_MigrationTableLacksSelectionRecords.xml_, the _Migration_ window displays the filter dialog when selecting the _Migration_ window from the menu:
screenshot_20180829_110346_migration_table

marcalwestf:
for completeness' sake: it refers to issue #1936

PR #1947, opened 2018-08-28
#1944 Deal with special case where a Y/N field has been added to a populated table.

### LINKED ISSUE #1944: First save of a record with new Yes-No field causes error.
Error is displayed as 

System will re-query record. IsDefault= false(java.lang.Boolean) != DB: null -> New: true(java.lang.Boolean)

The problem is that the PO reads the database value as null while the gui reads the original value as false.

Fix is to add 

`|| (oldValue instanceof Boolean && !((Boolean) oldValue) && dbValue == null)`

to the insertion test in GridTable.java

mckayERP:
Fixes #1944
When a record is being edited with a value in that field there is a problem.  When the comparison of the old value in the model where null = 'N' and the database where null = null don't match, the PO save fails.  Have to account for the fact that null='N' in the database as well and allow the mismatch.

### DIFF
--- a/base/src/org/compiere/model/GridTable.java
+++ b/base/src/org/compiere/model/GridTable.java
@@ -2134,14 +2134,20 @@ public class GridTable extends AbstractTableModel
 				Object dbValue = po.get_Value(poIndex);
 				if (m_inserting 
 					|| !m_compareDB
+					
 					//	Original == DB
 					|| (oldValue == null && dbValue == null)
 					|| (oldValue != null && oldValue.equals (dbValue))
 					//  #1944  Special case when the field is boolean and not initialized
+					//  false equivalent to null allowed
 					|| (oldValue instanceof Boolean && !((Boolean) oldValue) && dbValue == null)
+					
 					//	Target == DB (changed by trigger to new value already)
 					|| (value == null && dbValue == null)
 					|| (value != null && value.equals (dbValue)) 
+					//  #1944  Special case when the field is boolean and not initialized
+					//  false equivalent to null allowed
+					|| (value instanceof Boolean && !((Boolean) value) && dbValue == null)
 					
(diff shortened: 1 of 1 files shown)


marcalwestf:
fixes #1944

marcalwestf:
... somehow I came across such error in the past. 
If I remember well, this can happen also after creating a column Y/N and not giving a default value of Y or N.

Mike could you please provide with an easy example of reproducing this bug?
So a reviewer doesn't have to try out different scenarios until hitting the correct one.
Thanks.

marcalwestf:
@marekmosiewicz yes you're right in you explanation.

My case here is how to reproduce the error so I can test and accept the bug fix.

marekmosiewicz:
@marcalwestf  I was able to reproduce when set yes/no to yes and saving on record which was already exisitng before adding column. Th error is :
"Could not save changes - data was changed after query. 

System will re-query record. IsDefault= false(java.lang.Boolean) != DB: null -> New: true(java.lang.Boolean)"
Seems that problem is in PO.java

marcalwestf:
thanks @marekmosiewicz 
When I do as you explain, I get the error you mention when trying to save:
screenshot_20180830_105959_grid_yn_01

When instead I run the code of the pull request, the error disappears:
screenshot_20180830_105959_grid_yn_02


the code for discriminating this case is 
`|| (oldValue instanceof Boolean && !((Boolean) oldValue) && dbValue == null)`

I wonder when the other added  code of this pull request applies:
`|| (value instanceof Boolean && !((Boolean) value) && dbValue == null)`

perhaps can Mike explain.
From my side, this would be it.

mckayERP:
I only added the checks on "value" as it seemed prudent to make the two cases parallel.  The need to check value rather than old value is not obvious to me. I'm not sure when the "value" case would be relevant.  I'm relying on the comment which mentions that the value may be set by a trigger but I couldn't find an example.

Issue #1938, opened 2018-08-26
Error setting SeqNo when creating new migration step records.


mckayERP:
The following error appears in the logs when importing migration scripts or recording new migrations.

GridField.getDefault: (SeqNo) - Default SQL variable parse failed: @SQL=SELECT COALESCE(MAX(SeqNo),0)+10 AS DefaultValue FROM AD_MigrationStep WHERE AD_MigrationStep_ID=@AD_MigrationStep_ID@

Should be (I think)

@SQL=SELECT COALESCE(MAX(SeqNo),0)+10 AS DefaultValue FROM AD_MigrationStep WHERE AD_Migration_ID=@AD_Migration_ID@

PR #1949, opened 2018-08-28
#1938 Changed the Default Logic in AD_MigrationStep

mckayERP:
Changed the Default Logic in AD_MigrationStep to reference AD_Migration_ID = @AD_Migration_ID@ instead of AD_MigrationStep_ID = @AD_MigrationStep_ID@


mckayERP:
Used the wrong source branch.

Issue #1638, opened 2018-04-04
Bug Import Business Partner when exist a contact with same email


e-Evolution:
Currently, when try import the same contact with the same email for multiples Business Partner, Adempiere launch error, the reason is that Business Partner Importer not validate if an email from a specific business partner.

PR #1970, opened 2018-09-02
#1638 Import Business Partner when a contact with same email exists

e-Evolution:


SusanneCalderon:
Class ImportBpartner, Line 269
**Error in SQL**, the alias of table c_Bpartner is bp, but it is referenced with pb. After correcting this error I imported a business partner with the same email. 
An existing business partner had the same email.
errorimportbpartner

e-Evolution:
pull request close and replace by https://github.com/adempiere/adempiere/pull/2043

PR #1974, opened 2018-09-02
#1451 Add additional condition to an Standard Request https://github.…

e-Evolution:
…com/adempiere/adempiere/issues/1451

- Fixed sequence xml script


marcalwestf:
hi @e-Evolution is it rather better to change the file's name and keep the sequence instead?

For one thing, there is already a sequence number **03430** in the XML files.
Changing only the XML file's name to _03420_1451_AddAdditionalConditionToStandardRequestType.xml_ and keeping the sequence number as it is on the other side would additionally fill the sequence gap, because sequence **3420** is missing..

This is a really minor matter, but as we are dealing with it...

e-Evolution:
@marcalwestf I fixes the XML Script Sequence

Issue #1555, opened 2018-01-28
BankTransfer: Voiding Payments generated by the process BankTransfer


SusanneCalderon:
When you want to transfer money from one bank account to another you can do it by using the process BankTransfer,  which generates two Payments. 
![20180128_banktransfer_01](https://user-images.githubusercontent.com/15349639/35483382-dc2719e4-0406-11e8-9002-e0e9faa2f748.PNG)

The result is as follows, two payments which is correct.
![20180128_banktransfer_02](https://user-images.githubusercontent.com/15349639/35483367-a1078268-0406-11e8-9022-dff72359936b.PNG)

Usually when you void one payment, you want to void the other, too. Now it has to be done manually and it is easy to be forgotten. 
![20180128_banktransfer_04](https://user-images.githubusercontent.com/15349639/35483421-52d08f62-0407-11e8-8bbb-b69883687e65.PNG)

The enhancement is as follows:

- Similar to the column _reversal_ID_ the corresponding payment is identified by a new column _c_Paymentrelated_ID_ which relates both payments. 
- An additional flag _isVoidRelatedPayment_ in the window Payment decides if the related payment is also voided after voiding a Payment. The default value for this flag is no for backward compatibility. 

This enhancement helps to improve the efficiency of the department responsible for payments.

SusanneCalderon:
After voiding the first payment the related is voided, too.
![20180128_banktransfer_05](https://user-images.githubusercontent.com/15349639/35487874-8afde6fc-0446-11e8-8bb2-6a5a9e69fee7.PNG)

e-Evolution:
@SusanneCalderon 

When you defined a new columns where the table not exist , you not should use C_PaymentRelated_ID because the ADempiere standard logic will try a table called C_PaymentRelated.

So the good practice is if a column is external link the same table in this case C_Payment then do you should use PaymentRelated_ID because this will be defined as Table and not as TableDir so you not need the prefix C_

marcalwestf:
Now the name is defined as the convention.

yamelsenih:
I will test it

SusanneCalderon:
Rename paymentrelated_ID to relatedPayment_ID, similar a RelatedProduct_ID,
delete isVoidRelatedPayment.
Reason: when the payment is created by a banktransfer, the related payment should always be cancelled .

PR #1984, opened 2018-09-10
#1555 BankTransfer: Voiding PaymentsBankTransfer

SusanneCalderon:
fixes https://github.com/adempiere/adempiere/issues/1555

- Changes
Rename _Paymentrelated_ID_ to _RelatedPayment_ID_, Suggestion from @e-Evolution , in order to be conform with then name rule established in Adempiere(Example: RelatedProduct_ID in table M_RelatedProduct).

- Delete the flag _IsVoidRelatedDocument_ in _C_Payment._
Reason:
The process _BankTransfer_ and the new feature _Create Multiple Bank Transfers_ generate at the end 2 payments, a payment and a receipt for the payee. 
In order to relate these payments the IDs are stored in the field _RelatedPayment_ID_. 
When you cancel the payment/receipt the related document should always be canceled, too. Not depending of the flag IsVoidRelatedDocument.


e-Evolution:
@SusanneCalderon I review this improve and found some issues.

The code still use getC_RelatedPayment() and should use getRelatedPayment()

other problem is that if the original payment is a reverse correction the related payment si always void , is no ok , so should use the same Document Action that original payment.

SusanneCalderon:
Thank you @e-Evolution , I will correct it.

e-Evolution:
fixes

Issue #1676, opened 2018-04-30
Enhancement adds a new action to generate a manufacturing order from a project line.


e-Evolution:
This new feature allows to generate Manufacturing Order and integrate project functionality with MRP and  CRP

marcalwestf:
hi @e-Evolution  Victor, this seems an interesting enhancement.
I suppose the enhancement you suggest is to create a Manufacturing Order where there is a Product containing a BOM in the Project. This would be in the Project Phase and Project Task.

If this is so, this would be great, as we are about to start committing to the Constrution vertical (https://github.com/SusanneCalderon/adempiereSusanne/tree/feature/ConstructionIndustry).

One step here is to create a BOM for the Project Phase or Project Task which contains a Product with a  default verified BOM. This new BOM would be a copy of the default BOM, only for the Phase or Task and it would be editable.  Here a link to the project, phase and/or task would do the job.

If combined with the enhancement you describe here, we would be a step further. I imagine that the Manufacturing Order to be created would take the BOM created as explained before and use the copied and edited BOM.

e-Evolution:
The BOM and Workflow Manufacturing was adding to Standard Phase and Standard Task , so when a project use a Project Type , the Product , BOM and Workflow Manufacturing are set for a Project Phase or Project Task.

So, you can Create a Manufacturing Order from Phase or Task, this process create a Manufacturing Order link to Project , Phase , Task and a Project Line link to Manufacturing Order. so the relation is completed and bidirectional.

PR #1986, opened 2018-09-10
#1676 Add new action to generate a manufacturing order from a project line

e-Evolution:
integrating #1676

### DIFF
--- a/base/src/org/compiere/model/I_C_Phase.java
+++ b/base/src/org/compiere/model/I_C_Phase.java
@@ -62,6 +62,21 @@ public interface I_C_Phase
 	  */
 	public int getAD_Org_ID();
 
+    /** Column name AD_Workflow_ID */
+    public static final String COLUMNNAME_AD_Workflow_ID = "AD_Workflow_ID";
+
+	/** Set Workflow.
+	  * Workflow or combination of tasks
+	  */
+	public void setAD_Workflow_ID (int AD_Workflow_ID);
+
+	/** Get Workflow.
+	  * Workflow or combination of tasks
+	  */
+	public int getAD_Workflow_ID();
+
+	public org.compiere.model.I_AD_Workflow getAD_Workflow() throws RuntimeException;
+
     /** Column name C_Phase_ID */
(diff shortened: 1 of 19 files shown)


marcalwestf:
start testing

marcalwestf:
First: don't forget to run a _Role Access Update_!!! Otherwise, the process won't be displayed on the task bar (gear icon).

When a Product with a verified BOM is set in a Project Phase, it is possible to fire the process _ProjectGenerateManufacturingOrder_.
20181001_proyecto_om_03


But, if the fields _Start Schedule_ or _Finish Schedule_ of either _Phase_ or _Task_ aren't set, the following (not meaningful) error message is issued, and the user has not a clue what is wrong nor how to overcome it:
20181001_proyecto_om_04

The problem lies in the following lines
 ```
atomicDateOrdered.set(dateOrderedTask.orElse(dateOrderedPhase.get()));
 atomicDatePromised.set(datePromisedTask.orElse(datePromisedPhase.get()));
```
which expects that the dates are set on Phase/Task.

Once set the dates, the process works correctly and a Manufacturing Order is generated:
20181001_proyecto_om_05

Thus, to overcome this, either the error has to be caught  with a meaningful message for the user to know what's missing or the fields must be set to "mandatory".

**Comment by marcalwestf** [comment]
Same error: exception thrown when _Start Schedule_ or _Finish Schedule_ from project/phase/task are not set.

```
 atomicDateOrdered.set(Optional.ofNullable(dateOrderedTask.orElse(
                    Optional.ofNullable(dateOrderedPhase.get()).orElse(project.getDateStartSchedule()))));
atomicDatePromised.set(Optional.ofNullable(datePromisedTask.orElse(
                    Optional.ofNullable(datePromisedPhase.get()).orElse(project.getDateFinishSchedule()))));
```

If this is the case, an Exception should be thrown.
It is caught later in the program, but too late.

marcalwestf:
start testing...

marcalwestf:
with the latest changes when there is no _Start Schedule Date_ in Task, Phase or Project, there is a message:
20181008_mo_01

Same happens when there is no _Finish Schedule Date_ in Task, Phase or Project.

PR #1987, opened 2018-09-10
#1985 Add an action in Project Task to generate a Sales Order

### LINKED ISSUE #1985: Add an action in Project Task to generate a Sales Order
Generate Sales Order from Project Task of same way that Project Phase

e-Evolution:
Fixes #1985

### DIFF
--- a/base/src/org/compiere/model/I_C_Phase.java
+++ b/base/src/org/compiere/model/I_C_Phase.java
@@ -62,6 +62,21 @@ public interface I_C_Phase
 	  */
 	public int getAD_Org_ID();
 
+    /** Column name AD_Workflow_ID */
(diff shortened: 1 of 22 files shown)


marcalwestf:
starting test

marcalwestf:
I get the following error
20181001_so_01

marcalwestf:
The reason for the exception: the code supposes that every line has a product. This isn't correct, as it maybe records without products.

Thus, when a record (project, phase, task) without a product is found, the process _ProjectPhaseGenOrder_ tries to create an _Orderline_ for a product with **ID=0**, which generates the Exception.

Changing the code from 
`tasks.stream().forEach(fromTask -> {`
to
`tasks.stream().filter(task -> task.getM_Product_ID() !=0).forEach(fromTask -> {`

Solves the problem.

This is the case for tasks. Same must be done for project and phases.

PR #2029, opened 2018-09-27
#2028, fixed read only logic for POS order, reference to issue:

### LINKED ISSUE #2028: POS: read only for control buttons is unnecessary
Currently I want modify a line of POS order, I follow a steps:

 - Create a POS order (From POS).
 - Add lines
 - Go to first (it can be over any line) and set quantity to 0 (Zero)
 - Wow.... The buttons for control are disabled for current line
screenshot_20180926_222858
screenshot_20180926_222847

yamelsenih:
https://github.com/adempiere/adempiere/issues/2028

fixes #2028 

Hi guys, it issue fixed a error know in POS ui, see the follow steps:

 - Create a POS order (From POS).
 - Add lines
 - Go to first (it can be over any line) and set quantity to 0 (Zero)
 - Wow.... The buttons for control are disabled for current line

For more info see issue

### DIFF
--- a/org.adempiere.pos/src/main/java/base/org/adempiere/pos/service/CPOS.java
+++ b/org.adempiere.pos/src/main/java/base/org/adempiere/pos/service/CPOS.java
@@ -106,10 +106,10 @@ public class CPOS {
 	private MBPartner 			partner;
 	/**	Price List Version		*/
 	private int 				priceListVersionId;
-	/**	Price List Id		*/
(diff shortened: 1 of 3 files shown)


marcalwestf:
... I think this comes from the time when if the quantity was =0 the line was deleted,

yamelsenih:
Yes, can be a improve, however is a bug that when a quantity is 0 the current panel ara disabled

Issue #1953, opened 2018-08-29
Add CSV as report output type


adaxa-steven:
Adaxa made a change to the report output choice so that you could preselect CSV in addition to excel, html and pdf.

It let large reports of order detail and accounting fact run in half the time of excel reports and reduced the times when no report was produced due to various memory(?)  issues.  

I don't think it affects anything else.  If you think it worth adding we can provide the changes.

Regards 
Steven

marcalwestf:
... getting the classes and XML files from Adaxa to start testing....

marcalwestf:
@adaxa-steven  Steven, could you check if the title corresponds to the code changes?
As I see in the commit, it is added an additional export format "CSV", but it is not set to become the default instead of PDF.

Changes seem otherwise OK, I will add the Reflist translation and test.

adaxa-steven:
"but it is not set to become the default instead of PDF"
I dont think w set it as a selectable default, just another option in the dropdown that you can pre-select before running the report.

PR #2035, opened 2018-09-29
#1953 Add CSV as report output type

marcalwestf:
Issue: https://github.com/adempiere/adempiere/issues/1953
Original code from Adaxa repository.


marcalwestf:
Result of changes:
20180929_csv_01
20180929_csv_02

adaxa-steven:
Expected result was a file named <Output_filename>.csv would be created and saved... Same as if xls was selected. Did that happen?

marcalwestf:
@adaxa-steven 

> Expected  result was a file named <Output_filename>.csv would be created  and saved... Same as if xls was selected. Did that happen?

It behaves similar to xls.

Issue #1772, opened 2018-06-13
New feature for Dynamic Tree


yamelsenih:
Currently exist a branch for handle dynamic tree https://github.com/adempiere/adempiere/tree/feature/ADEMPIERE-442-Dynamic-Tree-Functionality
i propose integrate it with develop branch, is nice to have any table as tree table and many hard code will be removed

yamelsenih:
Here a document for functionality: [http://wiki.adempiere.net/Spin_Contribution:_Dynamic_Tree:_Configuration_and_Use](http://wiki.adempiere.net/Spin_Contribution:_Dynamic_Tree:_Configuration_and_Use)

yamelsenih:
This change will be affect:
 - Main Menu for login
 - Tree Maintenance
 - Verify Tree Process
 - Tree wiondow
 - Account Element
 - Financial Report
 - New Client Definition

PR #2040, opened 2018-10-02
#1772 New feature for Dynamic Tree 

e-Evolution:
https://github.com/adempiere/adempiere/issues/1772 
- Fixed error with sequence menus 
- Fixed error to load tree based on tree type and next table id


e-Evolution:
fixes

PR #2063, opened 2018-10-07
#1975 ZK Process Panel KeyListener remains active when window is not visible.

### LINKED ISSUE #1975: Key Listeners in ZK remain active when panel is not displayed.
To replicate:

1. In webui, login as GardenAdmin.
2. Open process Generate Invoices, fill in mandatory fields but do not execute.
3. Open window Sales Order, click New, select Business Partner and enter "Joe Block", hit Enter.
4. Generate Invoices process starts!?

mckayERP:
Fixes #1975.

Fixes error where the key listener in the ZK Process Panel remains active when the window is not visible.  Example:

1. In webui, login as GardenAdmin.
2. Open process Generate Invoices, fill in mandatory fields but do not execute.
3. Open window Sales Order, click New, select Business Partner and enter "Joe Block", hit Enter.
4. Generate Invoices process starts

The PR removes the key listener and replaces it by setting the OK button setFocus() method.  This traps the enter key and generates the correct behaviour.  The OK button has the focus when the process window is first opened or when a saved parameter set is selected (user selects an item and hits enters or tab.)

See #1975 for more details.


e-Evolution:
fixes

PR #2066, opened 2018-10-08
#2062 Prevent duplicate key errors when running processes with saved parameters.

### LINKED ISSUE #2062: Error running Aging report with saved parameter.
In ZK, GardenAdmin role, run Aging report with current date.  Result OK.
Close, reopen. Set parameters to today's date, check Account Date.  Enter "Today" in the save parameter field.  Run the report.  Result OK.
Close and reopen the aging report.  Select "Today" in the saved parameter field.  Parameters are updated correctly.  Click the OK button and this error appears.

    ** ERROR: duplicate key value violates unique constraint "t_aging_key" Detail: Key (ad_pinstance_id, c_bankaccount_id, c_bpartner_id, c_currency_id, c_invoice_id, c_invoicepayschedule_id, c_order_id, c_cashflow_id)=(1000640, 0, 117, 100, 0, 0, 1, 0) already exists.

Close and reopen the aging report.  Select "** Last Run **" - nothing changes in the parameter panel.  Click OK and the report is generated with no parameters.

mckayERP:
Fixes #2062 .

The Process Controller does not create new process instances when using a saved parameter set. This causes a duplicate key error in the constraints of the temporary table for some processes.  The change ensures process instances are not deleted or reused from the saved parameter instances.  Whenever a saved parameter set is selected, the saved instance is deleted - meaning its name is set to null and it is removed from the list - and a new instance is added using that name.

### DIFF
--- a/base/src/org/compiere/apps/ProcessController.java
+++ b/base/src/org/compiere/apps/ProcessController.java
@@ -61,6 +61,7 @@ import org.compiere.util.Util;
  * 		@see FR [ 566 ] Process parameter don't have a parameter like only information</a>
  *	@author Michael Mckay michael.mckay@mckayerp.com
  *		<li>BF [ <a href="https://github.com/adempiere/adempiere/issues/495">495</a> ] Parameter Panel & SmartBrowser criteria do not set gridField value
+ *		<li>BF [ <a href="https://github.com/adempiere/adempiere/issues/2062">2062</a> ] Prevent reuse of AD_PInstance_ID when selecting saved parameters
  *  @author Raul Muñoz, rMunoz@erpcya.com, ERPCyA http://www.erpcya.com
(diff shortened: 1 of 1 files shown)


marcalwestf:
I can reproduce the error following the explanation of #2062 :
20181026_error_process_parameters_01

marcalwestf:
... I cannot test the changes since I get an error I'm clueless at.
Perhaps it is because I am testing with an IDE or maybe the latest merge for this commit is from 3 weeks ago and something has changed inbetween. Maybe also the Tomcat-server configuration at my IDE.

Anyway, I put it here for documentation's sake; perhaps somebody can test the changes without hitting this problem:
20181026_error_process_parameters_02

marcalwestf:
...  a merge to the newest commit would be desirable just for testing without having problems at log in.

Issue #1348, opened 2017-09-29
Implement Reverse Accrual for all document that generate account fact


e-Evolution:
Currently, the system when doing a reverse correction is done validating if the period is open and with the date of the original document, implement the reverse accrual using the date of the current period.

It is necessary to take forecasts for multi-currency documents since to revert a loss or gain in exchange, should be with the original date of the document so that the reverse is correct.

My proposal to solve this issue is simple we can to add the currency rate in any document and using this to calculate the reverse a loss or gain in exchange.

e-Evolution:
First Stage Implement Reverse Accrual for Shipment / Material Receipts ,  Customer Invoice , Vendor Invoice and payment allocations.

Tests to be performed

Purchasing process

1.- Create Purchase Order , Created Material Receipt based on PO. and Create Vendor Invoice based on  Created Material Receipt , Reverse Correction Vendor Invoice , Reverse Correction Material Receipt.

2.- Create Purchase Order , Created Material Receipt based on PO. and Create Vendor Invoice based on  Created Material Receipt , Reverse accrual Vendor Invoice , Reverse accrual Material Receipt.

3.- Create Purchase Order , Create Vendor Invoice based on PO, Created Material Receipt based on vendor invoice,  Reverse Correction Material Receipt and next Reverse Correction Vendor Invoice.

4.- Create Purchase Order , Create Vendor Invoice based on PO, Created Material Receipt based on vendor invoice,  Reverse Accrual Material Receipt and next Reverse Accrual Vendor Invoice.

5.- Generate the step 1 to 4 with same Purchase Order.

6.- Generate Sales Order , Generate Shipment and Generate Customer Invoice , Reverse Correction the customer Invoice ,  Reverse Correction  of Shipment 

7.- Generate Sales Order , Generate Shipment and Generate Customer Invoice , Reverse Accrual the customer Invoice ,  Reverse Accrual  of Shipment

adaxa-steven:
"My proposal to solve this issue is simple we can to add the currency rate in any document and using this to calculate the reverse a loss or gain in exchange."
assume exchange rate 1 june  EUR1.00 = USD1.10   and then 1 July becomes EUR1.00 = USD1.20
we have eur 100 in the bank  with accounted value of usd110
we make a payment of EUR100 and now have zero in the bank. (assume the payment is not allocated)

we reverse correct the payment on 2 July
we now have eur100 in the bank but its value is USD120

[...]

PR #2068, opened 2018-10-08
#1963 Wrong Behavior when completing an allocation with date validation

### LINKED ISSUE #1963: Bug Behavior  incorrect when complete an allocation and  dates validation is executed
The current logic indicate the that the invoice accounting date or pay accounting date should be after that allocation accounting date.
when the behavior rigth is that allocation accounting date should be after to accounting date or pay accounting date

The behavior was introducing #1348.

e-Evolution:
Fixes #1963

- The behavior was introducing #1348.

(cherry picked from commit afa63988e14a18d34c6ba8aac525601c43a94cf4)

### DIFF
--- a/base/src/org/compiere/model/MAllocationHdr.java
+++ b/base/src/org/compiere/model/MAllocationHdr.java
@@ -453,7 +453,7 @@ public final class MAllocationHdr extends X_C_AllocationHdr implements DocAction
 								}
 								if (allocationLine.getC_Invoice_ID() > 0) {
 									I_C_Invoice invoice = allocationLine.getC_Invoice();
-									if (invoice.getDateAcct().after(getDateAcct())) {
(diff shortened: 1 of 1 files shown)


marcalwestf:
Hi @e-Evolution  Víctor,
With the new code as you suggest, it seems that write-off would be considered as error because at write-offs the invoice date is always before allocation date. 

I think it would be best to elaborate a table with all possible cases.

e-Evolution:
@marcalwestf if you check of code https://github.com/adempiere/adempiere/pull/2073/files the validations is incorrect because required that the Invoice Date is after that Allocation Date or Payment Date is after the that Allocation Date , when in real world the Allocation Date always is after the invoice or payment.

with current implementation my customers need are change the allocation date with before the invoice date and payment date this incorrect

e-Evolution:
Reverse this commits

PR #2073, opened 2018-10-10
#2072 Dashboard content User favorites is not translated

### LINKED ISSUE #2072: Dashboard content User favorites is not translated
The Dashboard Content uses the menu name, instead of the translation when the user entered with another language than base language.

SusanneCalderon:
fixes #2072 
Here I entered with language _es_MX_
imagen

### DIFF
--- a/base/src/org/compiere/model/MTreeFavorite.java
+++ b/base/src/org/compiere/model/MTreeFavorite.java
@@ -59,8 +59,10 @@ public class MTreeFavorite extends X_AD_Tree_Favorite
 	 */
 	public void loadNode(int AD_Tree_Favorite_ID)
 	{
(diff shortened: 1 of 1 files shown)


e-Evolution:
@SusanneCalderon , Great, this problem is from integration , I had a question what happens if not exist the translation , I think that should solve using base language , so using other approach can be a good idea , you can use a COALESCE (t.name , nodename) AS nodename function, this way if not exist translation the menu show base language.

kind regards

e-Evolution:
@SusanneCalderon with new approach yet need this code?

+					MMenu mMenu = new MMenu(Env.getCtx(), menuID, null);	
+					if (Env.getLanguage(Env.getCtx()).isBaseLanguage()) {
+						name = mMenu.getName();
+					}
+					else {
+						name = rs.getString(8);
+					}
 					img = mMenu.getAction();

PR #2076, opened 2018-10-11
#2074 Error launching jasper reports print format from any tab

### LINKED ISSUE #2074: Bug when try launch a jasper report print format from any Tab
1.- Create a new process with your jasper report form , ie any invoice formant 
2.- Create a new print format as form link with new jasper report process
3.- Use search button to gen the invoice that need to print.
4.- Select your new print format to show the print invoice 

The Report Engine only show black screen

image

e-Evolution:
Fixes #2074

(cherry picked from commit 22708e04f8251b66bea66af1306e4504b8d52405)

### DIFF
--- a/zkwebui/WEB-INF/src/org/adempiere/webui/panel/AbstractADWindowPanel.java
+++ b/zkwebui/WEB-INF/src/org/adempiere/webui/panel/AbstractADWindowPanel.java
@@ -1827,7 +1827,7 @@ public abstract class AbstractADWindowPanel extends AbstractUIPart implements To
         }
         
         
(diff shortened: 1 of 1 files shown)


marcalwestf:
@e-Evolution hi Víctor, just a tiny suggestion:

Changing the _if-clause_ to
`if(currentTab.getWhereClause() !=null && !currentTab.getWhereClause().isEmpty())`
		
don't you think that the _if-clause_ that follows will be obsolete?
`if(currentTab.getWhereClause().length()>0)`

Due to the fact that passing the first clause will obviously cause the second to pass.

So I'd suggest to eliminate the second _if-clause_.

PR #2078, opened 2018-10-11
#213 Add ability to automatically synchronize table and column changes with the database.

### LINKED ISSUE #213: Support for application dictionary changes and configurable automatic syncing with the database
I've had to modify a large chunk of an existing customization and found that ADempiere does not support renaming of columns.  In the database, if you rename a column and sync, you will get the old column with data and a new empty column.  Fixing this is time consuming and involves a lot of work in the database - a tedious task when facing changes to hundreds of columns.

I came up with a solution and applied it to about three hundred columns and about twenty tables.  I found it quite helpful.  With autosyncing, any change made to a table, element or column is immediately applied to the database in a way that no additional work is required.   It works well with the migration scripts which can record and keep track of the changes.

The main indexes and constraints are preserved through the table/column changes.  Special constraints may need to be managed manually.  The deletion of a column is not supported.

As an example, with this feature, I can change the column name of a system element and have the changes synced with the database in all the tables where that element occurs without having to visit each table and manually sync.

If this would be helpful, let me know and I'll make a pull request with the changes.

mckayERP:
Fixes #213.

Enable database autosync. For example, a change to a system element could be made that changes a column name and that change would be made in all affected tables in the database without the need to visit each column and run the ColumnSync process.

Also included is the filling of mandatory columns with default data if the column is added to an existing table.  A typical example is the addition of a mandatory yes/no field with a default of N.

There are some  limitations:
- check constraints are not renamed. 
- does not update software models or references to the table/column in views or queries.

This enhancement is really aimed at making it easier to create tables and columns on the fly during development.

The auto sync is configured to run as default but a system configurator entry could be added to turn it off.

Test as follows:

- Open your database admin tool to watch the changes.
- Log into ADempiereas system or superuser in sysadmin role
- Create a test element for a dummy column, say 'Test_Element'
- Create a new test table called 'Test_Table'. Save the record. Note the columns, especially the key column, are created. Do not click the Synchronize button.
- Check in the database admin tool that the table has been created and all the default columns exist.
- Add a column to the table with your test element. Save the record.
- In the database admin tool note that the column has been added to the database table.
- In ADempiere, change the table name. Note that the key column name has also changed.
- Check in the database admin tool that the table name has changed and that the name of the key column and associated constraint has also changed. A table with the original name should not exist.


e-Evolution:
Hi Mike, @mckayERP 

This seems a good improvement, but it also implies a great risk, generally you can make changes in the dictionary without synchronizing with the DB, for example in the database a value can be null but in the dictionary it is mandatory and that is fine in special when this is enabled in a database with record history with null values.

So the risk I see is that if you make a mistake, it also synchronizes. (a big risk for production environment)

As I am developing,  I usually build the model, prototype from domain model (Table and Columns), windows ,  tabs, fields, once what I see on the screen is fine, then at that moment, I synchronize with the database, this is why in all that process you can have a lot of changes as kind of data you are using, which of them are mandatory and which are not, what will be their default values, etc.

Once all this is well defined I synchronize with the database.

In case you have movements after synchronizing with the DB, as I am in a development environment, I simply delete the generated table and synchronize again, as simple as that.


So for me, it would be more useful a process that does that from the dictionary erase table from the database and synchronize same time.

kind regards
Victor

marcalwestf:
@mckayERP Hi Mike, 
taking up the comments by @e-Evolution Victor, can we perhaps combine the best of both worlds?

This could be achieved by defining for example a variable in _AD_Sysconfig_.

Depending on which value the variable has, it'd behave as it does now (I'd say the default value for backward compatibility) or as you suggest.

A simple value change in _AD_Sysconfig_ would arm the Application Directory to behave either way.

mckayERP:
@marcalwestf Its setup for the system config.  Right now, it defaults to 'Y' or enabled.  I could change it to default of 'N' or disable it by default and add an XML to add the record to the System Config.

@e-Evolution Thanks for the description of how you use the AD.  In my case, it was a redesign of many tables and columns based on changing the entity type.  I made the changes on a copy of the production system and applied them using the Migration Tool.  The code in this pull-request made it a far less of a daunting task.  You're correct that its not intended for use on a production system but is a development tool.

Instead of delete, the sync scripts use ALTER TABLE to make the changes.  Currently, if you create a table, sync and then change the name of the table or column and resync, you'll have two copies.  With alter table, it just makes the change, preserving any data.  Deleted records are an issue - those tables and columns are still left in the database.  The code also doesn't deal with all constraint types so some manual work is still be required from time to time.

For the setting of defaults, that would only happen if the record was not mandatory and it is changed to a mandatory field with a default assigned OR a new mandatory field is added to an existing table.  With a null value in the field, I wasn't able to edit the other field and save the record without the mandatory error.  We recently fixed this for Y-N fields but it is still an issue for other types.  

I take your point about history records though and will have to think about that.

BTW - I'm slowly adding Pull Requests for a bunch of development done on a custom system over the last three years.  Some of it may not be needed in the core, but I'll publish it as contributions and let you have a look.

marcalwestf:
> does not update software models or references to the table/column in views or queries.

Does this mean that if I change a column name used in a View with the process you suggest, this would be changed in the database but wouldn't be changed in the Application dictionary?

marcalwestf:
IMO this is a functionality a developer may find under certain circumstances useful.

As this functionality doesn't cover all eventualities, I do agree that it shouldn't be the default behavior and also that the developer must be totally aware of what (s)he's doing (well this should apply always, shouldn't it?).

mckayERP:
Shortly, I'll add a revision to the pull request with an XML to add the settings to the System Configurator and make the default disabled.  

Only new records or changes to regular table names or columns would be synced either auto or manually.  For tables, its strictly the name.

For columns, the following code in MColumn.afterSave() determines if the column data is sync'd or not:

    protected boolean afterSave (boolean newRecord, boolean success)
    {
        // #213 - AutoSync - check if we should auto-sync the column and table.
        if (isAutoSync()
                && (newRecord
                || is_ValueChanged(MColumn.COLUMNNAME_ColumnName)
                || is_ValueChanged(MColumn.COLUMNNAME_AD_Reference_ID)
                || is_ValueChanged(MColumn.COLUMNNAME_FieldLength)
                || is_ValueChanged(MColumn.COLUMNNAME_IsMandatory)
                || is_ValueChanged(MColumn.COLUMNNAME_IsKey)
                || is_ValueChanged(MColumn.COLUMNNAME_DefaultValue)
                || is_ValueChanged(MColumn.COLUMNNAME_ColumnSQL)
                || is_ValueChanged(MColumn.COLUMNNAME_IsParent))) 
        {
        	syncDatabase((String) this.get_ValueOld(COLUMNNAME_ColumnName));
        }
		
        //	Update Fields
        if (!newRecord)
        {
        ...

So, if the sync was enabled, you could change create a table and have it sync with the default columns. Then change the table name and have the table and ID column sync'd along with the primary key.

If you edit an element to change a database name, all columns affected would be synced.

There is no syncing with parameters, fields, views, or virtual columns.

e-Evolution:
Hi  Mike @mckayERP

I think it's appropriate to add a flag more in preferences as this is now work Log Migration Script and Dictionary Maintenance , something like Synchronize DDL automatically

image

yamelsenih:
Hi All, is nice if it can be used as a process, for me i have the follow:

- Add tables
- Add Tabs and windows
- Add changes, modify tables and columns

The last process can be: **Synchronize all entity** and it can be good

I don't have problem with add this change to core "**If it is optional and default is false**", is nice if can be a process with table or entity for synchronize.

Best regard

yamelsenih:
Also i think that is a good moment for create a process that can be called from **beforeSave** method on column or when is synchronize from column or from table or menu and it can be useful if I forgot if a column is created or changed.

mckayERP:
Only problem with a process run as a batch is the old name of the column/table is lost and the sync will result in a new column/table being created.

yamelsenih:
Hi @mckayERP I think that is nice to have a process from menu for synchronize all tables instead it.

For now this pull request make conflicts with MColumn.

Can you update your branch with last changes?

Best regard

mckayERP:
Hi Yamel,

I'll add the process to the menu.  I'll let you know when its ready.

Mike

mckayERP:
Fixes #2428 

Had some problems with the process but it is working now.  I added columns to AD_Column and AD_Table to identify which table/column needs to be synchronized with the database.  It can handle name changes with some limitations:

- If renaming an existing table/column, the new table/column name can't already exist - don't want to clobber existing data.
- Multi-key tables may need some manual intervention to rename keys.
- Inline constraints won't be renamed.
- If there are other more complex constraints, these may need to be manually renamed.

In adding columns to AD_Column and AD_Table I ran into an issue with the migration scripts where a database error occured. See #2428.  This pull request also solves this issue by explicitly naming the columns in POInfo.loadInfo() via the database metadata.  The list of columns is loaded on the first call and cached.  The list is resent when a column in AD_Table or AD_Column is synchronized.

I'll update the docs and post a link here shortly.

mckayERP:
Just found another bug - needs more testing.  I changed the label back to Waiting for User Input.

yamelsenih:
Hi @mckayERP already exist a verification for know if synchronization is required here:
<pre>
column.syncDatabase();
boolean currentNotNull = DatabaseMetaData.columnNoNulls == rs.getInt("NULLABLE");
				int currentDataType = rs.getInt("DATA_TYPE");
				int currentColumnSize = rs.getInt("COLUMN_SIZE");
				String currentColumnDef = rs.getString("COLUMN_DEF");
				String columnDef = getDefaultValueSQL();
				if(!DisplayType.isSameType(this, currentDataType, currentColumnSize)
						|| isMandatory() != currentNotNull
						|| (currentColumnDef != null 
							&& columnDef != null 
							&& !currentColumnDef.startsWith(columnDef))) {
					sql = getSQLModify(table, isMandatory() != currentNotNull);
				}
</pre>

The util method **DisplayType.isSameType** is used for it.

Is very complicated if is added a column for sync control because a DB user can make changes for DB and if RequireSync is false then never it can sync.

mckayERP:
Hi Yamel,

Thanks for the method.  I can use that for consistency.

There always need to be a way to force the sync and this happens as it does now - by clicking the button in the Column tab of the Table and Column window.

On adding columns to the table, here is what I was thinking.

The current sync method don't deal with name changes. Also the user has to specify which column to sync. Table name changes and column name changes are allowed but when you sync, a new table/column is created.  ID columns are not handled at all. For example, if you change "OldTable" to "NewTable" and then sync any column in the table, you will get a new table called "NewTable" but the id column will still be OldTable_ID.  You will also have a table "OldTable" that still has data.

If you want a process to synchronize columns, then you need a way to identify which ones to sync.  I don't think its practical to sync everything.   By adding the columns to track the name change and the need for a sync to AD_Column and AD_Table, you can track this info and sync an any time.

The user can always force the sync in the Column tab.

This needs more work - so wait until I change the issue lable before testing again.

Thanks.

Mike

mckayERP:
@yamelsenih 
Hi Yamel,

For the code that tests if an update is necessary, I'm running into a problem with the default values that use SQL or variables.  For example, see C_Invoice.IsSOTrx.  In the AD, the column is mandatory with a default value of @IsSOTrx@ but the database has the column set as Not Null with a default of 'Y'::bpchar.  This difference triggers a sync with the alter sql as 

    ALTER TABLE C_Invoice MODIFY IsSOTrx CHAR(1) DEFAULT NULL

This fails with a  org.postgresql.util.PSQLException: ERROR: Failed to recreate dependent view error.

The best I can do for a case like this is flag it for manual intervention.

Any ideas?

Mike

mckayERP:
I've added the ability to compare the application dictionary with the database structure and report on the differences or make changes.  Table and column name changes are supported.

The error/info reporting during the  synchronization has also been increased.

When I run it on the current seed there are several hundred columns that are out of sync with the database.  Most are mandatory columns where there are rows with null values in that column.  There are several where the data type is mismatched and many where the AD column is not a key or parent link but is part of a primary key constraint in the database.  Automatically fixing these is not always possible.

For documentation on the change, see the draft https://mckayerp.gitbook.io/adempiere/v/contrib_213/system-administration/the-application-dictionary/synchronizing-changes-with-the-database.


There are three ways to test it.  First, verify that the System Config for DATABASE_AUTO_SYNC is set to 'N'.

A: use the Column Sync Button
1. Create a custom element
2. Create a new table
3. Add a custom column using the custom element
4. Synchronize the custom column.  Verify that the table and all columns exist in the database.
5. Change the custom element column name.
6. Synchronize the column again.  Verify that the column name has changed in the database.
7. Change the table name. Synchronize the column again.  Verify that the table name, key column name, primary constraint have all changed in the database.

B. Use the Sync All process.
1. Repeat steps A.1 to A.3.
2. Run the Sync All process, selecting only "Only Known Changes"
3. Verify the results.
4. Repeat steps A.5 and A.7, running the Sync All process and verifying after each change.

C. Auto Sync
1. Set the System Config for DATABASE_AUTO_SYNC to 'Y'.
2. Repeat steps A.1 to A.3, A.5 and A.7.  Verify that the database changed after each save.

Let me know if you have any questions with it.

Mike

yamelsenih:
Hi @mckayERP  can you resolve conflicts?

I can review it after change.

Best regard

mckayERP:
@yamelsenih Hi Yamel, Sorry I've been away for a bit.  I'll fix the conflict over the weekend and let you know.

mckayERP:
@yamelsenih Sorry, that was a long weekend.  The conflicts are fixed and the Pull Request is merged with 3.9.3.  There is one XML file to run on the 3.9.3 seed prior to testing.  Let me know if you find anything.

yamelsenih:
Hi @mckayERP I will test it

yamelsenih:
Hi @mckayERP I have some comments about this pull request, for me you can change it and is ready for merge:

- MTable class have many changes from older versions: I think that is better if you use the last version as seed and compare this class whit your changes for merge, the follow is a example:
Screenshot_20200113_115602
- Can you use a name more explicit for it? **SyncAll**: maybe can use SynchronizeAllColumns.

Best regard

mckayERP:
@yamelsenih Hi Yamel,  I reverted all the previous commits to MTable and resynced with the development branch, then added all the changes back in one commit.  You should be able to see a clean difference.  I still have to fix the name so wait for that please.

mckayERP:
@yamelsenih I changed the name of the SyncAll process to SynchronizeAllTablesAndColumns.  Your right - its much clearer.

Ready for final review.

Mike

yamelsenih:
Thanks @mckayERP I will review it

PR #2081, opened 2018-10-15
GenerateCostDetails does not delete details for inactive cost elements.

### LINKED ISSUE #197: GenerateCostDetails does not delete inactive cost types or methods

mckayERP:
Fixes #197.  

Changes allow GenerateCostDetails to delete cost details related to inactive cost elements. 

Cost Elements can be deleted and will delete the related cost dimensions and cost details.  

Only one active default cost element per client is allowed and only one element of each cost type per client is allowed.  

Inactive cost elements in GW client have been removed.


yamelsenih:
Hi @mckayERP exists a error with build, please update or change

mckayERP:
Can't seem to re-run the tests.  The failure was a time-out.  I'll cancel and resubmit the pull-request.

PR #1946, opened 2018-08-28
#1935 Add IsDefault to AD_User_Roles.  

mckayERP:
Fixes #1935
Limit the flag to one record per user. Sort the roles in Login.java so that the default is first in the results.

PR #2082, opened 2018-10-15
#1935, Add back Compatibility for column on login, reference to issue:

### LINKED ISSUE #1935: Default role for the System account should be System Administrator
The default role is "Role Template" - or at least that is what appears in the Role combo box on login.  This is confusing for new users.  The Role Template is a special case and the user needs some expertise to use it.

yamelsenih:
https://github.com/adempiere/adempiere/issues/1935

Hi all, this pull request resolve a support with back compatibilitu of last change merged on pull request https://github.com/adempiere/adempiere/pull/1946

Please note that it make a validation for know if exist a column for **AD_User_Roles**

fixes #1935 

Best regard.

### DIFF
--- a/base/src/org/compiere/util/Login.java
+++ b/base/src/org/compiere/util/Login.java
@@ -341,6 +341,13 @@ public class Login
 		}
 		KeyNamePair[] retValue = null;
 		ArrayList<KeyNamePair> list = new ArrayList<KeyNamePair>();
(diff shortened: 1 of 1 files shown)


marcalwestf:
reproduced the error when a new code runs on an old database.

Starting the revision on this pull request.

PR #2088, opened 2018-10-17
#421 Ensures embedded tabs are updated

### LINKED ISSUE #421: Embedded tab is not updated
I have a sales order with an embedded tab for the order lines.

When searching for an order, the embedded tab is not updated, leading to confusion.
1. Open the window in form view and navigate to the first record.  Note the contents of the header and embedded tab.
2. Switch to spreadsheet view and click on a column header to sort the data differently. Select the first record.
3. Switch back to form view. Note that the header is correct but the contents of the embedded tab are those of the record in 1.
4. Navigate to the 2nd record and back and the correct contents of the embedded tab appear.

mckayERP:
Fixes #421.

In a window tab that has an embedded tab, like the Bill of Materials & Formula window, its possible to have the embedded tab and header record out of sync when there are several header records.  This happens when the header is put in grid view and then sorted differently by clicking on a column. On switching back to the form view, the embedded tab will show records associated with the original header, not the currently displayed header.

The fix is simply to refresh the tab data when the switch from grid to form view is made.

### DIFF
--- a/client/src/org/compiere/grid/GridController.java
+++ b/client/src/org/compiere/grid/GridController.java
@@ -159,7 +159,8 @@ import org.compiere.util.Util;
  * @contributor fer_luck @ centuryon  FR [ 1757088 ]
  * 
  * @author mckayERP www.mckayERP.com
- * 				<li> #283 GridController in swing will not set value to null in vetoableChange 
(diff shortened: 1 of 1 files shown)


marcalwestf:
... confirming described missing synchronization on window _Bill of Materials & Formula_.

Testing bug fix.

PR #2108, opened 2018-10-29
#2107 Validate Login User for Password reset

### LINKED ISSUE #2107: Forgot password not validate Login User
I have two users with the same user value:

- User 1:
  - Value: Test
  - Name Test User
  - Login User: Yes
- User 2:
  - Value: Test
  - Name Other User Test
  - Login User: No

When  I am trying reset password for user 1, a mail is sent to user 2 also, this is because both have the same user value but second user is not a login user

yamelsenih:
Fixes #2107 
Hi all, please take a minute for test it, is very important this change.


marcalwestf:
Good catch! 
I wonder if is correct that users are allowed to have identical values.

PR #2121, opened 2018-11-06
#2120 Add custom name for export file from payment print form

### LINKED ISSUE #2120: Add custom file for Payment Export Print
Currently when is **exported** a payment from **Payment Export/Print** is generated with generic name "paymentExport.txt"

I propose that it can be generated with a custom name, example:

**BankName_PaymentSelectionNo.txt
HSBC_010203.txt**

yamelsenih:
Fixes #2120 

Hi all, this pull request will be allows create a file for export payment with custom name, see a example: 

screenshot_20181106_155800
screenshot_20181106_155847

### DIFF
--- a/base/src/org/compiere/util/PaymentExportList.java
+++ b/base/src/org/compiere/util/PaymentExportList.java
@@ -17,8 +17,12 @@
 package org.compiere.util;
 
 import org.compiere.model.MBPBankAccount;
+import org.compiere.model.MBank;
+import org.compiere.model.MBankAccount;
+import org.compiere.model.MPaySelection;
 import org.compiere.model.MPaySelectionCheck;
 import org.compiere.model.MPayment;
+import org.compiere.model.MPaymentBatch;
 
 import java.io.File;
 import java.io.FileWriter;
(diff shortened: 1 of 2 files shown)


marcalwestf:
Run the commit on a newly installed instance. The result: file name doesn't change as expected.

Turn to debug on IDE.
When debugging , we stop at the place where the file should be changed.
20181109_export_print_01

Stepping into it, we notice **the variable file is null**; for that matter, the file name remains the same:
20181109_export_print_02

The variable _file_ must be initialized correctly for the file to get a new name.

yamelsenih:
Hi @marcalwestf the fileName shuld be changed inside CustomExportFile.java, your implementation can change file Name
see: https://github.com/adempiere/LVE/blob/master/financial/src/main/java/base/org/erpya/lve/bank/exp/Banesco.java#L89

https://github.com/adempiere/LVE/blob/master/financial/src/main/java/base/org/erpya/lve/bank/exp/LVEPaymentExportList.java#L167

You should use a method like it 
screenshot_20181109_191640

marcalwestf:
understand.

Issue #2159, opened 2018-11-28
Create Export for Surrogate Keys


yamelsenih:
This issue allows create a process for export all Surrogate Keys generated.

Note that this process must be runing after generate Surrogate Key #579

yamelsenih:
Here a gif
![current mp4](https://user-images.githubusercontent.com/2333092/49154387-a8a00e00-f2ee-11e8-9395-3c1bbd43b1fd.gif)

PR #2160, opened 2018-11-28
Feature/#2159 add export surrogate key

yamelsenih:
Hi all, this pull request allows create a Migration record based on Surrogate Keys generated.

For more info see: #2159 

Fixes: #2159


yamelsenih:
Here a gif
current mp4

PR #2178, opened 2018-12-06
Dashboard access Problem when executing role access update

### LINKED ISSUE #2079: Problem with Dashboard access when try execute role access update
Problem introducing  in #1891 #1893

The problem is caused by new table adding a new column AD_Dashboard_Access_ID as primary key and the logic to insert the security default permission not considering the primary key for table AD_Dashboard_Access

e-Evolution:
Fixes #2079

- Fix the Oracle compatibility using the nextidfunc function

(cherry picked from commit 5734d96ab99e17be499d0a55ca1f8708f508a834)


marcalwestf:
Test OK.
When conflict is solved, the PR can be accepted.

Issue #2031, opened 2018-09-27
Field LandedCostType in LandedCost


SusanneCalderon:
For documentation it is convenient to distinguish between different types of distributed costs.
According to the incoterms for any mode of transport we generated a referencelist _LandedCostType_
1. - EXW – Purchase Costs
2. - DIS - Discounts
3. - ECD - Export customs declaration
4. - CPE - Carriage to port of export
5. - UPE - Unloading port of export
6. - LPE -Loading on vessel/airplane in port of export 
7. - FRE - Freight costs
8. - INS -Insurance 
9. - UPI - Unloading in port of import 
10. - LPI - Loading in port of import 
11. - CPD - Carriage to place of destination 
13. - ICC - Import customs clearance
14. - IDT -  Import duties and taxes
15. OTH Others

Costs 1 to 6 include FOB, 1 to 8 CIF.
The table _C_LandedCost_  has a new field _LandedCostType_, mandatory, with a default value _Others_

e-Evolution:
@SusanneCalderon What is the difference of using a cost element?

Does the classification have any practical application or is it only for grouping information?

SusanneCalderon:
1. Example, costing method average invoice, 2 providers, first provider is foreign, second provider is local.
Using costelements for costing every costelement is calculated seperatly. So after the first purchase I have costs for import customs. When I purchase the next items from a local provider I have no additional costs for import customs, but in the calculation of the overall costs these costs are still considered and not updated, because I do not have any invoice for this cost element which could be distributed as landed costs.
To avoid these efffect I mix different cost types in 1 cost element, in this example material costs. 
But for reporting purpose I have to distinguish between the cost types.
2. This classification is serves just for reporting and grouping.

e-Evolution:
@SusanneCalderon , I understand, so you need a Landed Cost Type to filter the products that are linked with this landed cost, so I think that the best solution is to create a new table for M_LandedCostType, so any company can adding the landed cost type that they need. set as now a static list can be that is not applied for all case in all company or simply want management other elements.  so not should use you a list table because if is a way to order and group by landed cost type, the customer should have the freedom to create you landed cost type list with name and search key as they need. as now work the business partner group, product category, product group, product class, etc.

The Table based on the list are using when to implement a business logic for each element of the list as Delivery Rule, Invoice Rule

[...]

PR #2223, opened 2018-12-15
adempiere: #2031 Field LandedCostType in LandedCost

SusanneCalderon:
https://github.com/adempiere/adempiere/issues/2031
Correction of the XML files


SusanneCalderon:
This is the correction of the XML-files to generate the table C_LandedCosttype_ID

Issue #2231, opened 2018-12-20
Circular Dependencies in projects base - project - request


homebeaver:
Hi all - and merry christmas!
I have tried the development branch in my dev env (eclipse+windows) an I found some problems to be solved
- there are Circular Dependencies in 3.9.1 between `base - project - request`. Certainly you can solve it, see [here](https://stackoverflow.com/questions/48609810/how-to-solve-a-cyclic-project-dependency-in-eclipse-any-way-to-make-one-of-the). But it is no good practice to have cyclic dependences. A simple solution is to move project + request source to base.
- next: the lib `oracle.jar` is not defined in eclipse .classpathes - simple to solve
- some git projects are not visible in eclipse - the eclipse project files are missing - routine piece of work

That's all (hopefully) regards

marcalwestf:
thanks Eugen @homebeaver 
Will have a look at it.

osdevguy:
Hello Sir

How is it that this issue is not resolved or a fix is presented for this. This seem an inherently structural problem in the build and this is not prioritized over new functionality to be released in a new dev. As someone looking to get involved in the Adempiere project it seems that "own" efficiency is more important than "community" efficiency for all to benefit from having a release that works.

Is there a workaround for this or best to also fork and just run your own parallel system like many seem to be doing while the "core" is running for the benefit of a few.

Is there a document to show the governance of release structures of Adempiere project --- i.e. surely "build" is more important than "new functionality" in a release.

PR #2232, opened 2018-12-20
#2231 : here is a solution for cyclic projects and problems with eclipse

homebeaver:
Hi all,

as I suggested in #2231 the sources from `project + request` are moved to base. Thus there ara no circular dependencies anymore.
Also the minor problems with `oracle.jar` and the missing eclipse project definitions are solved.

I hope there are no side-effects for the IDEA-dev env!

Regards EUGen


yamelsenih:
Hi @homebeaver is a big change for source, exists over 2k files that will be changed.

Can you upgrade your source code and make a new pull request?

I can revew your new pull request.

Best regard

Issue #2247, opened 2019-01-08
ERROR: function getuuid() does not exist when start RUN_MigrateXML.sh


rcapecce:
I download the new release of adempiere.net and start the RUN_MigrateXML.sh with a 390 database and I'm getting the next error:

===========> DB.getSQLValueString: SELECT getUUID() [1]
org.postgresql.util.PSQLException: ERROR: function getuuid() does not exist
  Hint: No function matches the given name and argument types. You might need to add explicit type casts.
  Position: 8; State=42883; ErrorCode=0
        at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2440)
        at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2183)
        at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:308)
        at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:441)
        at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:365)
        at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:143)
        at org.postgresql.jdbc.PgPreparedStatement.executeQuery(PgPreparedStatement.java:106)
        at com.mchange.v2.c3p0.impl.NewProxyPreparedStatement.executeQuery(NewProxyPreparedStatement.java:76)
        at sun.reflect.GeneratedMethodAccessor1.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.compiere.db.StatementProxy.invoke(StatementProxy.java:100)
        at com.sun.proxy.$Proxy2.executeQuery(Unknown Source)
        at org.compiere.util.DB.getSQLValueStringEx(DB.java:1376)
        at org.compiere.util.DB.getSQLValueString(DB.java:1419)
        at org.compiere.model.PO.generateUUID(PO.java:4236)
        at org.compiere.model.PO.saveNew(PO.java:2758)
        at org.compiere.model.PO.save(PO.java:2194)
        at org.compiere.model.PO.saveEx(PO.java:2278)
        at org.eevolution.service.dsl.ProcessBuilder.generateProcessInstance(ProcessBuilder.java:160)
[...]

PR #2248, opened 2019-01-09
#2247, Fixed error with getUUID function

yamelsenih:
This pull request resolve a problem with old compatibility for ADempiere difference of versions

Add old compatibility using java UUID, reference to issue:
https://github.com/adempiere/adempiere/issues/2247

Fixes: #2247

### DIFF
--- a/base/src/org/compiere/db/CConnection.java
+++ b/base/src/org/compiere/db/CConnection.java
@@ -225,6 +225,8 @@ public class CConnection implements Serializable, Cloneable
 
 	/** Database Type       */
 	private String 		m_type = "";
+	/**	Supported DB UUID		*/
+	private boolean 	isSupportedUUIDFromDB = false;
 
 	/** Database Host       */
 	private String 		m_db_host = "MyDBServer";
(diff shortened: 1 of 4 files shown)


e-Evolution:
Hi everybody

the implementation of "Surrogate key" from 3.9.1 is mandatory, so the facilities must use the database that has support for this, both Oracle and Postgresql.

The reason for this is that from the version of 3.9.1 conflict resolution and change control for application dictionary and records, replication or export and import of data will be resolved by UUIDs, so it is not recommended to use ADempiere 3.9.1 without Surrogate key since this can bring future synchronization problems or the impossibility of future migrations.

In case you want to use migration based on Script XML, you should apply the migration script from a 390 version.

If you want to migrate the database from a distribution binary of 3.9.1 version, it is advisable to use the migration tool.

unfortunately, the proposed change ignores the approach

yamelsenih:
Here a change:

Fixes: #2249
current mp4

marcalwestf:
Confirming replication of error according to @yamelsenih 's  description in #2247 :
First case:
    * Get last DB seed from ADempiere
    * Run it with last source code

marcalwestf:
Now, turning on to the solution.

When running on an IDE the branch _hotfix/#2247-Add-Old-Compatibility-GetUUID_ of repository https://github.com/erpcya/adempiere/tree/hotfix/%232247-Add-Old-Compatibility-GetUUID with the latest commit 47c5d0933c65f37f6181af70ac224c5d8e033d5e , **I get the very same error**:
20190228_uuid_01


I did expect that with the modification on the above branch, the error would disappear.

So I wonder if I am wrong in my assumption, I tested wrongly or the error hasn't been eliminated.
Perhaps @yamelsenih  Yamel can clarify.
Thanks.

marcalwestf:
@e-Evolution Victor, I didn't understand your explanation of "surrogate keys" nor the later explanations.
I'd be grateful if you could elaborate so we can know if this proposed solution is wrong at all or a valid one.
Thanks!

Issue #2257, opened 2019-01-18
bug in script PostgreSQL ImportAdempiere.sh


homebeaver:
when performing `RUN_ImportAdempiere.sh` in 3.9.1 with postgres 10 I got

"create database failed with exit code 1"

but the database is created!

e-Evolution:
Hi @homebeaver 

What is the log that cause  code 1?

homebeaver:
After some investigation I found the bug in script PostgreSQL `ImportAdempiere.sh`

It happens when role `adempiere` exists

homebeaver:
a pull request follows

PR #2258, opened 2019-01-18
#2257 fix in utils/PostgreSQL/ImportAdempiere.sh

homebeaver:


homebeaver:
@e-Evolution 

one assignment fixes the problem :-)

Issue #2250, opened 2019-01-12
Add compatibility with Postgres 11 in migration tool


jdaison:
Now Select of Sequence in Postgres 11 have fewer columns in the result, this causes a problem with the migration tool 

![image](https://user-images.githubusercontent.com/6097506/51075861-52e5b200-165f-11e9-851e-fbe5c90c73d1.png)

The file to review is com.kkalice.adempiere.migrate.DBEngineInterface_Postgresql.java 
the method 

```java
public String sqlMetadata_sequenceDefinitions(String productVersion , String catalogName, String schemaName, String sequenceName) {
		if (productVersion.startsWith("10.")) { // add support for postgresql 10
			StringBuffer sql = new StringBuffer();
			sql.append("SELECT "
					+ "min_value AS MIN_VALUE, "
					+ "max_value AS MAX_VALUE, "
					+ "increment_by AS INCREMENT_BY, "
					+ "cycle AS IS_CYCLED ,"
					+ "cache_size AS CACHE_SIZE, "
					+ "last_value AS LAST_VALUE "
					+ "FROM ").append("pg_sequences WHERE sequencename='").append(sequenceName).append("' ");

			return sql.toString();
		} else { // support for postgresql 9
			return sqlMetadata_sequenceDefinitions(catalogName, schemaName, sequenceName);
		}
	}
```

PR #2259, opened 2019-01-18
#2250, Fixed error with supported version for postgreSQL, reference to

yamelsenih:
Add support to latest version of PostgreSQL

Fixes: #2250 
issue: https://github.com/adempiere/adempiere/issues/2250


marcalwestf:
here the result on Postgres 10:
20190227_migration_01

Issue #2266, opened 2019-01-21
Enlarge length for bank name - use system element BankName which is 120


homebeaver:
`c_bank.name` is limited to 60 chars. This is to short. See similar to c_location and email in #1340

In the 30 largest banks in germany, see [bankenverband.de](https://bankenverband.de/media/publikationen/16112017_Zahlen_und_Fakten_web.pdf) there are three to reach the limit:

```
Landeskreditbank Baden-Württemberg - Förderbank, Karlsruhe
WL Bank AG Westfälische Landschaft Bodenkreditbank, Münster
Bremer Landesbank Kreditanstalt Oldenburg - Girozentrale, Bremen
```

There is a system element `Bank Name` with length 120 - AD_Element_ID=52002 . But is is not used for `c_bank.name`. Why?

e-Evolution:
@homebeaver , I think that is ok , the next standard size is 256.

What do you think?

homebeaver:
For me 120 is enough. I'll provide a migration xml.

homebeaver:
@adaxa-steven  No - `c_bank` holds reference data for banks without bank accounts.
We import this data from the central bank - see [Beiträge here](http://www.adempiere.com/DE/Bank#Beitr.C3.A4ge) - there is an update quarterly a year from the Bundesbank

adaxa-steven:
120 sounds good

PR #2267, opened 2019-01-21
#2266 migration Enlarge_length_for_bank_name to 120

homebeaver:


e-Evolution:
@homebeaver , I revert this pull request because change a name for core table not is good for back compatibility , a lot the people can have build your views using name vs bank name , on the other side , when I generate model for C_Bank a the project not compile a lot the code use getName() method


image


image


image


image

image

image


So please only change of length size and not column name because break multiples code

PR #2333, opened 2019-02-12
Packout / Packin redefinition for handle UUID instead id of model

yamelsenih:
Hi everyone, this pull request make a enhancement over packin / packout tool. The main change are described follow:
- Change identifier for get ID from (Value / Name) to UUID from object
- Remove old implementation
- Add GenericPOHandler class for make packin and packout
- Add support to parent dependence from lookup
- Remove rounded code for distints components:
 - Window Access
 - Process Access
 - Browse Access
 - Task Access
 - Workflow Access
 - Org Access
 - Import Format
 - Print Paper
 - Print format item
 - Report View Column
 - View Column
 - View Definition
 - Browse Field
 - Tab
 - Field
 - Process Parameter
 - Column
 - Task
- Remove string calling for XML data and replace for interface calling
- Handle POFiller and Attributte data as loader and setter for packin and packout

I hope that this change can be tested and implemented for ADempiere In/Out

Best regard


SusanneCalderon:
Hi Yamel,
I tested for Table, Process and Workflow, works great.
Susanne

yamelsenih:
Hi @SusanneCalderon thanks for your test

PR #2341, opened 2019-02-14
Turns the Doc Status Indicator text into a link and shows all fields in grid.

### LINKED ISSUE #1243: Document Status Indicator window need to tick 'Display in Grid' = Y
all fields in the Document Status Indicator window need to be ticked as Display in Grid = Y

### LINKED ISSUE #2340: Document Status Indicator improvements.
The code that applied the style and color to the dashboard Document Tasks panel was removed in #1193.  The fields in the Document Status Indicator window can be hidden as they are no longer used.

Also, the Document Tasks are underlined but do not behave as links.  The I-bar cursor shows on hover rather than a finger.

mckayERP:
Fixes #2340.
Fixes #1243 
Changes the component used in the dashboard Document Task panel to a ToolBarButton which appears as a link.  This is better than a label with underlined text as the cursor turns to a "finger" rather than a text I-bar when hovering over the field.  

The change also sets all fields in the Document Status Indicator window to "Is Displayed in Grid"="Y".

### DIFF
--- a/migration/391lts-392lts/04440_2340_ImproveDocumentStatusIndicators.xml
+++ b/migration/391lts-392lts/04440_2340_ImproveDocumentStatusIndicators.xml
@@ -1,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<Migrations>
+  <Migration EntityType="D" Name="#2340 Document Status Indicator Improvements" ReleaseNo="3.9.2" SeqNo="4440">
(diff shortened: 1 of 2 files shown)


e-Evolution:
@mckayERP hi Mike  Can please set here a screenshot with result this way we can see the visual change?

mckayERP:
It looks the same.  Only the underlying component changed.

webui_dashboard_doctaskexample

Still need to re-enable the color and font.

Issue #1934, opened 2018-08-26
Generate docs from Application Dictionary


yamelsenih:
Is nice to have a documentation generator from application dictionary, i proposal create documentation with [reStructuredText](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) format, it allows define multiple language and is full used for all documents readers.

Is nice fi with a process from ADempiere a simple System user can generate docs

PR #2353, opened 2019-02-15
Add Documentation Generator from ADempiere

yamelsenih:
Hi guys, this pull request allows generate documentation from ADempiere Application Dictionary, the documentation can be generated for:

- [Markdown](https://en.wikipedia.org/wiki/Markdown)
- [reStructuredText](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html)

The result for it documentation can be some like here: [https://adempiere.github.io/](https://adempiere.github.io/) 
See this video for know what is step by step
current ogv

Fixex: #1934

### DIFF
--- a/base/src/org/spin/process/GenerateDocsFromMenu.java
+++ b/base/src/org/spin/process/GenerateDocsFromMenu.java
@@ -1,0 +1,427 @@
+/******************************************************************************
+ * Product: ADempiere ERP & CRM Smart Business Solution                       *
+ * Copyright (C) 2006-2017 ADempiere Foundation, All Rights Reserved.         *
+ * This program is free software, you can redistribute it and/or modify it    *
+ * under the terms version 2 of the GNU General Public License as published   *
+ * or (at your option) any later version.									  *
+ * by the Free Software Foundation. This program is distributed in the hope   *
+ * that it will be useful, but WITHOUT ANY WARRANTY, without even the implied *
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.           *
+ * See the GNU General Public License for more details.                       *
+ * You should have received a copy of the GNU General Public License along    *
+ * with this program, if not, write to the Free Software Foundation, Inc.,    *
+ * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.                     *
+ * For the text or an alternative of this public license, you may reach us    *
+ * or via info@adempiere.net or http://www.adempiere.net/license.html         *
+ *****************************************************************************/
+package org.spin.process;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.List;
+
+import org.adempiere.exceptions.AdempiereException;
+import org.adempiere.model.I_AD_Browse;
+import org.adempiere.model.MBrowse;
+import org.compiere.model.I_AD_Form;
+import org.compiere.model.I_AD_Process;
+import org.compiere.model.I_AD_Window;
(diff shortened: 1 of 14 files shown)


e-Evolution:
@yamelsenih Hi Yamel.

¿The documentation generate es multi-language, or we should add a new parameter for language, this way can generate the documentation for language that we need?

¿Exist a way using UUID to update application dictionary from git or GitHub, if somebody editor fixed some documentation, ie some contributor modify and extend the documentation or help, how we will update the application dictionary? , my suggestion is to use the UUID and language code to have some way the updated the application dictionary a once that a pull request to change the documentation is approved

yamelsenih:
Hi @e-Evolution my reply

>¿The documentation generate es multi-language, or we should add a new parameter for language, this way can generate the documentation for language that we need?

The Multi-language is variable from compile enviroment and format, example:
 - reStructuredText use a multi-language with distincts directory or branchs for habdle it.
 - Markdown is unsupported for mkdocs however exist a way for resolve it with forward and css

For now is unsupported multi-language for it first pull request but is a great start for generate documentation from ADempiere

>¿Exist a way using UUID to update application dictionary from git or GitHub, if somebody editor fixed some documentation, ie some contributor modify and extend the documentation or help, how we will update the application dictionary? , my suggestion is to use the UUID and language code to have some way the updated the application dictionary a once that a pull request to change the documentation is approved

My suggestion is that exist a ADempiere application server with real data and frienly window for update documentation.

This docs is only from application dictionary and can be versioned for each release.

I somebody want to add docs es better from ADempiere or using a different forlder like is loaded from gitbook from @mckayERP

mckayERP:
Hi Yamel,

Great work.  +1 to merge.

Some comments:

How are we going to add version info?  If I generate a version of the documentation for V3.9.2 say, all the links to this document should point at information that has the same version.  Would it be possible to include a version tag in the URL? Something like: `https://adempiere.github.io/v392/functional-guide/window/window-unit-of-measure.html.`  I can change the links in other docs en-mass with a global text change to point all the links to a new version when its ready.

Also would it be nice to have a way to capture screen shots?  Its a bit of a pain and I'm not sure it adds a lot of value  considering all the display logic.

yamelsenih:
Hi @mckayERP off course, github.io is only a compiler the important is the language and release version from docs code as tags or releases.

A example using a own compiler for mkDocs is it:
[http://docs.couchdb.org/en/stable/maintenance/performance.html#delayed-commits](http://docs.couchdb.org/en/stable/maintenance/performance.html#delayed-commits)
[https://docs.couchdb.org/en/1.6.1/maintenance/performance.html](https://docs.couchdb.org/en/1.6.1/maintenance/performance.html)

Both are the same documentation provider but distint version, the version control is a job of source + compiler. We are source and github.io is only a test for it but can be compiled with other owner compiler site

e-Evolution:
@yamelsenih I reviewed this contribution and is ok , you only need modify 3.9.2 the  04420_1934_Process_Generate_Docs  <Migration EntityType="D" Name="Generate docs from Application Dictionary #1934" ReleaseNo="3.9.1" SeqNo="4420">

And I think that the menu should move to General Rules under Generate Surrogate Keys

image

yamelsenih:
>@yamelsenih I reviewed this contribution and is ok , you only need modify 3.9.2 the 04420_1934_Process_Generate_Docs

Hahahaha this change was begining to develop for 3.9.0. fixed

>And I think that the menu should move to General Rules under Generate Surrogate Keys
Fixed
Screenshot_20190315_114931

Issue #2109, opened 2018-10-29
Add App Registration ADempiere


yamelsenih:
Is nice to have a table definition for handle ADempiere in/out operation, example:

I want to connet ADempiere with [Emitter Queue](https://emitter.io) or [RabbitMQ](https://www.rabbitmq.com/) and they are some common values:
  - Host
  - Port
  - Timeout
  - Keep Alive
  - Secret Token

yamelsenih:
Here a ER-Diagram
![app-registration](https://user-images.githubusercontent.com/2333092/47858953-2a1d8280-ddbb-11e8-9640-ae1b4d13fefc.png)

yamelsenih:
Here a Step by Step:
For ActiveMQ is created a topic named **printing**
![screenshot_20181101_105635](https://user-images.githubusercontent.com/2333092/47859558-9351c580-ddbc-11e8-8f9c-a3b77b543e3a.png)
![screenshot_20181101_104823](https://user-images.githubusercontent.com/2333092/47859210-cf385b00-ddbb-11e8-9a3d-62a472780e8b.png)
![screenshot_20181101_104843](https://user-images.githubusercontent.com/2333092/47859211-cf385b00-ddbb-11e8-8e0f-c9357654f9b3.png)
![screenshot_20181101_104850](https://user-images.githubusercontent.com/2333092/47859212-cf385b00-ddbb-11e8-9c65-a07a7c5276bb.png)
![screenshot_20181101_105018](https://user-images.githubusercontent.com/2333092/47859214-cf385b00-ddbb-11e8-8beb-54aa0025fb9b.png)
![screenshot_20181101_105045](https://user-images.githubusercontent.com/2333092/47859217-cfd0f180-ddbb-11e8-897d-4a0f5945ae46.png)
![screenshot_20181101_105058](https://user-images.githubusercontent.com/2333092/47859220-d0698800-ddbb-11e8-86e2-219053bd9132.png)
![screenshot_20181101_105108](https://user-images.githubusercontent.com/2333092/47859221-d0698800-ddbb-11e8-9c47-47e632e7d6cd.png)
Now note that exist a package inside **printing** topic
![screenshot_20181101_105842](https://user-images.githubusercontent.com/2333092/47859765-04917880-ddbd-11e8-807d-a4de86a5fdef.png)

Issue #672, opened 2016-11-16
Add abstract class for basic operation of POS


yamelsenih:
Currently exists a method named **printTicket** on the POS, it just send a print format to front end. But, one POS don't just need send a print format, it have other activities like:

- Show amount and messages on custom **Display**.
- Open drawer.
- Other.

The current methos is the follow:
<pre>
        /**
	 * 	Print Ticket
	 *
	 */
	public void printTicket() {
		if (!hasOrder())
			return;
		try
		{
			//if (p_pos.getAD_PrintLabel_ID() != 0)
			//		PrintLabel.printLabelTicket(order.getC_Order_ID(), p_pos.getAD_PrintLabel_ID());
			//print standard document
			ReportCtl.startDocumentPrint(0, getC_Order_ID(), false);
		}
		catch (Exception e) {
			throw new AdempierePOSException("PrintTicket - Error Printing Ticket");
		}
	}
</pre>

I Propose create a abstract class for handle it operation. The class to be defined like:

- Name: **POSTicketHandler**.
- Default methods:
  - **printTicket**().
  - **openDrawer**().
  - **showMessage**().
- The class by default can print a format like is currently.
- It can be extended for add more functionality and rewrite standard methods.
- It will be called from the POS Configuration.

yamelsenih:
The showMessage should be named showMsg(String)

PR #2376, opened 2019-02-21
Remove old approach for socket and local printing from POS

yamelsenih:
Hi everyone this pull request remove a old implementation of local printing from ADempiere ZK and POS.

I think that already exists other apporach for make it polssible.

Now from ADempiere 3.9.1 release is possible create a App Registration for make calling to external app.

A example of it can be it: [LocalPrinting](https://github.com/erpcya/LocalPrinting)

The new approach is with #2109 
The new approach for POS print ticket is: #672

Best Regard


marcalwestf:
Start doing the necessary steps to test.

PR #2384, opened 2019-02-23
Fix/fix for #197 delete cost detailsfor inactive cost elements

### LINKED ISSUE #197: GenerateCostDetails does not delete inactive cost types or methods

mckayERP:
Fixes #197.

Resubmitted.  Changes allow GenerateCostDetails to delete cost details related to inactive cost elements.

Cost Elements can be deleted and will delete the related cost dimensions and cost details.

Only one active default cost element per client is allowed and only one element of each cost type per client is allowed.

Inactive cost elements in GW client have been removed.


carlosaparadam:
hi @mckayERP i testing it, and the generate cost process only delete cost detail and not cost dimension.

In this example i inactive the cost elements Burden, Labor and Overhead.
cost_element

After run the Generate Cost Process all the cost elements inactive persist in the cost dimension.
cost_dimension

After run the Generate Cost Process all the cost elements inactive have delete on cost detail. 
cost_detail

Otherwise, i think so, at delete cost element delete cost detail and cost dimension in cascade be can dangerous.

Regards

mckayERP:
@carlosaparadam 
Hi  Carlos,

Thanks for testing.  I wasn't thinking of deleting the cost dimensions, just the cost details.  The problem occurred for me when I mistakenly created a second material cost method which duplicated the material costs.  I couldn't get rid of the duplicated cost details and these continued to inflate the cost of goods sold.

If you are happy the deletion of the cost details, that is fine with me.

Mike

carlosaparadam:
Hi @mckayERP, i don't see any problem on delete cost detail but not on delete cost element action, i think so that could be on after deactivate cost element and run Generate Cost Detail Process.

Waiting for you comments.

Regards

mckayERP:
@carlosaparadam 
H Carlos,
Thanks again for testing.

I'll add the deletion of the cost elements from the costs as well.  I think your idea will be better and less confusing for the users.

Mike

yamelsenih:
Hi @mckayERP any update?

mckayERP:
Hi Yamel, Sorry. Set this aside while working on other stuff.  I'll update it in the next two days.

yamelsenih:
> Hi Yamel, Sorry. Set this aside while working on other stuff. I'll update it in the next two days.

Thanks @mckayERP

mckayERP:
@yamelsenih @carlosaparadam 

Hi Yamel & Carlos,

Made the change.  Cost Dimensions are now deleted for inactive cost types and elements when the Generate Cost Details process is run.

Would it be a good idea to alert the user that the cost details need to be generated when a cost type or element is made inactive?  I don't suspect this happens often so it may not be worth the trouble.

Mike

carlosaparadam:
Hi, @mckayERP 

I testing it and it's correct, originally i have an physical inventory and the cost detail registered for each Cost Element and Type, see:
 
image

After inactive the cost element Freight, and run Generate Cost Detail process it's deleted the Cost Dimension and Cost Detail

image

Cost Dimension Result:
image

Cost Detail Result:
image

yamelsenih:
I will to approve this pull request

e-Evolution:
Hi Mike @mckayERP , Yamel @yamelsenih , Carlos @carlosaparadam .

Check the code quickly and I see that the logic of only allowing a record in the cost dimension by type of cost element has been added

https://github.com/adempiere/adempiere/pull/2384/files#diff-0000b87d8d78368608a5f798e4fe656fR240

Which is not correct, that is, the cost system is designed so that you can add multiple combinations in the combination of the cost dimension, even several elements on the same type of cost element.

Example of real life.

It is common that in the marketers import products from abroad and want to have a detailed landed cost type analysis, example I should have the possibility, in addition, to adding three elements of cost with the landed cost element cost type, to have the cost of freight, the costs of import and indirect by manage with same type of cost element but with cost element separately, in this case, Landed cost.

Thus the logic of having only one item of cost per type is not correct.

Please fix or revert this pull request

kind regards

yamelsenih:
Hi @e-Evolution you are right, this pull request must be revert.

I will to revert for fix it and merge

yamelsenih:
Hi @mckayERP please add changes described by @e-Evolution for merge.


Best regards

mckayERP:
@e-Evolution 
Hi Victor,

Was the previous logic correct?  Only one cost type of Material or Outside Processing was allowed, whether the cost element was active or not.  The other cost types could be included multiple times.

The current seed database violates this and has four Material entries, but three of the four are inactive.

Would it make sense to add the test for active cost elements to the logic there?

Mike

image

mckayERP:
@e-Evolution 

Also note the previous logic was testing a column which isn't available anymore - Costing Method.  I think it should have been testing Cost Element Type.

	protected boolean beforeSave (boolean newRecord)
	{
		//	Check Unique Costing Method
		if (
			(  COSTELEMENTTYPE_Material.equals(getCostElementType())
    //			|| COSTELEMENTTYPE_Resource.equals(getCostElementType())
    //			|| COSTELEMENTTYPE_BurdenMOverhead.equals(getCostElementType())
    //			|| COSTELEMENTTYPE_Overhead.equals(getCostElementType())
			|| COSTELEMENTTYPE_OutsideProcessing.equals(getCostElementType())
			)
			&& (newRecord || is_ValueChanged(COLUMNNAME_CostingMethod)))
               {
			String sql = "SELECT  COALESCE(MAX(M_CostElement_ID),0) FROM M_CostElement "
     				+ "WHERE AD_Client_ID=? AND CostingMethod=? AND CostElementType=?";
			int id = DB.getSQLValue(get_TrxName(), sql, getAD_Client_ID(), getCostingMethod() , getCostElementType());
			if (id > 0 && id != get_ID())
                       {
				log.saveError("AlreadyExists", Msg.getElement(getCtx(), "CostingMethod"));
				return false;
			}
		} ...

Issue #1662, opened 2018-04-18
Report Inventory Valuated To Date only works for costing level warehouse


SusanneCalderon:

SusanneCalderon:
The SQL in this process looks for qtys in warehouse in the table m_Costdetail. When you have costinglevel Client, the qtys in m_Costdetail are the overall qtys, not the qty for warehouse. 

_SELECT 1100864,tc.DateAcct,p.AD_Client_ID,
p.AD_Org_ID, tc.C_AcctSchema_ID ,tc.M_CostElement_ID,tc.M_CostType_ID, 
tc.M_Warehouse_ID,p.M_Product_ID,p.M_Product_Category_ID,tc.M_AttributeSetInstance_ID,
p.Group1,p.Group2,  
tc.qty + tc.cumulatedqty AS QtyOnHand, 
CASE WHEN tc.Qty < 0 OR (tc.qty = 0 AND tc.cumulatedqty < 0) THEN ((tc.costAmt + tc.costadjustment) * -1) + tc.CumulatedAmt 
ELSE ((tc.costAmt + tc.costadjustment) * 1) + tc.CumulatedAmt END  AS CostAmt, 
CASE WHEN tc.Qty < 0 OR (tc.qty = 0 AND tc.cumulatedqty < 0) THEN ((tc.costAmtLL + tc.costadjustmentLL) * -1)  + tc.CumulatedAmtLL 
ELSE ((tc.costAmtLL + tc.costadjustmentLL) * 1) + tc.CumulatedAmtLL END AS CostAmtLL 
FROM M_Product p  
INNER JOIN M_CostDetail tc ON (p.M_Product_ID=tc.M_Product_ID) 
WHERE tc.IsReversal='N' AND tc.DateAcct<= TO_DATE('2018-04-18','YYYY-MM-DD') 
AND p.M_Product_ID=1002192  --AND tc.C_AcctSchema_ID=?  
--AND tc.M_CostType_ID =?   AND tc.M_CostElement_ID=?  
AND tc.M_Warehouse_ID=1000020  AND tc.SeqNo = (SELECT MAX(SeqNo) 
										FROM M_CostDetail tc1 WHERE tc1.IsReversal='N' AND tc1.M_Product_ID=tc.M_Product_ID 
										AND tc1.M_Warehouse_ID = tc.M_Warehouse_ID 
										AND tc1.DateAcct<= TO_DATE('2018-04-18','YYYY-MM-DD') 
[...]

Issue #1502, opened 2017-11-27
Report Valuate Inventory effective date works only for costing level "Warehouse"


SusanneCalderon:
When the costing level _Client_ is selected, the values of beginningbalanceQty and beginningbalance in the table M_Costdetail are overall values(over all warehouses) and do not serve to calculate the qty/valuatedqty of a product at the chosen date for the report in a specific warehouse. 
In case of costing level _Client_ the column m_warehouse_ID only specifies the location of the corresponding transaction. For this costing level it is correct that  the beginninbbalanceQty/beginningbalance are  overall values, a new average cost is calculated on base of total quantities and total inventory value.

To get the correct qty and valuated stock for a warehouse at a specific date, the stock in the warehouse at this date must be calculated from the table m_transaction. This stock must be multiplied with currentcostprice + currentcostprice from the table m_costdetail, found in the lasted entry for the latest transaction before the selected date.

yamelsenih:
Any news?

PR #2394, opened 2019-02-24
adempiere  #1662 ValuationEffectiveDate only for costing level warehouse

SusanneCalderon:
fixes https://github.com/adempiere/adempiere/issues/1662

Fixes: #1502

### DIFF
--- a/base/src/org/eevolution/process/ValuationEffectiveDate.java
+++ b/base/src/org/eevolution/process/ValuationEffectiveDate.java
@@ -80,25 +80,25 @@ public class ValuationEffectiveDate extends ValuationEffectiveDateAbstract {
 		for (MAcctSchema acctSchema: acctSchemas) {
 			int no = 0;
 			String whereSeq = " (SELECT MAX(SeqNo) FROM M_CostDetail " + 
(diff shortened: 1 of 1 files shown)


base/src/org/eevolution/process/ValuationEffectiveDate.java
```diff
@@ -77,10 +77,38 @@ protected String doIt() throws Exception {
...
+			if (acctSchema.getCostingLevel().equals(MAcctSchema.COSTINGLEVEL_Client)) {
+				ArrayList<Object> params = new ArrayList<>();
+				params.add(getDateValue());
+				params.add(getDateValue());
+				params.add(getDateValue());
+				params.add(getAD_PInstance_ID());
+				StringBuffer update1 = new StringBuffer( "update t_Inventoryvalue iv ")
+						.append(" set costamt = (select coalesce(currentcostprice,0) from rv_m_transaction_costing cd where iv.m_Product_ID=cd.m_Product_ID AND cd.seqno= ")
+						.append(whereSeq);
+				update1.append(", costamtll = (select coalesce(currentcostpriceLL) from rv_m_transaction_costing cd where iv.m_Product_ID=cd.m_Product_ID AND cd.seqno= ")
```
yamelsenih: Can change it?
<pre>
set costamt = (select coalesce(currentcostprice,0) from rv_m_transaction_costing cd where iv.m_Product_ID=cd.m_Product_ID AND cd.seqno= 
</pre>

All SQL for ADempiere is better if you use SELECT with uppercase also the column like seqno is better if you can call it like SeqNo

PR #2403, opened 2019-02-27
Add conversion for bank statement helper method

yamelsenih:
This pull request add conversion from payment to bank statement, currently exists the follow method:

<pre>
MBankStatement.addPayment(paymentCash);
</pre>

**Bank Account Detail**:
 - Currency: USD

**Current Behavior:**
- Payment:
 - Payment Amount: 100
 - Payment Currency: EUR
- Bank statement:
 - Trx Amount: 100
 - Currency: USD (This is bad)

**After change:**
- Payment:
 - Payment Amount: 100
 - Payment Currency: EUR
- Bank statement:
 - Trx Amount: 124.90
 - Currency: USD

The previous method add automatically a payment to current bank statement, the problem is that if payment is created with other currency, it is not converted for bank statement.

Best regard

### DIFF
--- a/base/src/org/compiere/model/MBankStatement.java
+++ b/base/src/org/compiere/model/MBankStatement.java
@@ -76,14 +76,14 @@ public class MBankStatement extends X_C_BankStatement implements DocAction
 				.first();
 		if (bankStatement == null || bankStatement.get_ID() <= 0)
 		{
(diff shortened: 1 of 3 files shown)


e-Evolution:
@yamelsenih I send my pull request for this code https://github.com/erpcya/adempiere/pull/2 for use more functional programming, we should remove any use of null in our code

PR #2421, opened 2019-03-06
Add dependences for Smart Browse and process on packout

yamelsenih:
This pull request add dependence for Smart Browse and process when they are called from other entity.

Example:

 - **Process 1** have a Smart Browse named **Smart Browse 1**
 - **Smart Browse 1** have a process named **Process 2**

If I want to export **Process 1** the correct behavior is: 

- Export **Process 2**
- Export **Smart Browse 1**
- Export **Process 1**


marcalwestf:
Test Started: created as @yamelsenih suggested :
- **Process 1** having a Smart Browse named **Smart Browse 1**
- **Smart Browse 1** having a process named **Process 2**
- I added a window, and all components having translations.
See following images:
20190319_pack_01
20190319_pack_02
20190319_pack_03
20190319_pack_04
20190319_pack_05
20190319_pack_06
20190319_pack_07

marcalwestf:
..then I allowed for translations during two-pack
20190319_pack_08

marcalwestf:
..and exported the package.
The export file looks OK
20190319_pack_09

now, I only need to do the "pack-in" and see if everything is imported OK

marcalwestf:
...the packin defined:
20190319_pack_10

and executed.

**Result: the very same objects that were exported were imported!!!!**
Additionally all translations were also imported.

Interesting side-effect due to UUID usage: just changing the names of the objects **after** packout and executing the packin yields only setting the names to the original names. This makes sure that the objects are recognized even by packin.

PR #2433, opened 2019-03-15
Fixed error with UUID for Migration from model

yamelsenih:
Hi everybody, this pull request resolve a error with UUID for migration:

Currently and after 3.9.1 ADempiere version is added **UUID** column for all tables. For now this worked but when I try generate a migration from PO the **UUID** is not added to migration step.

Please review it.


marcalwestf:
Reproducing the error.
Creating an new window.
Result:
20190319_uuid_01
Error reproduced: UUID value is missing,

marcalwestf:
With changes: 
20190319_uuid_02

The value is now included in the migration.

PR #2442, opened 2019-03-26
Improve ActiveMQ support from app registration

yamelsenih:
Here a little change for a test case of App Registration with **ActiveMQ**
This is related to: https://github.com/adempiere/adempiere/pull/2376

Here a test with [ActiveMQ](https://hub.docker.com/r/rmohr/activemq)

Current ogv

### DIFF
--- a/base/src/org/spin/util/SendReportToQueue.java
+++ b/base/src/org/spin/util/SendReportToQueue.java
@@ -1,0 +1,118 @@
+/******************************************************************************
+ * Product: Adempiere ERP & CRM Smart Business Solution                       *
+ * This program is free software; you can redistribute it and/or modify it    *
+ * under the terms version 2 or later of the                                  *
+ * GNU General Public License as published                                    *
+ * by the Free Software Foundation. This program is distributed in the hope   *
+ * that it will be useful, but WITHOUT ANY WARRANTY; without even the implied *
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.           *
+ * See the GNU General Public License for more details.                       *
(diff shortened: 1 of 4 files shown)


marcalwestf:
hi, I tested from branch _feature/#Print-Local-Print-From-Server_.

I found 2 issues:
1.- The process _Registration Test_ , which is defined fot table _AD_AppRegistration_, has not been assigned to the role _SystemAdministrator_. That is why the process is not displayed for running.
See image:
20190401_ActiveMQ_01
Once the process _Registration Test_  has been added to the role, it appears.

2.- In class org.spin.util.support.mq.ActiveMQ , the file is hard-coded. If there is no file _/tmp/Factura_Nacional_1000158.pdf_, the test will fail
Perhaps is better defining the file in AD_SysConfig ?
20190401_ActiveMQ_03
20190401_ActiveMQ_04

marcalwestf:
...after manually granting access to the process and changing the file name (including path) to an existing file, I can see that ActiveMQ got a message to the queue _printing_, as it has been defined:
20190401_ActiveMQ_05
20190401_ActiveMQ_06

yamelsenih:
Hi @marcalwestf let me go to reply
>1.- The process Registration Test , which is defined fot table AD_AppRegistration, has not been assigned to the role SystemAdministrator. That is why the process is not displayed for running.
Just update role access

>2.- In class org.spin.util.support.mq.ActiveMQ , the file is hard-coded. If there is no file /tmp/Factura_Nacional_1000158.pdf, the test will fail

The class is just for test, is better if you define a new App registration parameter and get name from it or just aget name of report

Issue #1892, opened 2018-08-14
NPE when you press search in a search field


Raul-mz:
This only occurs in forms and in the reporting and process parameters. See images:
![image](https://user-images.githubusercontent.com/9552217/44102531-895868ae-9faf-11e8-8abd-0a2a133d6681.png)

![image](https://user-images.githubusercontent.com/9552217/44102561-9acd0478-9faf-11e8-8a59-528882f7d7a8.png)

PR #2447, opened 2019-03-28
#1892 Solved NPE in search fields

yamelsenih:
NPE for Bpartner search see #1892 

Fixed: #1892


nsarlabos:
The tests were done and it works ok!

Issue #2450, opened 2019-03-29
Forecast Run - problem generating first forcast run


mckayERP:
I had issues generating a forecast run for a product based on imported history data.

The forecast definition line draws data from the sales history with a join to the forecast run master table on product ID.  There was no way I could find to create the forecast run master entries.  The loop that looks at the queries only generates new master entries if entries already exist.

I modified the loop to create a new forecast run master if none were found and was able to generate the forecast.

PR #2451, opened 2019-03-29
#2450 Forecast Run has an issue generating the first run

mckayERP:
With imported sales history, I had difficulty in creating a forecast run.  The forecast run lines are based on a forecast run master.  The history is linked to the master table via the Product ID.  When creating the forecast lines, if there is no Master entry, there is no way to create a new one.

This pull request creates a new master entry if the query returns no results.
See https://github.com/adempiere/adempiere/issues/2450

Here is the original code in ForecastRunCreate with comments added to show the issue:

	private int generateForecastRunMaster(MPPForecastRun run,
			MPPForecastDefinition fd, MPPPeriod start, MPPPeriod end) {
		int count = 0;
		for (MPPForecastDefinitionLine fdl : fd.getLines(true)) {
                   //  Problem - note that the select statement requires an entry in PP_ForecastRunMaster
			final StringBuffer select = new StringBuffer(
					"SELECT DISTINCT sh.M_Product_ID , sh.M_Warehouse_ID FROM C_SalesHistory sh LEFT JOIN  PP_ForecastRunMaster m ON (m.M_Product_ID=sh.M_Product_ID AND ");
			select.append(MPPForecastRunMaster.COLUMNNAME_PP_ForecastRun_ID);
			select.append("=").append(run.get_ID()).append(" ) WHERE  ");
			select.append(fdl.getSQlWhere(run.getM_WarehouseSource_ID(), "sh"));
			select.append(MSalesHistory.COLUMNNAME_DateInvoiced);
			select.append(" BETWEEN ");
			select.append(DB.TO_DATE(start.getStartDate()));
			select.append(" AND ");
			select.append(DB.TO_DATE(end.getEndDate()));

			PreparedStatement pstmt = null;
			ResultSet rs = null;
			try {
				pstmt = DB.prepareStatement(select.toString(), get_TrxName());
				rs = pstmt.executeQuery();
                           //  Problem - if PP_ForecastRunMaster table is empty, the record set will not have
                           //  any results and the while loop will be skipped 
				while (rs.next()) {
					count++;
					int M_Product_ID = rs.getInt("M_Product_ID");
					MPPForecastRunMaster master = MPPForecastRunMaster
							.getByProduct(getCtx(), run.getPP_ForecastRun_ID(),
									M_Product_ID, get_TrxName());
					if (master != null)
						continue;

                                    //  Problem - in order to get here, the result set had to have returned 
                                    //  something
					master = new MPPForecastRunMaster(getCtx(), 0,
							get_TrxName());

					master.setAD_Org_ID(run.getAD_Org_ID());
					master.setPP_ForecastRun_ID(run.get_ID());
					master.setPP_ForecastDefinitionLine_ID(fdl
							.getPP_ForecastDefinitionLine_ID());
					master.setM_Product_ID(M_Product_ID);
					master.setM_Warehouse_ID(run.getM_Warehouse_ID());
					master.setFactorAlpha(fdl.getFactorAlpha());
					master.setFactorGamma(fdl.getFactorGamma());
					master.setFactorBeta(fdl.getFactorBeta());
					master.setFactorMultiplier(fdl.getFactorMultiplier());
					master.setFactorScale(fdl.getFactorScale());
					master.setFactorUser(fdl.getFactorUser());
					master.saveEx();
				}
			} catch (Exception e) {
				log.log(Level.SEVERE, select.toString(), e);
				getProcessInfo().addLog(getProcessInfo().getAD_PInstance_ID(),
						null, null, e.getLocalizedMessage());
			} finally {
				DB.close(rs, pstmt);
				rs = null;
				pstmt = null;
			}
		}
            //  With no results, count is zero and no master record was generated.
		return count;
	}

### DIFF
--- a/org.eevolution.manufacturing/src/main/java/base/org/eevolution/process/ForecastRunCreate.java
+++ b/org.eevolution.manufacturing/src/main/java/base/org/eevolution/process/ForecastRunCreate.java
@@ -162,6 +162,8 @@ public class ForecastRunCreate extends SvrProcess {
 			DataSet series = new DataSet();
 			series.setPeriods(m_run.getPeriodHistory());
 			MProduct product = MProduct.get(getCtx(), master.getM_Product_ID());
+			if (product == null)
+				continue;
 
 			List<MPPForecastRunDetail> details = MPPForecastRunMaster
 					.getDetails(getCtx(), master.get_ID(), get_TrxName());
@@ -250,7 +252,8 @@ public class ForecastRunCreate extends SvrProcess {
 		int count = 0;
 		for (MPPForecastDefinitionLine fdl : fd.getLines(true)) {
 			final StringBuffer select = new StringBuffer(
-					"SELECT DISTINCT sh.M_Product_ID , sh.M_Warehouse_ID FROM C_SalesHistory sh LEFT JOIN  PP_ForecastRunMaster m ON (m.M_Product_ID=sh.M_Product_ID AND ");
+					"SELECT DISTINCT sh.M_Product_ID , sh.M_Warehouse_ID FROM C_SalesHistory sh "
+				  + "LEFT JOIN  PP_ForecastRunMaster m ON (m.M_Product_ID=sh.M_Product_ID AND ");
 			select.append(MPPForecastRunMaster.COLUMNNAME_PP_ForecastRun_ID);
 			select.append("=").append(run.get_ID()).append(" ) WHERE  ");
 			select.append(fdl.getSQlWhere(run.getM_WarehouseSource_ID(), "sh"));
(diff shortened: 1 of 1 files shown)


mckayERP:
Fixes #2450 .

e-Evolution:
Hi Mike!

A question if there is no information in the sales history which is the reason for creating a sales forecast without a history?

Checking the code it seems right to validate that if there is no product on the bill does not generate any forecast, so that is correct in the correction of validate product id

In the last commits of the issue of forecast I noticed that the invoices generated from a sales order with immediate billing rule to not have linked the delivery line, did not consider the warehouse for these invoices which generated a discrepancy in the given those products if they were delivered but with another billing rule different from the one after delivery

mckayERP:
Hi Victor,

I imported sales history data.   Its not the sales history but the table PP_ForecastRunMaster.   As the comments in the code above show, the initial query only works if there is an entry in PP_FoecastRunMaster.   If PP_ForecastRunMaster is empty, the result set will not return any entries and the while(rs.next()) loop will be skipped.  New PP_ForecastRunMaster records can only be added **within** the while loop.

Is there another way to generate the PP_ForecastRunMaster record?

To test, create a new client, import a small set of data and try to generate a forecast.

PR #2452, opened 2019-03-29
#2434 Tax.java

### LINKED ISSUE #2434: Tax.java problem determining parent tax
The test for a parent tax uses the value of the parent ID rather than the summary checkbox.  The parent ID can be non-zero with the summary checkbox selected.  In the tax rate form, the parent tax rate is hidden if the summary is selected so there is no way for the user to know.

===========> zkoss.handleError:1130: >>org.adempiere.exceptions.TaxNotFoundException: Could not find tax - Tax Category:Sales Tax, Sales Transaction:Yes, Shipment (03/15/2019, XXX Road, XXX, Ontario XXXX -> XXX Rd., XXX, Ontario ), Invoice (03/15/2019, XXX Ontario -> XXX, Ontario )
>>	at org.compiere.model.Tax.get(Tax.java:608)

mckayERP:
Fixes #2434

When determining if a tax is a child tax or summary tax, Tax.java uses the test  ```tax.getParent_Tax_ID()!=0```  instead of testing ```!tax.isSummary()```.  In the gui, the parent tax field is hidden if the Is Summary checkbox is selected so there is no way for the user to know that Parent_Tax_ID is non-zero.  

See https://github.com/adempiere/adempiere/issues/2434

To test, create a parent and child tax rate pair.  On the parent tax, deselect the Is Summary checkbox and set the parent tax rate to some other tax rate.  Then reselect the Is Summary check box.  Note the parent tax field is hidden.

Set up an order that would use that tax.  An error occurs as no tax is found.

### DIFF
--- a/base/src/org/compiere/model/Tax.java
+++ b/base/src/org/compiere/model/Tax.java
@@ -39,6 +39,8 @@ import org.compiere.util.DB;
  * 
  * @author Teo Sarca, www.arhipac.ro
  * 			<li>FR [ 2758097 ] Implement TaxNotFoundException
+ * @author Michael McKay, mckayERP@gmail.com
+ * 			<li><a href="https://github.com/adempiere/adempiere/issues/2434">#2434</a> Wrong test for parent/summary tax.
  */
 public class Tax
 {
@@ -538,7 +540,7 @@ public class Tax
 			//
 			if (tax.getC_TaxCategory_ID() != C_TaxCategory_ID
 				|| !tax.isActive() 
-				|| tax.getParent_Tax_ID() != 0)	//	user parent tax
+				|| !tax.isSummary())	//	#2434 - only looking for summary taxes 
 				continue;
 			if (IsSOTrx && MTax.SOPOTYPE_PurchaseTax.equals(tax.getSOPOType()))
 				continue;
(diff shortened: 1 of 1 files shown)


base/src/org/compiere/model/Tax.java
```diff
@@ -538,7 +540,7 @@ protected static int get (Properties ctx,
 			//
 			if (tax.getC_TaxCategory_ID() != C_TaxCategory_ID
 				|| !tax.isActive() 
-				|| tax.getParent_Tax_ID() != 0)	//	user parent tax
+				|| !tax.isSummary())	//	#2434 - only looking for summary taxes
```
yamelsenih: - The old code continue when a tax have a parent tax
- The new code continue when a tax is not a summary

I think that some is wrong here

yamelsenih: Should be tax.isSummary() instead !tax.isSummary()

mckayERP:
It is confusing.

The condition is eliminating child taxes.  A child has a parent tax (ID !=
0) but so can a summary tax.  The better test is to eliminate !IsSummary().

Mike

yamelsenih:
hahaha you are right

best regard

yamelsenih:
Hi @mckayERP I test ti from a customer with data and exist a error:

The problem is that if a Tax is not flagged as **IsSummary** is not finded.

I will to revert it

I apologize for make a bad test

Issue #1983, opened 2018-09-09
Add iban to c_bp_bankaccount and bp window


homebeaver:
with iban validator for those countries using IBAN-standard. Iban is the current standard, see https://en.m.wikipedia.org/wiki/International_Bank_Account_Number . It replaces the old RoutingNo+AccountNo and can be validated.

adaxa-steven:
Hi Homebeaver 
The map on your Wikipedia link shows most parts of the world not using IBAN..  Not used in AU for instance. what are. You proposing?

marekmosiewicz:
I propose add field IsAcountNoIBAN. In case if it is true diable routingNo and have AccountNo able to be IBAN number. Alternativle there can be callout to check if it is IBAN and set this field automatically

homebeaver:
tkx @marekmosiewicz 
@adaxa-steven - my suggestion was not delete the "old" RoutingNo+AccountNo method. It can still coexist.

yamelsenih:
Hi @all I think that if IBAN code is not put in BP Bank Account is not necessary a validation else if IBAN is put then can be validated

PR #2455, opened 2019-03-31
#1983 : add iban col to c_bp_bankaccount

homebeaver:
Hi all,
In #1983 IBAN was suggested as alternative to RoutingNo+AccountNo.
IBAN (ISO 13616) is now used by 76 coutries - see https://www.swift.com/sites/default/files/resources/iban_registry.pdf 
In europe the use of IBAN is mandatory. [Countries where the use of IBAN is recommended.](https://www.dnb.no/en/business/transaction-banking/international-payments/example-iban.html)

grafik


homebeaver:
Hi @yamelsenih please review this PR
I wonder why the ad_column_id is 1000000 - in my opinion for system objects (AD dictionary) the id should be below 1000000!? - for example 92263 would be OK.

grafik

How can I get the right id?
Thanks in advance.

homebeaver:
Hi all, 
I add `commons-validator-1.6.jar` - with class IBANValidator for 74 countries. Bu I do not use it now,
because I miss validator for two countries VA and SV(El Salvador) which are defined in IBAN_Registry.
@marcalwestf - the validator for SV-IBANs is coded in the apache repo and [planned for 1.7.jar, see here](https://github.com/apache/commons-validator/commit/342b4b2ebeae12c774a3f325411fe19eb6df3d07), but I do not know when the jar will be released.

yamelsenih:
Hi @homebeaver have you a test case for validation?

Best regard

homebeaver:
@yamelsenih 
Yes - groovy see: How to use and extend the validator in explaind [here](http://wiki.adempiere.net/DE/Gesch%C3%A4ftspartner_Stammdaten#IBAN_Validator)
But ist not implemented in this PR.

yamelsenih:
Hi @homebeaver is nioce this change but I need a complete test case for include to source, Can you write a before save for **MBPBankAccount** for validate it? 

Other question:
This change is a generic? I think that can be added a flag for Client for validate IBAN

homebeaver:
Hi @yamelsenih - we use only groovy for Processes and Validators. It works fine: here an ui-example:

grafik

And a short instruction how to define it:

- define a Callout for the Column:
grafik
- and a rule with the groovy code:
grafik

```groovy
package com.klst.xxxxxx

import groovy.lang.Binding
import groovy.lang.Script

import org.apache.commons.validator.routines.IBANValidator

import org.compiere.model.GridTab
import org.compiere.util.CLogger

class CalloutIbanValidator extends Script {

	private static CLogger	log	= CLogger.getCLogger(CalloutIbanValidator.getName())
	String CLASSNAME = this.getClass().getName()
	
	public CalloutIbanValidator() {
	}

	public CalloutIbanValidator(Binding binding) {
		super(binding);
	}

	@Override
	public Object run() {
		log.config("${CLASSNAME}:run")
		IBANValidator validator = IBANValidator.getInstance()
		if(A_Value==null) { // new Object 
			log.config("${CLASSNAME}:run '${A_Value}' is null")
			return ""
		}
		if(validator.isValid(A_Value)) {
			log.config("${CLASSNAME}:run '${A_Value}' is valid")
			return ""
		}
		def iban_Value = A_Value
		println "${CLASSNAME}:run '${iban_Value}' is not valid"
		((GridTab)A_Tab).dataIgnore() // Ignore data changes
//		A_Tab.dataRefreshAll() // GridTable: Refresh all Rows - ignore changes
		return "'"+iban_Value + "' is not valid IBAN";
	}
	
	static main(args) {
	//		Script script = new CalloutIbanValidator()
	//		script.run()
		}
	
}
```

> can be added a flag for Client for validate IBAN

In my opinion there is no need for a flag. It is validating every time when changed.

yamelsenih:
Hi @homebeaver thanks for your response.
I think that a callout shoud not be used for a validation, is better if you use a trigger **beforeSave** for  **MBPBankAccount**

Can you add the code as java code?
Here a example:
Screenshot_20190424_091632

For BP Bank Account
<pre>
	/**
	 * 	Before Save
	 *	@param newRecord new
	 *	@return true
	 */
	protected boolean beforeSave(boolean newRecord) 
	{
		//	maintain routing on bank level
		if (isACH() && getBank() != null)
			setRoutingNo(null);
		// validate IBAN
		if(!Util.isEmpty(getIBAN())) {
			IBANValidator validator = IBANValidator.getInstance();
			if(!validator.isValid(getIBAN().trim())) {
				throw new AdempiereException("@ValidationError@ (@Invalid@ @IBAN@)");
			}
		}
		return true;
	}	//	beforeSave
</pre>

For Bank Account C_BankAccount

<pre>
	@Override
	protected boolean beforeSave(boolean newRecord) {
		// validate IBAN
		if(newRecord) {
			if(!Util.isEmpty(getIBAN())) {
				IBANValidator validator = IBANValidator.getInstance();
				if(!validator.isValid(getIBAN().trim())) {
					throw new AdempiereException("@ValidationError@ (@Invalid@ @IBAN@)");
				}
			}
		}
		//	
		return true;
	}
</pre>

Both import it:
<pre>
import org.adempiere.exceptions.AdempiereException;
import org.apache.commons.validator.routines.IBANValidator;
</pre>

Also you must generate **I_** and **X_** class for **C_BP_BankAccount**

homebeaver:
Hi @yamelsenih - this is the commit you requested for both entities.
I don't know why Travis CI - Build fails :-(

Regards EUGen

yamelsenih:
Thanks @homebeaver 

Can you create a new branch with all changes for make a new pull request with only a commit?

Is more atomic
Screenshot_20190425_102731


Screenshot_20190425_102604

Best regard

Issue #2427, opened 2019-03-11
Error When Creating Translations


marcalwestf:
1.- When an additional language is being added, it has to be set as _System Language_.
Let's take Italian:
![20190309_Creating_Language_00](https://user-images.githubusercontent.com/1789408/54100859-3a0b2b80-4386-11e9-8642-ea01ad987286.png)

2.- Then the process _Language Maintenance_ must be called
![20190309_Creating_Language_01](https://user-images.githubusercontent.com/1789408/54100879-57d89080-4386-11e9-9573-632a705c871e.png)

3.- An Error is thrown
![20190309_Creating_Language_02](https://user-images.githubusercontent.com/1789408/54100892-67f07000-4386-11e9-9e1f-a86576546ad0.png)

4.- In the IDE
![20190309_Creating_Language_03](https://user-images.githubusercontent.com/1789408/54100904-78084f80-4386-11e9-85e8-8efd853af533.png)
It is _MLanguage.addTable()_.

5.- The SQL confirms the findings:
![20190309_Creating_Language_04](https://user-images.githubusercontent.com/1789408/54101048-1c8a9180-4387-11e9-8ec6-4624bb392260.png)
The table _ad_browse_ has the definition of column Name as follows
`name character varying(255) NOT NULL,`
whereas table  _ad_browse_trl_ has the definition of column Name as follows
`name character varying(60) NOT NULL,`
This discrepancy causes the error.

5.- When incrementing the size of _Name_ in _ad_browse_trl_ to 255:
![20190309_Creating_Language_05](https://user-images.githubusercontent.com/1789408/54101062-2a401700-4387-11e9-9c2d-c8995913f671.png)

6.- The process  _Language Maintenance_ is executed without error
![20190309_Creating_Language_06](https://user-images.githubusercontent.com/1789408/54101106-6bd0c200-4387-11e9-9eff-fa8f21b5bbf8.png)

yamelsenih:
Hi @marcalwestf can you make a pull request with SQL improvement?

Best regard

PR #2470, opened 2019-04-09
#2427 Error When Creating Translations

marcalwestf:
Issue: https://github.com/adempiere/adempiere/issues/2427

Field AD_Browse_Trl increased both in AD and DB to 255.


e-Evolution:
fixes #2427

Issue #2472, opened 2019-04-09
When you generate the invoice history for the forecast, the sales rep is not used


e-Evolution:

yamelsenih:
Please @e-Evolution is nice if you can create or add some description for issue.

Best regard

PR #2476, opened 2019-04-11
#2472 When you generate the invoice history for the forecast, the sal…

e-Evolution:
…es rep is not used https://github.com/adempiere/adempiere/issues/2472


e-Evolution:
Related #2471

PR #2488, opened 2019-04-18
Add support to tables to exclude from packout

yamelsenih:
Add po model to export parent from packout instead id from context


yamelsenih:
Replaced by: #2498

Issue #2484, opened 2019-04-16
Add relation between Request Update and Invoice Line


e-Evolution:
Currently when an invoice is generated from a request there is no dependency on the line of the invoice generated with the update of the request that should be mark as processed

Issue #2483, opened 2019-04-16
Add relation between Request Update and Request Action


e-Evolution:
It is currently not possible to link a resource update with the request history, which makes it difficult to know how an invoice of a request is linked to the result of the request.

PR #2489, opened 2019-04-22
#2484  Add relation between Request Update and Invoice Line

e-Evolution:
…ithub.com/adempiere/adempiere/issues/2484

#2483 Add relation between Request Update and Request Action https://github.com/adempiere/adempiere/issues/2483


e-Evolution:
replace for #2507

PR #2492, opened 2019-04-23
BOM Drop form improvments.  

### LINKED ISSUE #2435: Improve BOM Drop functionality
I had to give a demo a few days ago that involved the BOM Drop. The example was basically a computer with variants and options for all the components.  It took some software changes to make it work well in the webui.  I think it is a very useful tool for product configuration and sales/purchasing when buying a collection of stuff with options and variants.

I'd like to propose some improvements though:

- Add the BOM Drop button to relevant documents (Orders, Invoices, Projects, etc...)
- Make the form like a smart browse rather than a custom form.
- Allow the options, variants and components to be grouped and collapsed in the form.  (In the example I was using there were six "features" and almost forty products. Being able to collapse the feature groups was very useful.)
- Make the features a list rather than a text field.
- Add compatibility checks to products so that the selected products are compatible.

Let me know if this would be useful to others.

mckayERP:
Fixes #2435 

This pull request fixes and improves the BOM Drop functionality.  The swing and ZK BOM Drop forms are equivalent and can be accessed from the menu or from an Order, Invoice or Project as a button/process.  A new radio button editor was added to support the optional/variant features.  A Plant Starter Kit product was added to Garden World, along with some subproducts to provide a test of the optional and variant features.

Note that a Role Access Update is required to access the BOM Drop process from the Order, Invoice or Project window/tab.

Draft documentation:  https://mckayerp.gitbook.io/adempiere/v/contrib_2435/introduction/products-and-material-management/bom-drop

See https://github.com/adempiere/adempiere/issues/2435

### DIFF
--- a/base/src/org/adempiere/controller/form/BOMDropController.java
+++ b/base/src/org/adempiere/controller/form/BOMDropController.java
@@ -1,0 +1,1249 @@
+/******************************************************************************
+ * Product: ADempiere ERP & CRM Smart Business Solution                       *
+ * Copyright (C) 2006-2019 ADempiere Foundation, All Rights Reserved.         *
+ * This program is free software, you can redistribute it and/or modify it    *
+ * under the terms version 2 of the GNU General Public License as published   *
+ * or (at your option) any later version.                                     *
+ * by the Free Software Foundation. This program is distributed in the hope   *
+ * that it will be useful, but WITHOUT ANY WARRANTY, without even the implied *
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.           *
+ * See the GNU General Public License for more details.                       *
+ * You should have received a copy of the GNU General Public License along    *
+ * with this program, if not, write to the Free Software Foundation, Inc.,    *
+ * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.                     *
+ * For the text or an alternative of this public license, you may reach us    *
+ * or via info@adempiere.net or http://www.adempiere.net/license.html         *
+ *****************************************************************************/
+
+package org.adempiere.controller.form;
+
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyVetoException;
+import java.beans.VetoableChangeListener;
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Properties;
+import java.util.logging.Level;
+
+import org.adempiere.exceptions.AdempiereException;
+import org.adempiere.exceptions.ValueChangeEvent;
+import org.adempiere.exceptions.ValueChangeListener;
+import org.compiere.model.MColumn;
+import org.compiere.model.MInvoice;
+import org.compiere.model.MInvoiceLine;
+import org.compiere.model.MLookup;
+import org.compiere.model.MLookupFactory;
+import org.compiere.model.MOrder;
+import org.compiere.model.MOrderLine;
+import org.compiere.model.MProduct;
+import org.compiere.model.MProject;
+import org.compiere.model.MProjectLine;
+import org.compiere.model.MTable;
+import org.compiere.model.MUOM;
+import org.compiere.model.MUOMConversion;
+import org.compiere.model.MValRule;
+import org.compiere.model.PO;
+import org.compiere.process.DocAction;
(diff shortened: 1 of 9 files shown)


e-Evolution:
Dear @Mike
Review the public documentation and it is an excellent initiative so that this documentation can arrive at ADempiere some time.

Some point that I do not see documented and that are currently modeled within the product domain model and lists of materials.

the concept of Drop BOM was replaced by the configured product which is identified by the material lists with the BOM Type Product Configure, so no is required set the product as not stocked it is simply validated to have a list of materials with a type of product configured.

image


On the other hand, in this domain model, there are two types of components for the bill of materials:

Variant: It allows define the quantity required for the configured product, establishing a predetermined quantity that can be changed or even not considered to manufacture the configured product.

image


Option: It allows you to choose a component of multiples options for a characteristic, but without modifying the units of the component for the finished product.

image


Feature: allows to group different components, whether variants or options, the characteristics guide the variants and options to build a configured product.

image


In the current functionality once finished the configuration of your product the system allows you to create a BOM with the configuration as an Alternate BOM.

image

Whether a product is included in detail or not in sales order is defined by the parameters "Print detail records on invoice"
"Print detail records on pick list" of the product, given that the product is created using Manufacturing Order.

I consider it convenient, before integrating this functionality, to try to make the already existing concepts compatible so that the new functionality is routed appropriately.

e-Evolution:
Some references:

http://wiki.adempiere.net/Sponsored_Development:_Libero_Product_Configuration_BOM
https://www.youtube.com/watch?v=yL3HRtEcRpU

mckayERP:
@e-Evolution 

Hi Victor,

Thanks for the feedback.

I didn't include the product configuration BOM as I thought it was sufficiently different task but the two could be merged.

The two tasks would be:

1. BOM Drop Form: Create document lines from a set of Product BOM lines following the option/variant model.  This is similar to ordering a set of compatible parts, as in building a computer with accessories.  See https://www.logicsupply.com/nuvo-6000/ as an example of an on-line store that follows this model.  There is no manufacturing involved.  Each line of the BOM would have to be on the price list.  The document total would be the sum of the selected BOM lines.
2. Product Configuration Form: Create a single document line that represents a "new" product which contains the lines selected in the Product Configuration form.  The new product would have a BOM created with only the selected components where options and variants would be changed to components.  This new product could then trigger demand and manufacturing orders so it could be built as configured.

I can see from the code in VProductConfigurationBOM that the implementation is not complete.  The code does not create a new instance of the configured product and doesn't search for matching instances of the BOM except for Orders.  At this time, I think the BOM Drop could be included as it is complete and does not exclude the 2nd approach.

I'm happy to work on this 2nd task but I have a few questions:
- Would this new product be added to the price list at the same price as the master product? If not, how would the price be determined?  Does the option/variant BOM Line need to include the price delta? 
- How would the new product be named/identified?
- How would it be described on the order line?
- How would the new product be differentiated as a configured product? Does it need a field for the master product it was created from?
- Should the new product be hidden on the price lists?
- Should it be excluded from the dynamic validation on manually entered document lines?  If so, the master product would have to be configured every time it is used.  If not, considering there could be hundreds of possible configurations, how would the user identify the configuration as the right one form the document line?

I'd be happy to hear what you think.

Mike

e-Evolution:
Hi Mike @mckayERP 

The implementation was done by Tim I suppose he did only what it required

But what matters is how we do it now.

I have reviewed the code of the sales order in the explodeBOM () method, I see here what you were referring to taking the components of the BOM and bringing them to the sales order I think the approach is correct. For this case

but also exists another case:

Second case: If there is a need to create an assembly order, it would work more like a Make-To-Kit, that is, we should have a way to indicate that that configuration generates a new list of materials Make to Kit or Make to Order, creating a new product with the specific configuration allowing you to create a manufacturing or assembly order to assemble the computer.

On the other hand, in the future, attribute instances could be used to generate more complex configurations that include sets of attributes, so the product that has the BOM as a configured type would only be a base product and base material list to create an infinite number of combination and attributes

each configuration would be saved as a new product, with its new specific instance and material list (make to kit or make to order).

image


the new product could be added to the price list by adding up the prices of the selected components.

In this way, you would still have the possibility to print the detail in the shipping order or invoice using the new product and the flags are Invoice Print Details or Is PickList PrintDetails how it works until now.

image

kind regards
Victor

Issue #2510, opened 2019-04-29
Change the method for printing standard for a batch mode without UI


e-Evolution:
Executes the printing of an order or invoice form background without UI, fails because it tries to show a UI messaje.

PR #2511, opened 2019-04-29
 #2510 Change the method for printing standard for a batch mode witho…

e-Evolution:
…ut UI https://github.com/adempiere/adempiere/issues/2510


yamelsenih:
Fixes: #2510

PR #2536, opened 2019-05-07
#2266 Enlarge_length_for_bank_name to 120

homebeaver:
Hi @e-Evolution ,
this is my 2nd try to launch a PR for #2266 

You have reverted the last one. I tested this with 3.9.2
grafik

regards EUGen


e-Evolution:
hi @homebeaver 

the problem no is script , the problem is that you are change a column name from a domain model that exist , this no is a good practice for back compatibility , given that a lot the customization that using name can fail, you can check the impact when generate the domain clases model for X_C_Bank and I_C_Bank the build fail because exist multiples code lines that use getName.

Please check my comments that set here https://github.com/adempiere/adempiere/pull/2267

So if you want change the length for Name column name is ok only validate that not exist views dependences , because PostgreSQL need drop and recreate view when changes some length 

kind regards
Victor

PR #2543, opened 2019-05-08
#2541 Allow to delete draft invoices generated from a request update.…

### LINKED ISSUE #2541: Allow to delete draft invoices generated from a request update.
As of version 3.9.2 the request updates have an integrity reference with the invoice lines, so it does not allow to delete the invoice until the dependency with requests is removed

e-Evolution:
…  https://github.com/adempiere/adempiere/issues/2541

resolved #2541

### DIFF
--- a/org.adempiere.request/src/main/java/org/compiere/model/RequestModelValidator.java
+++ b/org.adempiere.request/src/main/java/org/compiere/model/RequestModelValidator.java
@@ -18,6 +18,7 @@
 package org.compiere.model;
 
 import org.compiere.util.CLogger;
+import org.compiere.util.DB;
 import org.compiere.util.Env;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


org.adempiere.request/src/main/java/org/compiere/model/RequestModelValidator.java
```diff
@@ -90,6 +93,14 @@ public String login(int AD_Org_ID, int AD_Role_ID, int AD_User_ID) {
     @Override
     public String modelChange(PO entity, int type) throws Exception {
         Boolean isSOTrx = "Y".equals(Env.getContext(entity.getCtx(),"IsSOTrx"));
+        //#2541 Allow to delete draft or in process invoices generated from a request update.
```
yamelsenih: Hi, it can be replaced by model class wikth a call to Query object instead SQL native

e-Evolution: Hello Yamel

Originally I thought the same as you, but unfortunately the PO to update only one record, requires loading the entire instance in memory, a result set is executed, and then an update only to remove a reference by value null. is the reason why I do a pure update

yamelsenih: ok, is very lazy PO for it

Issue #2555, opened 2019-05-11
Printing Orders with Jasper Reports do not work anymore


marcalwestf:
This issue is a collateral effect of #2510 .

When defining a _Print Format_ for an Order, referencing to a _JasperReport_ process, there is an exception.
Here an example of a Print Format for an Order:
![20190511_PrinFormate_07](https://user-images.githubusercontent.com/1789408/57574622-a4c5fb00-73f9-11e9-95f0-b91f82e00f77.png)

Here the exception:
![20190511_JR_02](https://user-images.githubusercontent.com/1789408/57574608-73e5c600-73f9-11e9-952a-b9784ad00fb4.png)

The problem is that the parameter _processInfo_ of the method  _ServerReportCtl.runJasperProcess()_, which is in charge of executing the Jasper Porcess, has no _transactionName_, which leads to the exception:
![20190511_JR_01](https://user-images.githubusercontent.com/1789408/57574669-844a7080-73fa-11e9-9e24-d3c8c347f03d.png)

To solve the problem, the processInfo should be created with a proper transactionName.

Due to the fact that this is a core functionality, and the present error is severe, I suggest to quickly solve the problem by avoiding the exception, and leaving the proper correction for later, after duly testing.

Example of quick solution:
![20190511_JR_06](https://user-images.githubusercontent.com/1789408/57574713-47cb4480-73fb-11e9-91ab-5792eebb178f.png)

PR #2556, opened 2019-05-11
#2555 Printing Orders with Jasper Reports do not work anymore https:/…

e-Evolution:
…/github.com/adempiere/adempiere/issues/2555


e-Evolution:
resolve #2555

PR #2559, opened 2019-05-13
Fixed error missing translation on Custom Tree when the table is not translated

carlosaparadam:
When the table is not translated and the windows associated is tree type the window stay finding translation.

Screenshot_20190513_105725

Screenshot_20190513_105749


marcalwestf:
Just for testing's sake, and without any semantic thought, I created a custom tree for the window _Currency_, with the Type/Area=_Custom Tree_ and All Nodes=_Active_.
20190516_Trees_01

Then I made sure, the Table has columns _Name_, _Description_ and _isSummary_:
20190516_Trees_02
20190516_Trees_03

And define the tab to have a tree:
20190516_Trees_04

marcalwestf:
When trying to open the Currency window, the same error as @carlosaparadam Carlos is thrown (see above).

.. but if running it from the pull request branch, the window opens correctly:
20190516_Trees_05

PR #2565, opened 2019-05-21
Fix/fix for #197 delete cost details for inactive cost elements

### LINKED ISSUE #197: GenerateCostDetails does not delete inactive cost types or methods

mckayERP:
Fixes #197.

Resubmitted.  See comments on PR #2384.

This pull request corrects the error in the MCostElement beforeSave method that limited the CostElement to a unique CostElementType.  This change only limits the CostElementType "Material" to a single record.  All other CostElementTypes can have duplicates.  This allows Landed Costs, for example, to be duplicated to allow detailed tracking of different types of landed costs.

The only change compared to #2384 is in commit 7ee290c.


marcalwestf:
@mckayERP  hi Mike, can you give me a hint on how to set up a test environment (before-after)?
I'd then proceed to test.
Thanks.

SusanneCalderon:
Hi @mckayERP 
I will test it today.

mckayERP:
Hi Mario, Susanne,

You can use a fresh import of the seed. There are several deactivated cost elements in the seed.  Activate them, generate the cost data and try to delete them.   The deletion should also delete the related cost details.

Mike

Mike

SusanneCalderon:
Hi @mckayERP ,

Tested:

- Delete inactive Costelement, ok
- Create MCost costinglevel batch, ok

Regarding GenerateCostDetail I have a question. I do not see where the inactive costelements are deleted in MCostdetail. I only find it in Delete from MCost.
```

 private void deleteCostDimension(String trxName) throws SQLException {
    	
    	StringBuffer sqlDelete; private void deleteCostDimension(String trxName) throws SQLException {
    	
    	StringBuffer sqlDelete;
    	sqlDelete = new StringBuffer("DELETE FROM M_COST WHERE ")
    						.append(deleteCostWhereClause);
    	
        DB.executeUpdateEx(sqlDelete.toString(),
                deleteCostParameters.toArray(), trxName);

    }
```
The string deleteCostWhereClause is completed by the inactive costtypes.

```

 deleteCostWhereClause.append(" OR (").append(MCost.COLUMNNAME_M_CostType_ID).append("=? ")
            		.append(" AND EXISTS ("     // only delete costs related to inactive types
	            				+ "SELECT M_CostType_ID "
	            				+ "FROM M_CostType "
	            				+ "WHERE IsActive='N' AND M_CostType_ID = M_Cost.M_CostType_ID)")
            		.append(") ");
```
So the inactive costelements are deleted in MCost and not in MCostDetail. When you apply a transaction (shipment for example) the costs are the sum of the corresponding costdetails. 
I think the inactive costelements should be deleted from MCostDetail, too.

mckayERP:
Hi Susanne,

Sorry for the delay.

In GenerateCostDetail, see the method `deleteCostDetail`.

    /**
     * Delete Cost Detail
     *
     * @throws SQLException
     */
    private void deleteCostDetail(String trxName) throws SQLException {
        StringBuffer sqlDelete;
        //	BR [ 405 ]
        sqlDelete = new StringBuffer("DELETE FROM M_CostDetail WHERE ");
        sqlDelete.append(deleteCostDetailWhereClause);
        DB.executeUpdateEx(sqlDelete.toString(), deleteParameters.toArray(), trxName);
    }

The method `applyCriteria` sets the value of the `deleteCostDetailWhereClause`.  The default is to delete everything as limited by the criteria.

Also, in the models in the `beforeDelete` method in MCostElement, when the element is deleted, it will also delete all the associated Cost Details and Costs.

 Is this what you were looking for?

Mike

SusanneCalderon:
Hi @mckayERP , hi @e-Evolution 
Thanks, I think it is correct to define only one material cost.

e-Evolution:
Hi @SusanneCalderon Susanne, @mckayERP Michael

When the costing machine was redesigned the possibility of multiple parallel costs was considered, I mean that in ADempiere you can calculate different cost methods for the same company, it is important to mention that the cost method that will be used to register the accounting is the one defined in the type of cost selected in the accounting scheme.

Additionally, ADempiere will generate costs for all types of cost enabled.

In the Garden Admin records there were some obsolete cost records, because a cost element was created for each costing method, which was an incorrect approach because the cost elements are defined by the nature of the material cost, hand of work, indirect, additional costs, etc., independently for what method and type of cost are used.

In other words, the material cost element applies to any type of cost and cost method, using the same for everyone, but now Adempiere has control of each type of cost in the accounting dimension, so it is not necessary to create cost element for each costing method as before.

The system currently creates cost dimensions for all cost items marked as default.

And when a new company is created, the material cost element is created as default to be able to start with any configuration, by default standard cost is used.

yamelsenih:
Hi, exists a news here? What is the next step?

Best regards

yamelsenih:
Hello @mckayERP I will to close this pull request. Please if you have any change then a new pull request is welcome.

Issue #2570, opened 2019-05-21
The payroll process should be not always updated in payroll movement


e-Evolution:
Current the window for payroll movement can edit payroll process this is incorrect because can cause some inconsistence in the data.

PR #2572, opened 2019-05-21
#2570 The payroll process should be not always updated in payroll mov…

e-Evolution:
…ement https://github.com/adempiere/adempiere/issues/2570

Fixes: #2570


yamelsenih:
Great catch

PR #2579, opened 2019-05-27
Missing UUID are generated in seed base

### LINKED ISSUE #2564: Line with centralized id does not have uuid in the seed
I'm testing a seed database created from the TAG392 source and importing a PackIn but I'm getting a Duplicate Key error in AD_Column.

This happened because this row (AD_Column_ID=92435) havn't a UUID and is in the seed, causing that ADempiere try to create it instead of update it.

In the table AD_Column ther are 71 rows without UUID with centralized ID.

rcapecce:
Fixes #2564


yamelsenih:
Hi @rcapecce are you sure of size of file?

The reported bug said that exist many columns without UUID but this file is too many long, please check your seed.

I will to close it, make a new pull request with a little file with only missing UUID

Issue #2581, opened 2019-05-27
Add session reference to process instance


yamelsenih:
Currently exists many ways for make user audit, you can see it:

- Session
  - User Change
- Process Instance
  - Process Instance Parameters
  - Process Output
  - Process Log

The problem is that you don't have a way for link session with a process instance, it is nice to have for get current process from session

yamelsenih:
I have a change for add session to process instance

PR #2582, opened 2019-05-27
Add Session reference to process instance

yamelsenih:
Fixes: #2581 

This pull request allows link session to process instance.

Now you can see all process running from current session, see video:
Current ogv


yamelsenih:
Note that **AD_Session_ID** column is not mandatory for old compatibility

PR #2621, opened 2019-06-05
#2620 Serious error when validating the price limit in the sales orde…

### LINKED ISSUE #2620: Serious error when validating the price limit in the sales order or invoice line
## Bug report
Currently for the system validates that it is not possible to sell below the limit price when the "EnforcePriceLimit" flag is activated and the role of the user entering the sales order has not activated the flag to overwrite price limit.

This works only the first time that a line is entered and a price is placed below the limit.

but once the line was captured the salesperson can re-enter the order sales line in other time and change the price below the price limit and the system allow to erroneously entry and save a price below the limit, which is completely incorrect.

#### Steps to reproduce

1. Create a new Sales Order
2. Add a new line for Azalea Bush and modify the price by 6 USD
image

image

3. Close the window for this Sales Order
4. Get the previos Sales Order and go the line 
5. Change the price under price limit 6 USD and system allow to save new price

image

- Your OS: MacOS
- Java version: JDK 8
- ADempiere version: All
- Browser  safari
- DB Provider: PG
- DB version: ALL

e-Evolution:
…r or invoice line https://github.com/adempiere/adempiere/issues/2620

resolved #2620

### DIFF
--- a/base/src/org/compiere/model/CalloutInvoice.java
+++ b/base/src/org/compiere/model/CalloutInvoice.java
@@ -349,7 +349,6 @@ public class CalloutInvoice extends CalloutEngine
 		mTab.setValue("C_Currency_ID", new Integer(pp.getC_Currency_ID()));
 	//	mTab.setValue("Discount", pp.getDiscount());
 		mTab.setValue("C_UOM_ID", new Integer(pp.getC_UOM_ID()));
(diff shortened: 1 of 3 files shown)


yamelsenih:
I think that is better if it is validate for model

e-Evolution:
@yamelsenih 

Actually active interaction with the sales person  is required, I can not wait to save the line to make a decision on the price, freight, tax.

And that is the inteción of the callouts to have an interactive validation.

PR #2640, opened 2019-06-18
Fixed problem with new value as null, this bug was introduced by myself

yamelsenih:
Fixed error with new value as null.

Test case (As Garden World):
 - Go to Business Partner
 - Select a BP with a language configured
 - Unselect language
 - Save

The result is that language is not setted as null

This was introduced by myself here: https://github.com/adempiere/adempiere/pull/2628 I apologize by it


marcalwestf:
... reproduced the error.
Now testing the solution...

PR #2643, opened 2019-06-18
Add support to parent column and order column for tree window

carlosaparadam:
This pull request, add support to parent identities and order columns for tree windows. In the tree definition add sort and parent Columns, for setting parent record and order column for the tree. 

On this example i setting ParentElementValue_ID for parent and Value for sort column:

image

for test parent setting i add a new accounting element value with parent account 14-Inventory:

image

on save this element value this is regrouped inside 14-Inventory branch on the tree:

image

for test sort setting i add a new accounting element value with value 13.1:

image

on save this element value this is regrouped inside 13-Investments branch on the tree:

image

### DIFF
--- a/base/src/org/compiere/model/I_AD_Tree.java
+++ b/base/src/org/compiere/model/I_AD_Tree.java
@@ -49,6 +49,21 @@ public interface I_AD_Tree
 	  */
 	public int getAD_Client_ID();
 
+    /** Column name AD_ColumnSortOrder_ID */
+    public static final String COLUMNNAME_AD_ColumnSortOrder_ID = "AD_ColumnSortOrder_ID";
+
+	/** Set Order Column.
+	  * Column determining the order
+	  */
+	public void setAD_ColumnSortOrder_ID (int AD_ColumnSortOrder_ID);
+
+	/** Get Order Column.
+	  * Column determining the order
+	  */
+	public int getAD_ColumnSortOrder_ID();
+
+	public org.compiere.model.I_AD_Column getAD_ColumnSortOrder() throws RuntimeException;
+
     /** Column name AD_Org_ID */
     public static final String COLUMNNAME_AD_Org_ID = "AD_Org_ID";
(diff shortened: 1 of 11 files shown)


marcalwestf:
HI Carlos @carlosaparadam , I imported the XML file without errors and could reproduce the example of window _Account Element_.

Then I proceeded to test another trees, for example _Organization_, towards what I first configured the corresponding tree:
20190618_Tree_02

But then, when opening the _Organization_ window, I get an error:
20190618_Tree_03
Stating that column `ad_orginfo.issummary does not exist`, which is true when looking at the database.

If I then nevertheless do try to create a new Organization, the very same error pops up when pressing the _New_ icon. Moreover the _Parent Organization_ field displays nothing:
20190618_Tree_04

marcalwestf:
Additionally, two small remarks:

- _Release No_ should be =**3.9.3**
- _Entity Type_ is usually in Migration =Dictionary, whereas here is Spin

carlosaparadam:
hi @marcalwestf, this error is by where clause on table reference i will correct it and change entity type.

image

marcalwestf:
thanks @carlosaparadam , I reset the database to seed Adempiere 392 and imported the XML file.

I proceeded to test organizations, for which I configured the _Organization_ tree to order by the field _Value_:
20190618_Tree_06

Then, I created a new summary organization called _Understores_ and created 3 organizations, which differ in the _Value_ and in _Name_ fields:
20190618_Tree_05

According to the _Organization_ tree order column, the tree of the Organization window should be ordered depending on field _Value_.
This means, the order of the entries after Understores should be

1. 1000 Prior-to-First Organization
2. 1500 Middle Organization
3. 2000 First Organization

Instead, you can see another order on the tree of the last image. It is the order of creation.

carlosaparadam:
hi @marcalwestf, the target of the sort column is insert automatically the nodes inside parent node by coincidence.

PR #2651, opened 2019-06-24
fixed error with language for process from ASP change

yamelsenih:
Hi everibody, this pull request resolve a little bug with ASP for process name, the step for reproduce it is follow:

- Make login as System Administrator using es_MX language:
 - Go to Role Access Update
 - Run process
- Now make login as System Adminitrator using en_US (Base Language):
 - Note that Role Access Update now is showed like: **Actualizar Rol** 

See this gif:
Before_Change

See the process name, description and help changed:
Screenshot_20190624_132626

The problem is that tha cache object is used for change process statistics instead loaded object.

After change you can see the same test with cache reset:
After_Change


e-Evolution:
Hi @yamelsenih 

The change is right but I think that if you use an after saved is the best use the logic with a transaction and remove the legacy code save by saveEx

yamelsenih:
Hi @e-Evolution no problem.

e-Evolution:
Hi @yamelsenih 

2 binaries files come with last commit can please remove , this not should be updated

yamelsenih:
Thanks @e-Evolution I was created other PR here: https://github.com/adempiere/adempiere/pull/2652

PR #2680, opened 2019-07-20
Fix error on generate cost detail with reversed inventory documents

carlosaparadam:
When one inventory document is reversed and generate cost transaction the process delete the cost transaction but don't generate again.

see this example:
Bad_Generate_Reversal_Cost_Transaction

see this example with the fixes:
Fix_Generate_Reversal_Cost_Transaction


yamelsenih:
Hi @e-Evolution can you review it?

SusanneCalderon:
Hi @carlosaparadam 
I reviewed the code and think it is ok.
Regarding ProjectIssue , we made an enhancement which treats a ProjectIssue as as document with reversal_ID and method void and reversecorrectIt, in a new  class MProjectIssue.
I will raise an issue  and when your PR is included I will enhance this new class MProjectIssue with the method isreversalparent so that in case our PR is accepted it will not clash with your improvement.

PR #2681, opened 2019-07-24
Add Support to Project Based on Items

carlosaparadam:
Currently, the Project managment on Adempiere is based on Project, Phases and Task, this limits on projects more complex.

The purpose is create one figure of  project items in the when one item could be a phase, task, activity, milestone, revision, mistake... all in only one figure defined as a tree.

Advantages of this change:

- Maintenance only two dimension (Project and Project Item),not as currently  (Project, Project Phase, Project Task)

- One only entity for definition of one project item.

- More flexibility for developing for the project, being that only will working with one entity, not as worked currently with Project Phase, Project Task And Project Line.

see this preview:

Project_Based_On_Items


yamelsenih:
Hi everyone, I update this branch just now and work. Let a step by step here:
- Import XML: **07920_Add_Support_Project_Based_Project_Items.xml**
- As Garden World: Update role and re-Login
- Go to Tree and create two tree:
  - **Standard Project Line**:
    - Type: Custom
    - Table: C_StandardProjectLine
    - Parent Column: Parent_ID
    - Order Column: Name
    
Screenshot_20210707_160514


  - **Project Line**:
    - Type: Custom
    - Table: C_ProjectLine
    - Parent Column: Parent_ID
    - Order Column: Name
    
Screenshot_20210707_160501

- Go to Project Management -> Project Line Type and create a new
- Go to Project Management -> Project (Lines) and create a new project
- Go to Project Items Tab and create lines, note that is showed as tree
Screenshot_20210707_160554

yamelsenih:
@marcalwestf @e-Evolution can you check it? I update the branch and XML for last seed and work for me

Best regards

e-Evolution:
@carlosaparadam Hi Carlos! The idea sounds pretty good, but back compatibility has to be resolved.

When Compiere designed the projects, there was no way to support a tree for any table, as ADempiere currently supports this is really good. but I wonder how we could support back compatibility.

The projects were originally intended to support project financial control, so their structure is strongly tied to this approach.

So today we can follow any project by its references C_Proyect_ID, C_ProyectPhase_ID, C_ProyectTask_ID, that is, all the documents that generate an entry in the accounting allow the registration of this structure.

The Fact Accounting and Reports, Reporting Hierarchy Report Cubes, Financial Report Engine uses C_Proyect_ID, C_ProyectPhase_ID, C_ProyectTask_ID structure, if we implement the new approach that you propose would imply losing these benefits and this is not acceptable.

The right question here is how can we ca to support the current have and at the same time implement this enhancement?

Would we simulate the same behavior if affect what currently so as not to lose the current functionality?

Project Dimension in document:

image

Fact Accounting:

image

Report Cube: 

image

kind regards
Victor

yamelsenih:
Hi @e-Evolution this just allows create project as tree and using project lines like is Open Project and other standard project management. Is nice to have as part of core.

Note that thids change keep the same functionality old

marcalwestf:
... xml loaded and applied without errors

marcalwestf:
I found an error: it is possible to create _Project Line Tpyes_ without names:
image

For once, I don't know if it would be possible to create more than one _Project Line Tpye_; on the other side, in window _Project Type,_ tab _Standard Line_ the combo box would show a -1, which would be ambiguous:
Selection_037


After setting a proper name, the combo box displays a proper name:
image

Steps to correct this error: define field _Name_  in window _Project Line Tpye_ as (table _C_ProjectLineType_) mandatory.

e-Evolution:
> Hi @e-Evolution this just allows create project as tree and using project lines like is Open Project and other standard project management. Is nice to have as part of core.
> 
> Note that thids change keep the same functionality old

Hi @yamelsenih , If I understand well the approach that is being sought is to allow a line to be selected as a phase or task in the project line type, then how will this be reflected in the accounting, since the project line type is not are into the any document or the fact account, the idea is that work as now work the phase and the task into of all documents and fact accounting generated.

As I can see the functionality is being duplicated because now two different approaches will be allowed to manage a project using the phase or task as it currently works or project line type. but the second option does not implement the accounting functionality

I wonder we shouldn't look for a way to unify and leave one focus and deprecate the other.

For this we must resolve how the phase or task is handled today.

An idea would be to add project line type in all the documents that have project and in fact_acct and find a way to emulate the current behavior.

There should also be a strategy and migration route from those who today have that approach to the new
kind regards
Victor

marcalwestf:
continuing the test, creating a project with the new window _Project (Lines)_ and adding _Project Lines_ on the tab _Project Item_ as suggested by @yamelsenih :
image

marcalwestf:
... now referencing this project in an Order:
image
..and Order Line
image

marcalwestf:
..problem is only when zooming the project on documents, the "wrong" (i.e. the "Standard") Project window is displayed...

marcalwestf:
invoice:
image

Invoice line:
image

So I suppose, the project can be referenced from **any** document.

marcalwestf:
now testing the actual project functionality _Project Issue_: select a Material Receipt that should be consumed by the project:
image

And here the result:
image


The report _Material Transactions_ delivers the consmption:
image

Here the documentation of the Project Issue in window _Project (Lines/Issues)_
image

marcalwestf (approved):
- XML loaded and applied  -> OK
- Projects created with lines ordered as tree in random depth  -> OK
- Project Issue  -> OK
- Verify all results around Project Issue  -> OK

Result: 

- this PR enables to organize Project Lines as trees.
- case 1: if a user chooses to work only with Project Lines, he can do it with the difference that the lines can additionally be organized as a tree of any structure.
- case 2: If the user chooses to create the Project Line tree below phase or task, it is also possible.
- case 3: if a user chooses to work with phases/tasks/lines without organizing the project lines as tree as hitherto, he can do it. This implies a backward compatibility.

marcalwestf:
by the way... 
while testing this functionality, I found some errors: https://github.com/adempiere/adempiere/issues/3510

e-Evolution:
@carlosaparadam please can update with develop branch and solve the conflict , this way I can review this enhancement

e-Evolution:
Dear @carlosaparadam, @yamelsenih, @marcalwestf, @niclopup  here is my detailed analysis:

After testing and reviewing this idea I think it has a lot of potential.

After a detailed analysis these are the lines of action that will satisfy the backwards compatibility and the new requirements without requiring major changes or duplicating the data model.

The great dilemma of the current structure is that the phases hinder more than they help, complicating the management of a hierarchical tree and complicating navigation, so starting from the premise how we can eliminate the phases and add hierarchy in the tasks and at the same time do not duplicate the model.

1.- Add the hierarchical tree behavior to the standard tasks and tasks, this is achieved only by adding the IsSummary field to the C_ProjectTask and C_Task table

On the other hand, it is not required to duplicate the model for Project Line Type (C_ProjectLineType) and Standard Line C_StandardProjectLine, now we are two tables


2.- We deprecate the C_ProjectPhase and C_Phase tables and create a callout so that when the task is selected anywhere in the system, the C_ProjectPhase field is filled with the hierarchy established in the tree.


3.- Regarding Project Line type, we could preclude from this field since in the task we have the Category for the task but I have no problem that a new classifier called task type is added.

This approach has the advantage of two approaches in one with the following benefits.

1.- The current structure is used and the use of the phase that does not allow simple and practical navigation is discontinued.

2.- Duplication of the C_ProjectTask and C_Task model with C_ProjectLineType and C_StandardProjectLine and all their fields is avoided.

3.- It is allowed to support financial accounting because the phase is replaced by the parent task id to which the document is linked, it can be easily solved with a callout.

4.- The hierarchy is added in the task table that is very useful visually using the tree navigation.

5.- The current functionality of Project Line is not changed, which is designed to link the tasks to different documents, that is, for a tariff there are multiple lines.

6.- Only one process is required to migrate the phases to accumulated tasks for existing projects, so the migration would be relatively easy and not very painful.

7.- Total backward compatibility for customers who already use this functionality.

kind regards
Victor

yamelsenih:
I think that this is a greatl functionality. 

I not have problem for close it and implement on myself.

I have only a requirement: If any want implement some similar to it then I think that should be take this pull request as start point

Issue #2690, opened 2019-07-26
Not is possible generate Invoice on Credit from POS


erprafaelrodriguez:
Hello Dear Team ADempiere.

Im doing a exercise of sales from POS with any Document Type: On Credit Order and Order Standar but my client not make payment, it is a Credit Sale...

My impression is that all Sale generate from POS must be payment receipt to generate the Invoice  **(Remember that it is a Credit Sale)**

- The POS have option "Generate Inmediate Invoice" but not is funtional.
- The POS have option "Complete Prepared Order" but only leave In Process the Order Sales.

#### Steps to reproduce
1. Generate Sales from POS
2. Selection Document Type Order Standar or On Credit Order and Order
3. Option Generate Immediate Invoice
4. Option Complete Prepared Order (Result: In Process the Order Sales)

#### 
<img width="1381" alt="Captura de pantalla 2019-07-26 a la(s) 3 44 19 p  m" src="https://user-images.githubusercontent.com/20327157/61980754-43a3d180-afbd-11e9-8702-da015c93be12.png">

<img width="1380" alt="Captura de pantalla 2019-07-26 a la(s) 3 49 43 p  m" src="https://user-images.githubusercontent.com/20327157/61980766-48688580-afbd-11e9-996e-67fa6f573d7b.png">

<img width="1381" alt="Captura de pantalla 2019-07-26 a la(s) 3 53 18 p  m" src="https://user-images.githubusercontent.com/20327157/61980883-9f6e5a80-afbd-11e9-9730-8f867885d838.png">

<img width="1381" alt="Captura de pantalla 2019-07-26 a la(s) 3 54 51 p  m" src="https://user-images.githubusercontent.com/20327157/61980939-cb89db80-afbd-11e9-995c-e63e4626419a.png">

#### Expected behavior
Complete the Sales Transaction but a Credit (Not Generate Payment)

#### Other relevant information
- Your OS: iOS, Linux
- Java version: 1.8.0_151
- ADempiere version: 390a, 391
- Browser chrome, safari
- DB Provider: PostgreSQL
- DB version: 9.6 and 11

#### Additional context
Pending your Comments

PR #2691, opened 2019-07-27
 #2690 Cat not Complete Sale Order on Credit using Complete prepared …

e-Evolution:
…Order from POS

https://github.com/adempiere/adempiere/issues/2690


marcalwestf:
.... the change itself is ok, but the Travis checks fail. 
I don't know how to overcome this problem.

Issue #2692, opened 2019-07-27
When find Business Partner from POS get SQL error


e-Evolution:
## Bug report

When find Business Partner from POS get SQL error

#### Steps to reproduce

1.- Login as GardenAdmin , GardenWorld and Store Central
2.- Open POS and create a new POS Order
3.- Set some date the Business Partner field and try find

#### Expected behavior
When the Business Partner field is filling then the POS should find based in value , taxid , name , name , contact , email ,phone , city

#### Other relevant information
- Your OS: MacOS 
- Java version: 1.8
- ADempiere version: 391 , 392
- Browser: safari
- DB Provider: PG
- DB version: 11.2

PR #2693, opened 2019-07-27
#2692 When find Business Partner from POS get SQL error https://githu…

e-Evolution:
…b.com/adempiere/adempiere/issues/2692


marcalwestf:
same situation as in pull request https://github.com/adempiere/adempiere/pull/2691 .
Waiting to solve check conflicts to accept this pull request.

Issue #2696, opened 2019-07-31
Can not call a Financial Report background  process and process transaction 


e-Evolution:
#### Bug report
Can not call a Financial Report background  process and process transaction 
#### Steps to reproduce
Create a process to call the financial report background  process passing of transaction

`            val executeFinancialReport = ProcessBuilder.create(ctx)
              .process(FinReportAbstract.getProcessId())
              .withRecordId(I_PA_Report.Table_ID, financialReportId)
              .withParameter(FinReportAbstract.C_PERIOD_ID, periodId)
              .withoutBatchMode(
              .withoutPrintPreview()
              .execute(trx.getTrxName)`

#### Expected behavior
Use the transaction from the process that was called

PR #2697, opened 2019-07-31
#2696 Can not call a Financial Report background process and process …

e-Evolution:
…transaction https://github.com/adempiere/adempiere/issues/2696


marcalwestf:
same situation as with https://github.com/adempiere/adempiere/pull/2691 and https://github.com/adempiere/adempiere/pull/2693 : conflicts have to be solved.

Issue #2740, opened 2019-08-28
NPE when set Bank Statement Line description


e-Evolution:
# Bug report
NPE when set Bank Statement Line description

#### Steps to reproduce

Create a line bank statement and try save

#### Expected behavior

the line should be save

PR #2741, opened 2019-08-28
solve issue #2740 NPE when set Bank Statement Line description 

e-Evolution:
solves #2740


marcalwestf:
hi @e-Evolution  Victor, how can I reproduce the error and test the sulution?
Thanks

Issue #2742, opened 2019-08-30
[Feature Request] Generate receipt from Inbound order


yamelsenih:
## Feature request
Generate recipt automatically from Inbound Order.

#### Describe the solution you'd like
Is nice to have a feature that allows generate receipt automatically from express receipt when is complete then.
#### Describe alternatives you've considered
Add a process for it but the express receipt should generate automatically it.

#### Additional context
None

PR #2743, opened 2019-08-30
Generate receipt automatically from Express Receipt

yamelsenih:
This pull request allows create receipt from express receipt after complete 
Fixes: #2742


e-Evolution:
Dear @yamelsenih , Thank a lot to implement this logic is great .

Can please adding the logic to validate that receipt only is generate if isSOTrx = 'N'

`//	Generate receipt
 	if (!IsSOTrx())  generateReceipt();
`

Issue #2747, opened 2019-08-31
[Feature Request] Generate Freight Order from Outbound


yamelsenih:
## Feature request
This feature is very easy for generate a freight order from a outbound order

#### Describe the solution you'd like
Just a process for generate a freight order from a outbound order
#### Describe alternatives you've considered
A smart browser with it but this process will be a first solution
#### Additional context
See a image
![Screenshot_20190830_234456](https://user-images.githubusercontent.com/2333092/64058965-c56cb900-cb80-11e9-95c1-0d8041202aa5.png)

PR #2750, opened 2019-09-01
Feature/#2747 generate freight order from outbound

yamelsenih:
Add process for generate freight order from outbound order
Screenshot_20190830_234456
Fixes: #2747


e-Evolution:
@yamelsenih Hi Yamel.

The idea of creating a line in the freight order is based on the total weight per freight category address of origin to destination, so it should not be created for each line item of order, since it would not make sense to have the same information.

The freight payment is made by freight category, and section of the direction of origin and destination (route)

yamelsenih:
> @yamelsenih Hi Yamel.
> 
> The idea of creating a line in the freight order is based on the total weight per freight category address of origin to destination, so it should not be created for each line item of order, since it would not make sense to have the same information.
> 
> The freight payment is made by freight category, and section of the direction of origin and destination (route)

Hi @e-Evolution I just add a process because already exist product reference inside Freight order as manual freight.

e-Evolution:
@yamelsenih  But the product or charge is use to the service with which the freight will be paid, not the product that generated the freight, that is in the Shipment or movement document lines

yamelsenih:
No problem, I want to add lines for freight order but I can change it very
easy no problem. Let me one minute

Thanks

PR #2751, opened 2019-09-01
Add Support in Tree for Account Element 

Raul-mz:
When you create Element Value  in ADempiere the tree is not updated and does not allow you to sort the accounts.

### DIFF
--- a/base/src/org/adempiere/process/ASPGenerateLevel.java
+++ b/base/src/org/adempiere/process/ASPGenerateLevel.java
@@ -108,7 +108,7 @@ public class ASPGenerateLevel extends ASPGenerateLevelAbstract {
 		int AD_Tree_ID = clientInfo.getAD_Tree_Menu_ID();
 		//	Yamel Senih [ 9223372036854775807 ]
 		//	Change Constructor
-		MTree thisTree = new MTree (getCtx(), AD_Tree_ID, true, true, null, get_TrxName());
+		MTree thisTree = new MTree (getCtx(), AD_Tree_ID, true, null, get_TrxName());
 		//	End Yamel Senih
 		MTreeNode node;
 		if (menuId > 0)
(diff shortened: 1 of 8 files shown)


marcalwestf:
@Raul-mz could you please write an scenario to test it?
Thanks

yamelsenih:
Hi @marcalwestf is very easy replicate it if you have a tenant with User1_ID enabled for account schema.

The step by step with a customer with User1_ID enabled

1. Open Account element Window
2. Go to Element Value tab
3. Create a new account with name and value
4 .Close Window and open again
5. Note that tne new account is missing

The problem is that **insertTree** method of **PO** make a insert on **AD_TreeNodeU1** instead **AD_TreeNodeEV**.

marcalwestf:
thanks @yamelsenih for the explanation , I will test it now

marcalwestf:
Hi @Raul-mz  , @yamelsenih I cannot reproduce the error with the current version of the _develop_ branch:

STEP 1: Create a tree:
Selection_141

STEP 2: Create an Element, referencing to the tree:
Selection_142

STEP 3: Create a dimension with Type=_UserList_1_ and the new created element:
Selection_143

STEP 4: Create an Element Value for this Element:
Selection_144

RESULT: The elements are displayed on the tree immediately and even after a _Cache Reset_.

I tried it first on Swing, for later test on ZK.
Maybe I chose a wrong workflow, here perhaps somebody can give me a hint about the correct testing steps.

Thanks.

yamelsenih:
hi @marcalwestf the problem is when you try add a element to account instead user 1

e-Evolution:
Hi @Raul-mz Raul, @yamelsenih 

It is bad practice to change the signature of public methods or construction as this can break things within the system itself or some customization, please can overload the methods leaving the original signatures and adding the new ones with the new signature

carlosaparadam:
replaced by this #3195

PR #2753, opened 2019-09-02
Add Vehicle Group Window

yamelsenih:
This pull request add a window for Vehicle group like a attribute more of vehicles
Screenshot_20190901_211500


e-Evolution:
@yamelsenih why no use the asset group for your goal ?

PR #2754, opened 2019-09-02
Feature/#sc add print document from process

yamelsenih:
This pull request allows print many document with one Ask.
Also allows select a print format or get from document type or document type target

You can use some helper methods:

```Java
printDocument(List<PO> documentList, int printFormatId, boolean askPrint)
printDocument(List<PO> documentList, boolean askPrint)
printDocument(PO document, int printFormatId, boolean askPrint)
printDocument(PO document, boolean askPrint)
printDocument(PO document, String printFormantName)
```

Note that old methos is keepped:

```Java
printDocument(PO document, String printFormantName)
```

See a result:

Current ogv


yamelsenih:
Hi everyone, any update here?

PR #2756, opened 2019-09-02
Fixed error with import bpartner

yamelsenih:
This pull request resolve a bug when a business partner is imported and have two users with the same email

See:
```Java
-------------------------------
===========> ImportBPartner.process: org.postgresql.util.PSQLException: ERROR: more than one row returned by a subquery used as an expression [89]
org.postgresql.util.PSQLException: ERROR: more than one row returned by a subquery used as an expression; State=21000; ErrorCode=0
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2440)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2183)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:308)
	at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:441)
	at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:365)
	at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:143)
	at org.postgresql.jdbc.PgPreparedStatement.executeUpdate(PgPreparedStatement.java:120)
	at com.mchange.v2.c3p0.impl.NewProxyPreparedStatement.executeUpdate(NewProxyPreparedStatement.java:105)
	at sun.reflect.GeneratedMethodAccessor43.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.compiere.db.StatementProxy.invoke(StatementProxy.java:100)
	at com.sun.proxy.$Proxy4.executeUpdate(Unknown Source)
	at org.compiere.util.DB.executeUpdateEx(DB.java:1118)
	at org.compiere.util.DB.executeUpdateEx(DB.java:1178)
	at org.compiere.util.DB.executeUpdateEx(DB.java:1169)
	at org.compiere.process.ImportBPartner.doIt(ImportBPartner.java:263)
	at org.compiere.process.SvrProcess.process(SvrProcess.java:175)
	at org.compiere.process.SvrProcess.startProcess(SvrProcess.java:128)
	at org.adempiere.util.ProcessUtil.startJavaProcess(ProcessUtil.java:160)
	at org.adempiere.util.ProcessUtil.startJavaProcess(ProcessUtil.java:105)
	at org.compiere.apps.ProcessCtl.startProcess(ProcessCtl.java:625)
	at org.compiere.apps.ProcessCtl.run(ProcessCtl.java:374)

```

### DIFF
--- a/base/src/org/compiere/process/ImportBPartner.java
+++ b/base/src/org/compiere/process/ImportBPartner.java
@@ -160,7 +160,7 @@ implements ImportProcess
 		//	Set Country
 		sql = new StringBuilder ("UPDATE I_BPartner i "
 				+ "SET C_Country_ID=(SELECT C_Country_ID FROM C_Country c"
-				+ " WHERE i.CountryCode=c.CountryCode AND c.AD_Client_ID IN (0, i.AD_Client_ID)) "
+				+ " WHERE UPPER(i.CountryCode)=UPPER(c.CountryCode) AND c.AD_Client_ID IN (0, i.AD_Client_ID)) "
 				+ "WHERE C_Country_ID IS NULL"
 				+ " AND I_IsImported<>'Y'").append(clientCheck);
 		no = DB.executeUpdateEx(sql.toString(), get_TrxName());
@@ -185,7 +185,7 @@ implements ImportProcess
 		//
 		sql = new StringBuilder ("UPDATE I_BPartner i "
 				+ "Set C_Region_ID=(SELECT C_Region_ID FROM C_Region r"
-				+ " WHERE r.Name=i.RegionName AND r.C_Country_ID=i.C_Country_ID"
+				+ " WHERE UPPER(r.Name)=UPPER(i.RegionName) AND r.C_Country_ID=i.C_Country_ID"
 				+ " AND r.AD_Client_ID IN (0, i.AD_Client_ID)) "
 				+ "WHERE C_Region_ID IS NULL"
 				+ " AND I_IsImported<>'Y'").append(clientCheck);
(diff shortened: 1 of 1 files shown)


e-Evolution:
@yamelsenih Hi Yamel

I am not sure if placing MAX in some data implies that you want to modify the same business partner, what happens when you have the same contact with same mail in different business partners, placing the MAX can cause bad behavior

marcalwestf:
by the nature of the SQL:
```
// Existing User. Lookup by AD_User.Email - BPartner
sql = new StringBuilder ("UPDATE I_BPartner i "
+ "SET C_BPartner_ID="
+ "(SELECT C_BPartner_ID FROM AD_User u "
+ "WHERE i.EMail=u.EMail AND u.AD_Client_ID=i.AD_Client_ID "
etc.
```
the result of the select can only be **one** ID. 

In this code snippet , the Business Partner to be imported is won out of the email of an existing user.
If it is the case that two users have the very same email, then the solution presented here is to solve it technically: take just one of them, without taking care of the implications.

Another -perhaps more dogmatic- solution could be not to set the Business Partner if the email is repeated on other users. 

I would accept the solution presented, because it eliminates an NPE.
Then I'd see to avoid repeating email on users.

The code is OK.

marcalwestf:
@e-Evolution Victor, here also, it would be required to know if you are against this pull request.

In my opinion, though this pull request -as you correctly comment-  does not solve the issue to a 100%, at least it avoids the Exception, which is what we must avoid at all costs. 

So if I'd have to decide whether to avoid an Exception or to have a less than 100% solution without Exception, I'd no doubt. That is why I am in favor of accepting this pull request.

PR #2758, opened 2019-09-03
Resolve problem with printing margin for iReport

yamelsenih:
This pull request resolve a legacy problem with paper print formats and orientation for print formats maked from iReport.

This problem happens with format like halfletter portrait

### DIFF
--- a/JasperReports/src/org/compiere/report/ReportStarter.java
+++ b/JasperReports/src/org/compiere/report/ReportStarter.java
@@ -13,7 +13,6 @@
  *****************************************************************************/
 package org.compiere.report;
 
-import java.awt.print.PrinterJob;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


marcalwestf:
@yamelsenih  how can this pull request best be tested??

On the one side we have to consider all reports which do work with JasperSoft Studio. Here the question is: what do we have to focus on? 

On the other hand -in legacy implementations- what must be done to test it? Open iReport files with JasperSoft Studio and compile them?

Thanks.

yamelsenih:
Hi @marcalwestf I have many customers running it and just affect:
- Direct Printing from iReport
- Printing for matrix point printers

Befor it I has a patch for resolve it from 3.6 versions to last

e-Evolution:
@yamelsenih Currently ADempiere uses the name of the printer in a print format to indicate to which print job a certain report will be sent when you have multiple devices to print in java, this code although it can solve the problem of margins, no is good remove the functionality to send a report created in jasper report to a printer defined in the print format.

image

please integrate your solution with the print queues which is as before

PR #2765, opened 2019-09-05
Add Report for Outbound order detail

yamelsenih:
Add Outbound order detail report for wath:
- Outbound:
  - Delivered
  - Invoiced
  - Open to Invoice
  - Open to Delivered
- All
Screenshot_20190905_130501


marcalwestf:
@yamelsenih  when testing another pull requests, I noticed that when executing _05220_Add_Outbound_Detail.xml_, the following error occurs 

> Application failed.  org.postgresql.util.PSQLException: ERROR: column il.wm_inoutboundline_id does not exist   Hint: Perhaps you meant to reference the column "iobl.wm_inoutboundline_id" or the column "dol.wm_inoutboundline_id".   Position: 2264

See the following screenshots:
Selection_131

Selection_132

yamelsenih:
Hi Mario the problem is that migration apply doesn't apply columns before
add view, note that if you go to table and synchronize column manual it
work.

PR #2770, opened 2019-09-08
Remove missing column XML

yamelsenih:
Related to #2747 generate freight order from outbound #2750
Remove missing column


yamelsenih:
I apologize for no include this change on before pull request

Issue #2303, opened 2019-02-01
MPPProductBOM.getDefault() doesn't find default.


mckayERP:
The code for MPPProductBOM.getDefault(MProduct product, String trxName) doesn't find the default if there are multiple BOMs.  The code simply returns the first one found that is active and has the same value/search key as the Product.  This will be the first bom created unless the BOM search key was changed.

The BOM record has a "IsDefault" column but this is not checked. 

Would it not be clearer for the user to use the IsDefault to select the default BOM?

Otherwise, its easy to break things by changing the search key.

Also suggest beforeSave checks to make sure at least one record has the default flag set.

PR #2772, opened 2019-09-09
Set IsDefault Field as default for search BOM, reference to issue:

yamelsenih:
This pull request set **IsDefault** column for BOM as Default search value

Fixes: #2303
https://github.com/adempiere/adempiere/issues/2303

Best regard

### DIFF
--- a/base/src/org/eevolution/model/MPPProductBOM.java
+++ b/base/src/org/eevolution/model/MPPProductBOM.java
@@ -14,7 +14,6 @@
  * Contributor(s): Victor Perez www.e-evolution.com                           *
  *                 Teo Sarca, http://www.arhipac.ro                           *
  *****************************************************************************/
-//package org.compiere.mfg.model;
 package org.eevolution.model;
 
 import java.sql.ResultSet;
@@ -129,8 +128,8 @@ public class MPPProductBOM extends X_PP_Product_BOM
 	 */
 	public static MPPProductBOM getDefault(MProduct product, String trxName)
 	{
-		MPPProductBOM bom = new Query(product.getCtx(), Table_Name, "M_Product_ID=? AND Value=?", trxName)
-				.setParameters(new Object[]{product.getM_Product_ID(), product.getValue()}).setOnlyActiveRecords(true)
+		MPPProductBOM bom = new Query(product.getCtx(), Table_Name, "M_Product_ID=? AND IsDefault = 'Y'", trxName)
+				.setParameters(new Object[]{product.getM_Product_ID()})
 				.setOnlyActiveRecords(true)
 				.setClient_ID()
 				.firstOnly();
--- a/migration/392lts-393lts/05280_Add_Old_Compatibility_BOM_Default.xml
+++ b/migration/392lts-393lts/05280_Add_Old_Compatibility_BOM_Default.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<Migrations>
+  <Migration EntityType="ECA02" Name="default for search BOM, reference to issue: #2772" ReleaseNo="3.9.3" SeqNo="5280">
+    <Comments>See: https://github.com/adempiere/adempiere/pull/2772</Comments>
+    <Step DBType="ALL" Parse="Y" SeqNo="10" StepType="SQL">
+      <SQLStatement>UPDATE PP_Product_BOM SET IsDefault = 'N';
+UPDATE PP_Product_BOM SET IsDefault = 'Y' WHERE EXISTS(SELECT 1 FROM M_Product p WHERE p.M_Product_ID = PP_Product_BOM.M_Product_ID AND p.Value = PP_Product_BOM.Value);</SQLStatement>
+    </Step>
+  </Migration>
+</Migrations>


e-Evolution:
Hi @yamelsenih the current functionality for default BOM in Manufacturing by convention is uf bom search value is same that product search value ; MRP use this convention , if you need define other configuration then can use planning data for product, this tab can define the default bom for any combination organization, warehouse, plant and product.

So I not sure if this change cause confusion the user that currently use manufacturing

yamelsenih:
Hi Victor, if exists other effect I can change it. Also I think that is
nice to have a query for set all bom as default when Value are the same of
product. IsDefault field is used in all ADempiere for identify if a record
is flagged as default.

Best regards

marcalwestf:
As @e-Evolution  says, there is a convention that when the _BOM Value_ is the same as the product-to be-produced's value, this BOM is taken as a the default BOM and the column _isDefault_ is set to true. This is also how MRP works.

So far so good.

On the other hand, if somebody -for whatever reason- wants to set a BOM to become the new default BOM even if there is already a BOM which has the same value as the value of the product to be produced, it must be possible to do it. There is no reason to deny it.

Not allowing this would constraint the users to change the values of the BOMs involved when wanting to change the default BOM. 

How to solve this conflict?
The solution would be to change MRP to look for _isDefault_ instead of looking for _Value_. 

Problematic would be for **legacy installations** when migrating them to new Adempiere versions.  Perhaps there are installations that do not consider the field _isDefault_ at all.

Here the solution would be to create and execute an SQL update which makes sure that all BOMs which have the same value as the product become _isDefault_.

So if this pull request additionally
1. changes MRP to look for _isDefault_ instead of for value and
2. delivers a SQL query updating all BOMs for legacy implementations

I would say it corrects the actual behavior and preserves legacy.

e-Evolution:
Hello Yamel, Mario!

Well, a subject that I think is not working well in ADempiere today is the Is Default flag for any data entity, given that today it is possible to place multiple records as an is default record.

On occasion, I had to lose some hours of work looking for why the system behaves in a certain way and at the end, it was that there were two records with the Is default flag activated.

This should be changed and validated as part of the core of the persistence engine. but this other subject.

Regarding the  the Bill of Materials (BOM) codes.

First, we have to understand that they are BOM Codes, in the past, in old versions, you could only have one list of materials per product, which was not efficient, therefore the codes of material lists were implemented, that implies that a product,  now can have multiple lists of materials for manufacturing.

Thus in the world of manufacturing, there are two important concepts for the Bill of Materials:

1.-A code of Bill of Material Default, it is the one that has the same name as the product code by convention, so by default, you do not have to configure anything.

In general, all manufactured products start with a list of materials and this is encoded with the same product code.

In some cases, there is a need to create codes of lists of materials different from the standard and this is where the second concept, the lists of alternative materials, arises.

2.- The Alternative BOM are those that have a different code than the product. and they are generally created for the following reasons.

- Design changes or product improvements
- Significant changes in the cost
- Changes of raw material by conditions of the plant or the place where it is manufactured.

Thus the definition of a  Standard BOM of an alternate is implied by the convention.

This happens in the same way for standard and alternate Workflow Manufacturing (WFM).

To define that BOM you want to use for the planning of production processes, the system has the tab of product planning data where you can determine which is the BOM or WFM that will be used for a certain configuration, this allows a configuration more specify and granularity that just marks a flag.

image

Since a list of materials may be predetermined for one plant, but not for another.

So today it is possible that when creating a manufacturing order or approving it you can select the list of materials with which you want to manufacture a product, so the user has no limitation.

image


But this brings great implication since the system performs the calculation and cost integration based on the standard BOM and WFM standard , this is crucial for standard cost accounting since with this the system can calculate all variations of the standard cost, variation in method, variation in use, variation in rate.

In summary I do not see much sense to add the default field when there is already a convention that does not require any action other than naming the code of the BOM with the same code as the product, and if some other configuration is required this can be defined in planning data, which implies not only defining the BOM or WFM,  but also the order modifiers that generally change by plant or region.

Making this change involves changing MRP, Cost Calculation and some other logic, but there is little benefit.

It is not easier for Production Light to adopt the conventions of manufacturing functionality, on the other hand, the last thing I have left is that Production Light was intended to be something simpler than manufacturing with MRP, we are not duplicating functionality to solve the same problem ??

So, my question is what you are trying to solve that now ? 

can not solve based on convention?

yamelsenih:
Hi, I add a old compatibility, I don't know what is the reason for use default as Value instead natural and legacy Compiere / ADempiere **IsDefault** flag. 

If the problem is a method getDefault then I was changed if for resolve for **IsDefault**, if the problem is Old compatibiility then I was add a XML change it and ready.

e-Evolution:
@yamelsenih 

Well, the problem is more complex than a default BOM, there are Standard and Alternate BOMs, a standard BOM is used to determine the standard cost and standard planning.

maybe the confusion is that we define a  BOM as default is same that Standard BOM

My question what is the goal of having a default BOM

What is it for?
What is the objective?

What I understand is that the logic of using the Standard BOM does not cover your requirement, even if when requirement it is not clear.

Perhaps the easiest is to add the BOM column and have the user choose or have your logic simply search for the first one they find. Be it standard or alternate.

To me, in manufacturing, this is more complex than a flag that turns on or off so that depends on the multiples factors

marcalwestf:
Let me try to summarize:

1. @e-Evolution comments that it is costume (good practice?) that a BOM which has the same _Value_ as the product to be manufactured is considered to be the _standard BOM_. In other words: the standard BOM is expected to have the same value as the product.

2. So it is implemented in ADempiere: in the different parts where the standard BOM is searched for, it is expected that the _Value_ is alphanumeric equal to the product's. No entry found with the same value? This means no standard BOM.

3. @yamelsenih suggests (at least so I understand) that additionally to this implicit standard BOM the field _isDefault_ is used: when there is no standard BOM defined in an implicit way as explained in item 1, then it should be looked for an entry which has the field _isDefault_ set.

4. This would need the _isDefault_ field to be unique for the different BOMs of a product, which is not the case today: this pull request shuold add the constraint.

5. If implemented according to the items 1-4, then implementations using the implicit standard BOM would work as well as the ones wishing not to. Having the implicit standard BOM definition preponderance over _isDefault_, legacy implementations are not affected.

e-Evolution:
As I said it makes no sense "is default flag"  for BOM , Standard is not the same concept as "Is default".

What happens if I want a default for organization 1 and another one predetermined by organization 2 or any other criteria, that is to say, it makes no sense.

On the other hand, the concept of the standard BOM has nothing to do with the is default concept , is here where all its confusion is.

A standard code BOM defines the standard cost of materials and labor for the finished product, defines the standard planning for MRP of a manufactured product.

Thus the Rollup costs are calculated based on the list of standard BOM and standard workflows, the explosion of materials to generate the suggested manufacturing order use the standard BOM and this is only overwritten by the specific definition of a combination of parameters in the planning data

This is universal for all organizations, plants, and warehouses.

So if not exist a Standard BOM the system no calculate costs, not generate planning order because not exist a standard,  although when exist alternate BOM 

When creating a material list code, the system automatically takes the product values, which guarantees that there is a list of standard materials, also in this way I can know that those material lists that have a search key other than the product are alternate materials list.


About Production Light
So if even so, the standard BOM not is useful to you and require a different logic for your production light functionality, 

the most appropriate thing is that you add PP_Product_BOM_ID to your production order so that the user can choose of BOM. anyway this functionality is support currently in Manufacturing Order,  maybe your customer need move to Manufacturing Management that is more advanced.

If you still need to mark a list of materials in some way, I recommend that you make a new window that satisfies your requirement so that users do not confuse their concept of is default with a standard BOM

On the other hand I agree with Yamel that if today the IsDefault field causes confusion it will be hidden from the BOM

marcalwestf:
reviewing the open pull requests, I notice this is a small on, it is not resolved, though heavily commented.
It seems @e-Evolution  Victor has some issues with it.

IMO, the changes suggested here are acceptable, if accompanied by a SQL that makes sure the isDefault is unique for  a product BOMs.

As I understand, these changes are not disruptive to Libero nor break legacy implementations.

So, in order to accept or reject this pull request, I'd like Victor to state here the approval or rejection so we move on.

yamelsenih:
I can resolve it with a Model validator without complications, is very strange that a little and standard flag used for any entity is unused here but repect all comments.

Best regards

Issue #2773, opened 2019-09-09
[Feature Request] Allows select a BOM for Production Light


yamelsenih:
## Feature request
This feature allows select a specific BOM for a Production Batch Order.

#### Describe the solution you'd like
This functionalidty allows select a BOM for Production Batch and set it for Production, now the BOM used always is the default BOM for product and sometimes is very necessary select a specific BOM for a Batch Order.
#### Describe alternatives you've considered
None

#### Additional context
This is a little change but is a great benefit

PR #2774, opened 2019-09-09
Add support to BOM for Production Batch Order and production for one

yamelsenih:
This pull request allows select a BOM of Product for Production batch and production.

Fixes: #2773 

Best regard

### DIFF
--- a/base/src/org/compiere/model/I_M_Production.java
+++ b/base/src/org/compiere/model/I_M_Production.java
@@ -398,6 +398,21 @@ public interface I_M_Production
 	  */
 	public boolean isPosted();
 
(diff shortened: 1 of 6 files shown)


marcalwestf:
hi @yamelsenih  a nice and very useful enhancement.
I tried it out and works.

Only to make it round, it would be desirable to add the field _BOM & Formula_ to the  tabs _Production Batch_ as well as _Production_ of window _Planned Manufacturing Order_.

Right now they are missing, as it can be seen on the following images where the BOM is missing:
Selection_112

Selection_113

yamelsenih:
Hi @marcalwestf this is changed yet

Thanks

PR #2777, opened 2019-09-09
Add report View for Storage Detail:

yamelsenih:
This pull request just add two Report Views for Storage detail report:
- Only Stock: Show All Storage when it have Stock
- Only Available: Show All storage when it have quantity available


yamelsenih:
Hi @marcalwestf this change was done

Issue #2437, opened 2019-03-21
Next Line in "Description Field" by pressing "Shift + Intro" 


niclopup:
By the correction made in the #2329  the function of allow the correct movements in a "Description Field" it ok but only with the Arrows of the Keyboard.

The problem is that when you want to go down to the next Line in a Field of "Description" when the window is in "Single Record" View, the cursor get lost. 

I couldnt find the reason but when you get to a Register from the "Recent Documents" sometimes the problem is solved, but when you get the same register searching it from the window from the Manú it happens again the Bug. 

I add a Video to make it easier to see the problem. 

https://youtu.be/BmvWmV-PNFg

PR #2778, opened 2019-09-09
Bugfix/#2437 next line with shift intro

rcapecce:
Fixed the enter in a multiline field.
Fixes: #2437

### DIFF
--- a/zkwebui/WEB-INF/src/org/adempiere/webui/component/GridPanel.java
+++ b/zkwebui/WEB-INF/src/org/adempiere/webui/component/GridPanel.java
@@ -79,8 +79,8 @@ public class GridPanel extends Borderlayout implements EventListener
 	private static final int KEYBOARD_KEY_RETURN = 13;
 	private static final int MOVE_SIZE = 7;
 
-	public static final String		CNTRL_KEYS				= "#f5#del^d^s#enter$#enter";
-	private static final String		KEYS_MOVE			= "#pgup#pgdn#end#home#up#down#left#right";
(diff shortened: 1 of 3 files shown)


marcalwestf:
hi @rcapecce  this is an annoying bug.

Does it solves only the problem of hitting the _ENTER_ key when editing fields like _Description_ or _Help_?  Or does it something additionally?

I wonder how to leave editing fields like _Description_ or _Help_ and get the focus to the next field?

Thanks,.

marcalwestf:
@rcapecce  thanks Raul, I will test accordingly.

marcalwestf:
@rcapecce  Raul , I am about to merge your changes.

I have just one question when editing on the grid ("multirecord").

The following key combinations work:
- CTL + LEFT ARROW (=go to the beginning of the text)
- CTL + RIGHT ARROW (=go to the end of the text)
- SHIFT + LEFT ARROW (=select character-wise to the left)
- SHIFT + RIGHT ARROW (=select character-wise to the right)

I wonder which combination would yield a "**move one character to the left**" or  "**move one character to the right**"  when editing on the grid ("multirecord"). 

The arrows are already reserved for moving to the next field to the right or to the left when editing on the grid ("multirecord"), so they cannot be used. On monorecord, the keys LEFT ARROW and RIGHT ARROW work correctly.

Thanks.

rcapecce:
That is correct, with Ctrl + arrow steps between words and with shift move selecting the characters, in multirecord mode only the single left right arrow is catched for the grid navigation.

Issue #2802, opened 2019-09-14
[Bug Report] Set Document No from Payment Selection Ckeck


yamelsenih:
## Bug report
When a Payment Selection is completed, the payment selection check is generated with a number for check no. the problem is that if a payment selection is exported the document no for export payment is getted from payment selection check and all payments generated have a document number distinct of exported number.

When a user load a bank statement not exists reference between Payment number and bank statement reference, see a example:

#### Steps to reproduce
1. Create a Payment selection
![Screenshot_20190914_174711](https://user-images.githubusercontent.com/2333092/64913984-f2d86b80-d717-11e9-8405-6118699261ae.png)
2. Complete it
3. Review payment selection check tab
4. Export payment selection
![Screenshot_20190914_174726](https://user-images.githubusercontent.com/2333092/64913989-fa981000-d717-11e9-8ab9-05d850ac1b02.png)

Note that if you follow the step the payment selection check will be generated with a sequence like this:

Payment Selection Check **Document No**: **10**
Payment Generated **Document No**: **PCxPN-1000032**

When is loaded bank statement the reference for match is **10** but ADempiere payment reference is **PCxPN-1000032**

#### Other relevant information
- Your OS: Debian 9
- Java version: 8
- ADempiere version: 3.9.2
- Browser Version [e.g. 22]
- DB Provider: PostgreSQL
- DB version:

#### Additional context
This is a little change but a big benefits

PR #2803, opened 2019-09-14
Set document no for payment from payment selection

yamelsenih:
Set document no from payment selection check
Fixes: #2802


e-Evolution:
@yamelsenih only I think that not need this code line

https://github.com/adempiere/adempiere/pull/2803/files#diff-7d783e1e7e5b4c8c15e896ef6aa6f7e8R310

Because the sys config had a flag to apply this behavior

Issue #2823, opened 2019-09-24
[Feature Request] Use DocType POS Order when using configuration "Cash_As_Payment"


SusanneCalderon:
## Feature request
For the configuration Cash_as_Payment = 'N' the document type POS Order generates a register in C_CashLine. 
When the configuration Cash_As_Payment is Yes, instead of Cash and CashLine a bank account type Cash is used for the cash Payments.
To use the Doctype Pos Order under this configuration a payment must be generated using a default bank account with bank type cash.

#### Describe the solution you'd like
Solution:
Define a bank Account of a bank with bank type cash a default bank account.
Completing a POS Order the payment is created using this default bank account.
```

private String payCashWithCashAsPayment() {
		int posId = Env.getContextAsInt(getCtx(),Env.POS_ID);
		MPayment paymentCash = new MPayment(getCtx(), 0 ,  get_TrxName());
		if (posId > 0) {
			MPOS pos = MPOS.get(getCtx(), posId);
			paymentCash.setC_BankAccount_ID(pos.getC_BankAccount_ID());
		} else {
			MBankAccount bankAccount = MBankAccount.getDefault(getCtx(), getAD_Org_ID(), X_C_Bank.BANKTYPE_CashJournal);
			if (bankAccount == null) {
				processMsg = Msg.getMsg(getCtx(), "No Default Cash defined for Organization");
				return DOCSTATUS_Invalid;
			}
			paymentCash.setC_BankAccount_ID(bankAccount.getC_BankAccount_ID());
		}

		paymentCash.setC_DocType_ID(true);
        String value = DB.getDocumentNo(paymentCash.getC_DocType_ID(),get_TrxName(), false,  paymentCash);
        paymentCash.setDocumentNo(value);
        paymentCash.setDateAcct(getDateAcct());
        paymentCash.setDateTrx(getDateInvoiced());
        paymentCash.setTenderType(MPayment.TENDERTYPE_Cash);
        paymentCash.setDescription(getDescription());
        paymentCash.setC_BPartner_ID (getC_BPartner_ID());
        paymentCash.setC_Currency_ID(getC_Currency_ID());
        paymentCash.setPayAmt(getGrandTotal());
        paymentCash.setOverUnderAmt(Env.ZERO);
        paymentCash.setC_Invoice_ID(getC_Invoice_ID());
		paymentCash.saveEx();
		if (!paymentCash.processIt(X_C_Payment.DOCACTION_Complete))
[...]

PR #2824, opened 2019-09-24
adempiere  #2823  POS Order when using configuration "Cash_As_Payment"

SusanneCalderon:
fixes https://github.com/adempiere/adempiere/issues/2823


e-Evolution:
@SusanneCalderon , can please remove the unnecessary import org.eevolution.model.I_HR_Concept_Acct

PR #2830, opened 2019-09-27
add support to many delivered quantity from same outbound order

yamelsenih:
This pull request allows generate many delivery from same outbound order,

Best regard.


e-Evolution:
solve with this pull request #2862

PR #2831, opened 2019-09-27
Fixed error with process parameters for process builder

yamelsenih:
Add support to process parameter To on process builder.

If you use ProcessBuilder and add a parameter with **From** and **To**

### DIFF
--- a/base/src/org/compiere/model/MPInstance.java
+++ b/base/src/org/compiere/model/MPInstance.java
@@ -21,8 +21,10 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.Timestamp;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.Properties;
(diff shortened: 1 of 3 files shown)


marcalwestf:
a question @yamelsenih : is it right that all parameters have to be tested because the whole method was changed?
Thanks

yamelsenih:
Hi @marcalwestf this change apply for all functionalities that are using ProcessBuilder, currently the range parameters is not implemented for this functionality.

e-Evolution:
@yamelsenih please give support to range using a polymorphous method, this way the build pattern DSL is rightly implemented 

`ProcessInfo processInfo = ProcessBuilder.create(getCtx())
.process(GenerateInvoiceInOutBound.getProcessId())
.withParameter(GenerateInvoiceInOutBound.DATEINVOICED, 
TimeUtil.getDay(2019, 1, 1),
TimeUtil.getDay(2019, 1, 31))
.execute(get_TrxName());`

yamelsenih:
hi @e-Evolution this change was done

PR #2847, opened 2019-10-03
Fixed error on Accounting of Shipment/Receipt

carlosaparadam:
When the accounting of original receipt or shipment is reset don't accounted the original document and reversal document

Before fixed:

Original Document
image

Reversal Document
image

After fixed:

Original Document
image

Reversal Document
image

### DIFF
--- a/base/src/org/compiere/acct/Doc_InOut.java
+++ b/base/src/org/compiere/acct/Doc_InOut.java
@@ -95,7 +95,8 @@ public class Doc_InOut extends Doc
 			
 			DocLine docLine = new DocLine (line, this);
 			BigDecimal Qty = line.getMovementQty();
-			docLine.setReversalLine_ID(line.getReversalLine_ID());		
+			if (!line.isReversalParent())
+				docLine.setReversalLine_ID(line.getReversalLine_ID());		
 			docLine.setQty (Qty, !getDocumentType().equals(DOCTYPE_MatShipment));    //  sets Trx and Storage Qty
 			
 			//Define if Outside Processing 


e-Evolution:
@carlosaparadam , @yamelsenih  Hi Carlos , Yamel A behavior that must be considered in the case of documents that affect the inventory, which the reverse require to perform based on the original accounting entry, this is because in some occasions from one accounting period to another they could change the configuration of a accounting account and the effect you require is for the accounting to cancel the original accounting entry.

This reason the document type flag Is Reversed with original Accounting

marcalwestf:
@carlosaparadam  I cannot reproduce this error.
When executing a _Reverse-Correct it_ action on a completed _Material Receipt_ document, I still get accounting facts from both documents:
Selection_164
Selection_165
Selection_166
Selection_168


or perhaps it is another action/process which should be executed?

carlosaparadam:
hi @e-Evolution , @marcalwestf , i understand the beavior of reversal document that you comment @e-Evolution but the problem is that after reversed the documents and reset accounting when you need re-accounting the reversed documents don't accounting, the original document search the reversed and the reversed the original and none is finded why the accounting for the two was reset. @marcalwestf if you test resetting accounting and re-accounting you see the bug.

Regards

SusanneCalderon:
hi @carlosaparadam , hi @e-Evolution 
I have tested it and without the flag _isReversedWithOriginalAcct_ it works correctly. 
@e-Evolution, for the case _isReversedWithOriginalAcct_ I do not see where a new accounting configuration is used, as updateReverseLine in Doc_InOut copies the original accounting too.
When I activate _isReversedWithOriginalAcct_  the accounting of the reversal document is null.

SusanneCalderon:
@carlosaparadam , @e-Evolution 
with the code of the head revision the reversal applies correctly for the case of _isReversedWithOriginalAcct_

marcalwestf:
@carlosaparadam  it seems that you should consider in your implementation the case @e-Evolution mention and @SusanneCalderon reproduced.

With your implementation, the error is solved, but it breaks old functionality.

carlosaparadam:
hi @marcalwestf , @SusanneCalderon , @e-Evolution thanks for their comments, i will testing again for fix the bug without break the current functionality.

Regards

carlosaparadam:
hi @marcalwestf , @SusanneCalderon, this gif show how to reproduce the error
Accounting Receipt

marcalwestf:
@carlosaparadam  reviewing this long-waiting PR, I see that you uploaded a gif file displaying the bug. We can reproduce the bug. With your changes the bug is fixed and I would accept it.

Reading the comments the issue here -as I understand it- is that @e-Evolution asks for a _backwards compatibility_  for the case that (quoting Victor):

> ... in some occasions from one accounting period to another they could change the configuration of a accounting account and the effect you require is for the accounting to cancel the original accounting entry....

If I understand Victor well, he means **if** a company changes the accounting configuration on the fly **and** there is a need for reversal, the situation Victor mentions occurs.
In our experience, companies usually do not change accounting configurations that often. 

Thinking again about it Victor, it happens the following: there is a bug, also in "old" implementations. So I wonder what backward compatibility should we stick to with a bug? 

Or do you perhaps mean that the present bug fixing should consider an "old" configuration and reverse accordingly? I would not know how this would be accomplished.

carlosaparadam:
Hi @marcalwestf the problem is resolved with isReversedWithOriginalAcct but only for reversed document, but the original document stays searching the reversed document and if this document don't have posted, display the error.

Regards

marcalwestf (approved):
In code is the only change the following:
Selection_957

This is sensible, and corrects wrong accounting.
In short: **ReversalLine_ID** is set only if the line belongs to the document used to revert.

Before this change **ReversalLine_ID** was set  in both the document used to revert and in the reverted document, which could lead to wrong accounting in the latter.

Issue #1439, opened 2017-10-31
Reversed based on the accounting of the original document


e-Evolution:
Is necessary to generate the accounting for a document reversal based on the accounting of the original document.

Currently, when you reverse any document your accounting facts are generated based on the accounting engine and current standard configuration, the issue with this approach is if the accounting setting change, then the accounting effects are different than original document and not get the goal of canceling the original accounting facts.

To solve this issue, a new column "IsReverseWithOriginalAccounting" will be added to Document Type and defined by default as "No", this allows  to define the behaving by document type

e-Evolution:
![captura de pantalla 2017-10-31 a la s 14 12 06](https://user-images.githubusercontent.com/786968/32246728-8ad93452-be45-11e7-94f1-369e10797df9.png)

Issue #1239, opened 2017-07-10
Error for reverse document using Allow Negative Posting as 'N'


e-Evolution:
Current when a document is reverted and the Allow Negative Posting  is No into accounting schema only the source amount are change absolute value , but the account amount not is convert to absolute values it cause the inconsistence data

PR #2849, opened 2019-10-03
Fixed Error on Accounting of Match Invoice

carlosaparadam:
Currently, at reversing invoice, credit note, receipt or vendor return , this generate a new invoice matching  associated to reverse document, but the accounting for this document is incorrect.

Original Document:
image

Reversed Document:
image

Accounting of documents:
image

Accounting of documents after fixed:
image

### DIFF
--- a/base/src/org/compiere/acct/Doc_MatchInv.java
+++ b/base/src/org/compiere/acct/Doc_MatchInv.java
@@ -19,6 +19,7 @@ package org.compiere.acct;
 import java.math.BigDecimal;
 import java.sql.ResultSet;
 import java.util.ArrayList;
+import java.util.Optional;
 
 import org.compiere.model.MAccount;
 import org.compiere.model.MAcctSchema;
@@ -69,6 +70,9 @@ public class Doc_MatchInv extends Doc
 	private MInOutLine		m_receiptLine = null;
 	
 	private ProductCost		m_pc = null;
+	
+	Optional<MMatchInv> matchInv = Optional.empty();
+	Optional<MMatchInv> matchInvOrig = Optional.empty();
 
 	/** Commitments			*/
 //	private DocLine[]		m_commitments = null;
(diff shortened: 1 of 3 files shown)


e-Evolution:
@carlosaparadam Hi Carlos 

The accounting entries for a reversal must not be of absolute value so as not to inflate or distort the debts and credits.

Thus the reversals must be equal to the original accounting entry but with a negative value, as any other reverse. 

Do you check is reversal with original Accounting?

image

carlosaparadam:
hi @e-Evolution , i try it and it is not working, i don't see when the field IsReversedWithOriginalAcct of document type is supported, in the change i dont use absolute value, i find the original accounting and use the method updateReverseLine from FactLine class.

see this example: 
Document Type Setting
image

Original Document
image

Reversal Document
image

Accounting Comparation
image

e-Evolution:
@carlosaparadam this account working but I new changes was introduced Cost Engine , you need check 
 
@SusanneCalderon test here and  no exist this issue 

https://github.com/adempiere/adempiere/issues/1439
https://github.com/adempiere/adempiere/issues/1348
https://github.com/adempiere/adempiere/issues/1239

carlosaparadam:
ok @e-Evolution, i will test and comment again

e-Evolution:
carlosaparadam the Cost Engine is very complex and sencible to any change 

So any change need complice with requieremets
- The reverse accounting should be equal original, but with opposite sign
- the cost detail transaction should be rigth for costing method , the inventory value should ve right
- the quantity in any accconting fact should be equal that original quantiy,  but with opposite sign

Please consider these important points.

carlosaparadam:
@e-Evolution, i tried again with IsReversedWithOriginalAcct and you have reason, but for it work the table sould be have the column of document type, i will close this pull request and open a new for creating document type for this table and setting a document type by default.

thanks for your help and suggest, i will hope for your comments.

Regards

Issue #2850, opened 2019-10-04
The document type for Freight Order show all document type


e-Evolution:
#Bug report
When a Freight Order is create the document type show all types 

#### Steps to reproduce
1.- Create a Freight Order
2.- Select document type 

#### Other relevant information
- Your OS: MacOS
- Java version: 1.8
- ADempiere version: 3.9.2
- Browser safari
- DB Provider: PG
- DB version: 11.4

PR #2851, opened 2019-10-04
The document type for Freight Order show all document type

e-Evolution:
The document type for Freight Order show all document type #2850


marcalwestf:
hi, the Release No is missing:
Selection_160

Issue #2880, opened 2019-10-16
[Bug Report]2880 Not defined process


SusanneCalderon:
## Bug report
the process _GenerateRequest_ is not generated in the code

#### Steps to reproduce
Try to run process GenerateRequest in Request Management

#### Screenshot or Gif
Process in AD

![imagen](https://user-images.githubusercontent.com/15349639/66909674-61843f80-efca-11e9-8bec-a6f45a4d82bb.png)

Running the Process

![imagen](https://user-images.githubusercontent.com/15349639/66909714-77920000-efca-11e9-87e6-9f08aeeafb0a.png)

PR #2881, opened 2019-10-16
adempiere #2880 [Bug Report]2880 Not defined process

SusanneCalderon:
fixes https://github.com/adempiere/adempiere/issues/2880


yamelsenih:
Hi @SusanneCalderon maybe is better if the process is created instead

Best regard

marcalwestf:
@yamelsenih  I agree with you that it would be better to create the process.

Meanwhile the process is not developed, the process should not be called.

yamelsenih:
Ok for it

PR #2883, opened 2019-10-21
Add transaction name for business partner on beforeSave method

yamelsenih:
Currently when a business partner is saved is called a BP group from it but transaction name is not used for it.

### DIFF
--- a/base/src/org/compiere/model/MBPGroup.java
+++ b/base/src/org/compiere/model/MBPGroup.java
@@ -41,18 +41,28 @@ public class MBPGroup extends X_C_BP_Group
 	private static final long serialVersionUID = 3037428352124938328L;
 
 	/**
-	 * 	Get MBPGroup from Cache
+	 * Get MBPGroup from Cache
+	 * @param ctx
+	 * @param C_BP_Group_ID
+	 * @return
+	 */
+	public static MBPGroup get (Properties ctx, int C_BP_Group_ID) {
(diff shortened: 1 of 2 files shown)


e-Evolution:
@yamelsenih 

There is a problem in the objects that are cached with a specific transaction, since the object that is created is included in the cache with the transaction that may be closed after adding the object to the cache, that is why it is advisable to use the cache  only for reading objects rather than writing.

In this case what happens if the partner group that was added to the cache is closed or terminated but in the cache is it storing an object with a closed transaction?

yamelsenih:
Hi Victor the problem is when a business partner and group is created own same transaction then finding Business partner group not found because is used null instead a bp transaction

e-Evolution:
@yamelsenih If this is indeed why you should not use the MBPGroup.get method (Properties ctx, int C_BP_Group_ID)

if you add a partner group with a transaction other than null that element to the chache can cause problems when trying to save something in the group because it can say that the transaction was already closed, that is why the elements that are stored in the cache are No transaction or null transaction.

Mi recomendación is when need to create Business Partner group using same transaction use

 `new MBPGroup(getCtx() , getC_BP_Group_ID(), get_TrxName()); `

yamelsenih:
Hi @e-Evolution I added this change here https://github.com/adempiere/adempiere/pull/2883/files#diff-b9ae85fc058442262330642693cc12d0R915

The problem is transaction name is unsupported for BG Group

PR #2892, opened 2019-11-13
Add consolidate document flag for smart browser "Generate Invoice from

yamelsenih:
This PR allows generate a invoice from many Sales Orders from same business partner, note that is added a new flag for consolidate all document.

Is nice for customers with invoices with many orders. See pic:
Screenshot_20191112_214041


marcalwestf:
thanks @yamelsenih  for this nice enhancement.

We'll test as soon as we can.

PR #2915, opened 2019-12-30
#2914 - During rollback of a Migration Step, force delete of an entity type.

### LINKED ISSUE #2914: Migration rollback of an entity does not delete the entity.
When rolling back a migration step that inserts a new entity into the application dictionary, the entity is not deleted.   Entities can only be deleted if forced.

mckayERP:
Fixes #2914 

During rollback of a migration, its not enough to use deleteEx with force set to true.  For Entity Types, the flag .setIsDeleteForced must be set to true prior to the delete.


e-Evolution:
@mckayERP , I think that us a good enhanced , reviewing I find that IsDeleteForced is a new column but no found in current model , How was added this column , or exist a modification in your PO class ?

mckayERP:
Hi Victor,

That flag is just a field in the MEntity.java, not a column.   It was added four years ago in #168 to handle deletes of Entity types which, generally, can't be deleted, even when forced.  The flag provides a way to bypass this restriction.

Mike

yamelsenih:
Hi @mckayERP @e-Evolution I approve this pull request.

Issue #2917, opened 2019-12-30
Add support to the WMS to support multiple exit strategy definitions


e-Evolution:
## Feature request
Add support to the WMS to support multiple exit strategy definitions, considering the outbound location.

#### Describe the solution you'd like

Now is possible to define multiple definitions by establishing application order by sequence and priority

PR #2918, opened 2019-12-30
#2917 Add support to the WMS to support multiple exit strategy defini…

e-Evolution:
…tions https://github.com/adempiere/adempiere/issues/2917


yamelsenih:
Fixes: #2917

Issue #476, opened 2016-08-04
Use preferred method of converting double to BigDecimal in Matching Invoices


mckayERP:
The Match PO-Invoice-Receipt window calculates the quantity to match as a double variable.  This is converted to a BigDecimal using the code new BigDecimal(qty).  This conversion is unpredictable and can lead to strange behaviour.  For example, 

new BigDecimal((double) 24.41) =24.410000000000000142108547152020037174224853515625

This small difference causes problems when trying to post the matched invoice as the reversal entry is found by searching for the invoice with the unpredictable quantity and it is not found, leading the accounting engine to guess that the invoice is not posted yet.

Tested on postgres.adempiere.io.  See Invoice 10000018, MR 10000516, and Match Invoice 10000001.  All the quantities match (0.1 tons of Phosphorous) but the Match Invoice can't be posted.

The java docs recommend to convert double variables to BigDecimals by using the string representation.  This conversion is exact and is performed by the static call BigDecimal.valueOf(double).  The change affects Match.java, VMatch.java and WMatch.java.

yamelsenih:
Hi @mckayERP i think that is better if you use the precision and just BigDecimal instead Double, like the [Allocation class](https://github.com/adempiere/adempiere/blob/master/client/src/org/compiere/apps/form/Allocation.java#L432) it allow handle big numbers.

Best regard

yamelsenih:
Hi, any update for it?

carlosaparadam:
solved on pull request #2931

PR #2931, opened 2020-01-07
Changing from double to BigDecimal for major precision on Matching

carlosaparadam:
On matching window for invoice or order and receipt and the quantities for both documents have decimals, the document of matching is register with a different quantity by decimals.

for fix the problem was changed the quantity field from double to BigDecimal.

See this example:
Invoice document with quantity 23.65 and receipt with 36.87

image

After create the match document, the invoice remains pending for assigned.

image

On the matched invoice window appear with quantity assigned 23.65

image

The query sql returns:

image

Matched invoice, after the change

image

After create the match document, the invoice is not pending for assigned.

image

The query sql returns:

image


mckayERP:
Fixes #476.  Thanks.

PR #2935, opened 2020-01-09
Fixed error with duplicate key for inventory import

yamelsenih:
The problem: if you try import a physical inventory from ADempiere importer currently not exists a validation for duplicate key before import then the error throw when physical inventory line is saved but the user can't see the error, a console log from debugger can show the problem but import inventory line don't have error:

```Shell
org.adempiere.exceptions.AdempiereException: ERROR: duplicate key value violates unique constraint "m_inventoryline_productlocattr"
  Detail: Key (m_inventory_id, m_locator_id, m_product_id, m_attributesetinstance_id)=(1000554, 1000048, 1000401, 0) already exists.
``` 

This change allows see the error line by line and validate before import, is very beautiful for end user

See a gif after fix it:
Current ogv

### DIFF
--- a/base/src/org/compiere/process/ImportInventory.java
+++ b/base/src/org/compiere/process/ImportInventory.java
@@ -30,6 +30,7 @@ import org.compiere.model.MInventoryLine;
 import org.compiere.model.MProduct;
 import org.compiere.model.X_I_Inventory;
 import org.compiere.util.DB;
+import org.compiere.util.Msg;
 import org.compiere.util.TimeUtil;
 
 /**
@@ -37,6 +38,7 @@ import org.compiere.util.TimeUtil;
  *
  * 	@author 	Jorg Janke
  * 	@version 	$Id: ImportInventory.java,v 1.2 2006/07/30 00:51:01 jjanke Exp $
+ * 	@author Yamel Senih, ysenih@erpya.com, ERPCyA http://www.erpya.com
  */
 public class ImportInventory extends SvrProcess
 {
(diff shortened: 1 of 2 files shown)


marcalwestf:
starting test...

marcalwestf:
@yamelsenih just thinking aloud:

As defined in _m_inventoryline_:
```
CREATE UNIQUE INDEX m_inventoryline_productlocattr
  ON m_inventoryline
  USING btree
  (m_inventory_id, m_locator_id, m_product_id, m_attributesetinstance_id);
```

It would mean that the combination (_m_inventory_id, m_locator_id, m_product_id, m_attributesetinstance_id_) has to be unique in every inventory, thus not allowing to count separately.

If we picture how inventory is done, the people in charge do not have all products sorted by ASIs so that they take the line at once, but the products are most probably scattered all over the place.

So in order to comply with the unique index and at the same time making the process efficient, I inagine a sort of callout that instead of trying to add a repeated line, it finds the repetition and adds the quantity to the existing line.

This would be a separate issue, and has nothing to do with the error here described, only I wanted to document it.

marcalwestf:
...continuing the test...

marcalwestf:
Result of trying to import a duplicated Inventory:
PR_Duplication_InventoryImport_02

The count specified in the error message is from a product already imported. This should not be an error.

Please @yamelsenih exclude entries already processed and imported.

marcalwestf:
Hi @yamelsenih with the recent changes, the records get the error message, even if there is already an imported line with the same data.

What now occurs is that the count is wrong. This means, that if there is an already imported line and two repetitions of the same line, not yet imported, the Qty=3 instead of 2.

Moreover, I'd suggest to better replace the @Qty@ message with another one like "number of lines", "number of occurrences" etc., because a user maybe think that the quantity meant is the one of the inventory line, instead of the quantity of error occurrences.

yamelsenih:
Hi Mario let me test it and get back to you

yamelsenih:
Hi @marcalwestf the message just show all lines counted with the same **(Product, Locator and Attribute set instance)**.

A physical inventory don't allow the same combination for more than one line **(Product, Locator and Attribute set instance)**

PR #2932, opened 2020-01-07
Add Sales Order line reference on create Receipt from Invoice

carlosaparadam:
When the receipt is create form an invoice allocated an sales order the sales order line is not referenced.

image

PR #2936, opened 2020-01-09
Little change for add link to order line when a Material Receipt is

yamelsenih:
Just a little change for add order line reference to Material receipt Line when is created from Invoice

Best regard


yamelsenih:
Duplicated here #2932

Issue #2948, opened 2020-01-16
The accounting distribution does not work until after loading the accounting distribution in the cache.


e-Evolution:
## Bug report
The accounting distribution does not work until after loading the accounting distribution in the cache.
#### Steps to reproduce
1. Validate that accounting Is immediate 
2. Create a new Charge for Account Salary
3. Create Invoice using charge 
4. Complete the invoice and check account and distribution this not was generated.
5. Repost the document in this moment the distribution is generate.

#### Other relevant information
- Your OS: MacOS 
- Java version: 1.8
- ADempiere version: 3.9.3
- Browser safari
- DB Provider: PG
- DB version: 12.1

PR #2949, opened 2020-01-16
#2948 The accounting distribution does not work until after loading t…

e-Evolution:
…he accounting distribution in the cache. https://github.com/adempiere/adempiere/issues/2948


marcalwestf:
starting test..

PR #2951, opened 2020-01-22
Fix ambiguous column and optimize previous balance query in Financial Statement Report

carlosaparadam:
Currently, at generating Financial Statement report and filter by Organization, generate an error by a ambiguous column.

See Parameters filter:
Screenshot_20200122_151821

Errors with PostgreSQL database:
Screenshot_20200122_152316

Errors with Oracle database:
Screenshot_20200122_152453

Also, the query for Beginning Balance is generating response times very high, by the subqueries in the select clause.

see current method:
Screenshot_20200122_152532

see the proposed method:
Screenshot_20200122_152822

### DIFF
--- a/base/src/org/compiere/report/FinStatement.java
+++ b/base/src/org/compiere/report/FinStatement.java
@@ -200,6 +200,8 @@ public class FinStatement extends FinStatementAbstract
 		if (getAccountType() != null && !getAccountType().isEmpty())
 			sb.append(" AND  ev.AccountType = '").append(getAccountType()).append("'");
 
+		//Client Filter 
+		sb.append(" AND ev.AD_Client_ID = ").append(getAD_Client_ID());
 		
 		int no = DB.executeUpdate(sb.toString(), get_TrxName());
 		log.fine("#" + no + " (Account_ID=" + getAccountId() + ")");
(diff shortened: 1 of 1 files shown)


SusanneCalderon:
Hello @carlosaparadam 
I tested the process in an instalation with 4 adempiere Clients. The result showed all accounts from all clients, see the result of this sql:

> select distinct ev.ad_Client_ID,ad_Pinstance_ID
> from
> adempiere.T_Reportstatement t
> INNER JOIN adempiere.c_elementvalue ev on t.account_ID=ev.c_elementvalue_ID
> where ad_Pinstance_ID=1037216

Result : AD_Client_ID      AD_Pinstance_ID
            1000023	       1037216
            1000022	       1037216
            1000000	       1037216
            11	               1037216

carlosaparadam:
Hi @SusanneCalderon, thanks for review. I fixed this bug now.

Screenshot_20200205_112505

SusanneCalderon:
Hi @carlosaparadam 
I tested it, too. Ready to merge.

e-Evolution:
@carlosaparadam Hi Carlos , @marcalwestf Mario , @SusanneCalderon Susanne!

This patch introduces multiple problems since it poorly reflects the balance in accounts that have debits and credits, the credits appear in negative which is not correct, on the other hand if you set the account type parameter the report is not executed, the proposed query to optimize complicates the use of filters and does not add correct optimization, since to optimize indexes, not subqueries, are created.

Please be more careful in your tests because this causes the account statements to come out wrong starting from this version.

I'm going to revert and implement another approach.

carlosaparadam:
Hi @e-Evolution Victor, this pull request only fix an sql error for ambiguous column and move the subqueries that were in the select to the from to improve the query, the error has been introduced here https://github.com/adempiere/adempiere/pull/2050 which is where you start using the **AcctBalance** function that works according to the nature of the account

e-Evolution:
> Hi @e-Evolution Victor, this pull request only fix an sql error for ambiguous column and move the subqueries that were in the select to the from to improve the query, the error has been introduced here #2050 which is where you start using the **AcctBalance** function that works according to the nature of the account

Hi @carlosaparadam  Carlos , @SusanneCalderon Susanne , @marcalwestf  Mario!

Although the company id was added which is correct and the approach to optimizing the query is appropriate, the problem is that the SQL that calculates the credits and the final balance is incorrect, here you not only changed the query but the logic. on the other hand, when adding this query, the accouttype remained as an ambiguous column.

What we have to improve are our tests and these have to be more exhaustive and not so simple, to cover the majority of cases, here I send the correction https://github.com/adempiere/adempiere/pull/4298

kind regards
Victor

Issue #2973, opened 2020-02-03
Add support to attach file result of packout


rcapecce:
## Feature request
When the packout is generated in the harddisk, now it is attached to the current record.

![image](https://user-images.githubusercontent.com/14218144/73624732-b9e65800-4620-11ea-9c76-e255a2d28d0e.png)

When the button "Export Package" is executed, and the result zip pack file is generated, it is attached to the current process.
![image](https://user-images.githubusercontent.com/14218144/73624745-c7034700-4620-11ea-91a3-46cb79fe027c.png)

Issue #2135, opened 2018-11-09
Whitespace for git changes


yamelsenih:
Hi everybody, I note that some files are changed completely when is pushed a change into ADempiere.

A example of it is: https://github.com/adempiere/adempiere/pull/2134/commits/ff65981e40cda20780cdb4c327c8666ca2bc39f9

The problem is that each git user can have distinct types [CR LF](https://en.wikipedia.org/wiki/Newline), then the order for reproduce this error is:

1. User 1 create a new file and User 1 have CR
2. User 2 add a change to file created by User 1 but User 2 have CRLF

I find a way for resolve it with a git command

<pre>
git config --global core.autocrlf false
</pre>

Best regard

PR #2974, opened 2020-02-03
#2973 - Add support to attach file result of packout

rcapecce:


marcalwestf:
hi @rcapecce it seems a nice feature to attach the result of packout.

On the other hand, perhaps there are implementations which prefer not to attach; for these cases an additional field "SaveResultToAttachment"  on "Export Package" tab (with default value=false) or similar would be worth considering.

The value==false would be for backward compatibility.

rcapecce:
That is correct, next I need add this flag to indicate if the package needs to attach.

yamelsenih:
Hi @rcapecce thanks for it, iseem that you hace a changes for all lines of Packout class:

- Have you different codification?
- see it: #2135

Screenshot_20200216_001641

rcapecce:
Hello @yamelsenih . I check the settings and tell you.

yamelsenih:
Hi @rcapecce maybe this PR can be rebuilding with a fresh changes from develop, can you make it for review? If is not possible just let me know for close meawhile

Issue #2975, opened 2020-02-03
[Bug Report] Rounded Price and Quantity for Order / Invoice


yamelsenih:
#### Previous data
Product information required:
- Product: 
  - Code: TP-001
  - Name Toilet Paper 9"
  - Product UOM: Each
  - Conversion: 1 Pack (4 Units) = 4 units

#### Steps to reproduce
1. Create a Order
2. Add new line
3. Quantity Entered (Unit of purchase): **5 Packs**
4. Quantity Ordered (Uom of Product): **20 Each**
5. Price Entered (Price on Pack): **$ 182,193.74**
6. Price Actual (Price on Each): **$ 45,548.435**

#### Expected behavior
- Total Line: **$ 910,968.70**
- Form used: Total Line = **Price Entered (Price on Pack)** * **Quantity Entered (Unit of purchase)** ~ **$ 182,193.74** * **5** = **$ 910,968.70**

#### Current bad result:

- Total Line: **$ 182,193.80**
- **Price Entered (Price on Pack)** / **Quantity Ordered (UOM of Product)**: **$ 9,109.687**
- **Rounded Unit price (ADempiere Internal calc)**: **$ 45,548.435** ~ **$ 45,548.44**
- **Form used**: **Total Line** = (Price Entered (**182,193.74**)  / Conversion Rate (**4**)) * Quantity Ordered (**20**) ~ **182,193.74** / **4** ~ [**45,548.435** ~ rounded is **45,548.44**] * **20** = **$ 182,193.80**

#### Other relevant information
- Your OS: Debian
- Java version: 1.8
- ADempiere version: 3.9.3
- Browser [e.g. chrome, safari]: Chrome
- Browser Version [e.g. 22]
- DB Provider: PostgreSQL
- DB version: 10x

marcalwestf:
hi @yamelsenih  I cannot reproduce the error.
I took Adempiere without your changes first.

Here what I did:
1.- Create product _TP-001_ with UOM _Each_ and add a conversion rule:
![Selection_335](https://user-images.githubusercontent.com/1789408/75103111-d3dddf80-55bb-11ea-92da-1727a1f15c6b.png)

2- Add the product to the Purchase price list:
![Selection_336](https://user-images.githubusercontent.com/1789408/75103122-fcfe7000-55bb-11ea-8bf2-5564af1c86cc.png)

3.- I created a purchase order for a vendor with a purchase price list:
[...]

PR #2976, opened 2020-02-03
#2975, Fixed error with round for AP / AR documents and Orders,

yamelsenih:
reference to issue: https://github.com/adempiere/adempiere/issues/2975

Fixes: #2975

### DIFF
--- a/base/src/org/compiere/model/CalloutInvoice.java
+++ b/base/src/org/compiere/model/CalloutInvoice.java
@@ -500,37 +500,37 @@ public class CalloutInvoice extends CalloutEngine
 			return "";
 
 	//	log.log(Level.WARNING,"amt - init");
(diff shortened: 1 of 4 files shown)


marcalwestf:
hi @yamelsenih thanks for the PR.

I would suggest that in cases like this, when changing code and at the same time doing code refactoring, it would be done in separate commits: the first commit containing only the refactoring and the second one the functional changes.

This way the comprehension of the changes and thus the testing would become much easier.
Thanks!

yamelsenih:
Hi @marcalwestf thanks, the escential chances are here:
https://github.com/adempiere/adempiere/pull/2976/files#diff-59a6439cfc79b6afd66cf7c207c1f50aR1120
https://github.com/adempiere/adempiere/pull/2976/files#diff-c4ef1715d6809c4761043c6533c98998R460
https://github.com/adempiere/adempiere/pull/2976/files#diff-62c4b1c3d15ca213b10c5b3688b8db80R368

Issue #2991, opened 2020-02-09
Building and Launching 3.9.3 in eclipse


homebeaver:
Hi, there are still some works on eclipse build and launch in 3.9.3+ 

- adjust eclipse .project and .classpath in some eclipse subprojects
- inter project dependences have to be defined in eclipse .classpath
- adapt .gitignore not to store .class .jar or .war in the repo, example `lib/ADInterface-1.0.war`
- adjust launchers

I have a PR - regards

PR #2992, opened 2020-02-09
#2991 eclipse 3.9.3+build and launch

homebeaver:
Issue #2991 : with little changes on eclipse .project and .classpath the build is OK and AD can be launched from eclipse workspace

regards


homebeaver:
Hi @trifonnt - you are a build & eclipse expert.
Would you like to review this PR?
tnkx EUGen

marcalwestf:
Dear Eugen, @homebeaver , I see a lot of changes.

As an Eclipse user, what use cases do you recommend to follow?
My question is aiming at not testinn in the blue, but with real-life cases.

Thanks!

PS/ If you could resolve the conflict, I could start testing

homebeaver:
Hi Mario @marcalwestf , the conflict was due to commit 4ce260e - yamelsenih committed 28 days ago - It is resolved now.

Here my eclipse use case:

1. you can start from scratch (with an empty eclipse workspace) and clone AD into this WS (it's resource and time-consuming)
1. or you can take an non empty git clone in an eclipse WS and delete all eclipse projects - not from disk - just the projects

1. then you checkout the `klst-de:3.9.3+build+dev` branch 
and import all the AD projects: >file>import>existing projects into ws
( I prefere to import all the nested projects base, client, ... etc 
but not the adempiere root project - this is because eclipse lists the projects twice )

1. the automatic build should start and build all projects

1. AND you can launch the ADempiere.launch which is located in base project 

grafik

regards EUGen

marcalwestf:
@homebeaver  thanks Eugen.
I will test ASAP.

marcalwestf:
... starting a test now

marcalwestf:
@homebeaver  could we arrange a meeting? The goal: to explain in detail what is exactly the changes and to really understand the functionality.

I also work on Eclipse and I think this would be useful for other Eclipse users.

I am in the timezone GMT - 6, you are most probably in GMT + 2 now.
We can coordinate date+time on Gitter.

Regards,
Mario

marcalwestf:
after a meeting with @homebeaver  Eugen, I at last understood the motivation of this Pull Request. 
This pull request concerns only Eclipse. Users of another GUI will **not** be affected.

First it has to be explained that Adempiere can be opened in Eclipse on three different ways: 

1.  As a single project
2.  As a bunch of different projects (e.g. base, client, zkwebui, etc.)
3.  As a combination of 1 and 2

Case 1 is useful when the developer has several Adempiere projects of different repositories and doesn't want to be confused by the subprojects. If the developer opens two Adempiere repositories as subprojects, it would be unmanageable.

Case 2 and 3 can be useful when the developer has only one single Adempiere project and perhaps one o few customization projects. Concentrating on one subproject makes it perhaps easier to manage the complexity of Adempiere.

Though all cases are justified, I have only used case 1 and ignored at all cases 2 and 3.

Either way, all cases ares accomplished by Eclipse with the files .classpath and .project either on the project root or on the sub projects  (base, client, zkwebui, etc.).

The problems Eugen tackled with this pull request concern the subprojects. So if you use only one single project, this pull request is of no interest to you.

The changes have to be done in such a way that the compilation or build for the "single project" case shouldn't be hampered by these changes.

marcalwestf:
@homebeaver  Eugen: you have to add the files  lib/ADInterface-1.0.war and lib/mysql-connector-java-5.1.13-bin.jar just this time!!

I know that in the pull request eliminates this dependency, so this would be the last time.

marcalwestf:
perhaps @eduardolopezuy  , @carlosaparadam  or anybody who uses Eclipse can also test.

homebeaver:
@marcalwestf 

by merging 'origin/develop' I resolved Conflicts:
	lib/ADInterface-1.0.war
	lib/mysql-connector-java-5.1.13-bin.jar

homebeaver:
Hi @marcalwestf - the conflict is because of

commit ae83afb942f23feccb9ad977b4654845846f517d
Committer: Yamel Senih <ysenih@erpya.com> 2020-06-05 19:16:50

I try to resolve

Issue #2995, opened 2020-02-09
Payselection used for Bank Transfers. The charge is not copied


SusanneCalderon:
**Using the document Payselection for a bank transfer**
1. Create bank transfer with document type bank transfer
2. In the line choose a charge, the amount and the receiving bank
3. Complete the pay selection
4. Generate the payment with Print/Export payselection
Result: the field charge in the generated payments is empty

**Document Payselection**
![Payselection1](https://user-images.githubusercontent.com/15349639/74095851-37e0ae00-4abc-11ea-98e7-d35273b81fad.png)

**Generated payment**
![Payselection2](https://user-images.githubusercontent.com/15349639/74095852-3911db00-4abc-11ea-89db-87721be58ca5.png)

Solution:
Copy the charge from the payselectionLine in the method MPayselectionCheck.confirmPrint()

```
receiptAccount.setC_BankAccount_ID(line.getC_BankAccountTo_ID());
							receiptAccount.setIsReceipt(!payment.isReceipt());
							receiptAccount.setC_DocType_ID(!payment.isReceipt());
							receiptAccount.setRelatedPayment_ID(payment.getC_Payment_ID());
							receiptAccount.setTenderType(MPayment.TENDERTYPE_DirectDeposit);
							receiptAccount.setC_Charge_ID(line.getC_Charge_ID());
							receiptAccount.saveEx();
							receiptAccount.processIt(DocAction.ACTION_Complete);
							receiptAccount.saveEx();
							payment.setRelatedPayment_ID(receiptAccount.getC_Payment_ID());
							payment.setC_Charge_ID(receiptAccount.getC_Charge_ID());
							payment.saveEx();

```

PR #2996, opened 2020-02-09
adempiere Payselection for Bank Transfers. Charge is not copied. PayAmt>Open Amount for invoices

SusanneCalderon:
Fixes https://github.com/adempiere/adempiere/issues/2995

### DIFF
--- a/base/src/org/compiere/model/MPaySelectionCheck.java
+++ b/base/src/org/compiere/model/MPaySelectionCheck.java
@@ -374,10 +374,13 @@ public final class MPaySelectionCheck extends X_C_PaySelectionCheck
 							receiptAccount.setC_DocType_ID(!payment.isReceipt());
 							receiptAccount.setRelatedPayment_ID(payment.getC_Payment_ID());
 							receiptAccount.setTenderType(MPayment.TENDERTYPE_DirectDeposit);
+							receiptAccount.setC_Charge_ID(line.getC_Charge_ID());
 							receiptAccount.saveEx();
 							receiptAccount.processIt(DocAction.ACTION_Complete);
 							receiptAccount.saveEx();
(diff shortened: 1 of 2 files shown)


e-Evolution:
@SusanneCalderon , My comments below :

about the #2995 I need to check with detail the code, but @yamelsenih created this functionality to transfer money between banks from the pay selection line, so I think that was testing before, Why should change, @yamelsenih do you think the need of charge?

about the #2995 I need to check with detail the code, but @yamelsenih created this functionality to transfer money between banks from the pay selection line, so I think that was testing before, Why should change, @yamelsenih do you think the need of a charge?

about #2997, I think that no is a good idea or practice allow pay an invoice for mayor amount that open amount, this can cause fraud or embezzlement for any company.

I have not any  business case where allow a payment mayor of open amount for an Invoice is ok, anyone would say it's a bug

yamelsenih:
Hi everyone

yamelsenih:
Hi @SusanneCalderon I suggest the follow change:

Keep the follow code or implement a flag for Payment selection document for allows pay document with over amounts:
```Java
if (getC_Invoice_ID() > 0 && getOpenAmt().subtract(getPayAmt()).subtract(getDiscountAmt()).signum() < 0 )
			throw new AdempiereException("@PayAmt@ > @C_Invoice_ID@ @Amount@ @OpenAmt@ ");
```
Also I agree with solucion for issue:#2995 but I think that is better if you create two pull request: one for bugfix of charge for bank transfer and other for feature of pay selectionj for AP document with over amount

SusanneCalderon:
@yamelsenih 
OK, I will generate two pull requests.

SusanneCalderon:
Hi @yamelsenih , @e-Evolution 
I reverted the modification in MPayselectionLine, now the commit only includes issue 2995.
@e-Evolution , the over amount is precisely what our newest customer wants.

Issue #3004, opened 2020-02-09
MHRProcess periodNo is set to context with name period


SusanneCalderon:
The name period in the scriptcontext is used twice.
```

scriptCtx.put("_Period", getHR_Period_ID());
    scriptCtx.put("_Payroll", getHR_Payroll_ID());
    scriptCtx.put("_Department", getHR_Department_ID());
    scriptCtx.put("_From", dateFrom);
    scriptCtx.put("_To", dateTo);
scriptCtx.put("_Period", payrollPeriod.getPeriodNo());
```

Issue #761, opened 2017-02-13
Add Payroll variables for Scripts


yamelsenih:
Currently, the payroll processing have the follow values:

- By Run:
  - **process**: MHRProcess Object
  - **_Process**: HR_Process_ID int
  - **_Period**: HR_Period_ID int
  - **_Payroll**: HR_Payroll_ID int
  - **_Department**: HR_Department_ID int
  - **_From**: StartDate (Payroll) Timestamp
  - **_To**: EndDate (Payroll) Timestamp
- By Employee run:
  - **_DateStart**: StartDate (Employee) Timestamp
  - **_DateEnd**: EndDate (Employee) Timestamp
  - **_Days**: Days between StartDate (Payroll) and EndDate (Payroll) int
  - **_C_BPartner_ID**: C_BPartner_ID int
  - **_HR_Employee_ID**: HR_Employee_ID int
- By Concept run
  - **_HR_Concept_ID**: HR_Concept_ID int
  - **_HR_Concept**: MHRConcept Object
  - **_HR_PayrollConcept_ID**:  HR_PayrollConcept_ID int

I propose the follow variables for help to user to get important values from Ctx:
- By Run:
  - **_HR_Payroll_Value**: Current Payroll Value String 
- By Employee run:
  - **_C_BPartner**: MBPartner Object
  - **_HR_Employee**: MHREmployee Object
  - **_HR_Employee_ValidFrom**: (Employee valid from): <pre>(_DateStart > _From? _DateStart: _From)</pre>
  - **_HR_Employee_ValidTo**: (Employee valid to): <pre>(_DateEnd < _To? _DateEnd: _To)</pre>
  - **_HR_Employee_Payroll_Value**: Payroll value of Employee String

e-Evolution:
Hi @yamelsenih 

Only have a question , I worried to add _C_BPartner because the instance in memory is very big , Do you test that this change not increase the procesing  payroll time?

yamelsenih:
Hi @e-Evolution i test it on a customer with over 400 employees and it worked nice, is better if you have the **_C_BPartner** like a object inside of Script rule engine instead of make a instance for each rule when it is necessary.

By each instance the server make a query to DB and instance a new java object.

PR #3005, opened 2020-02-09
MHRProcess periodNo is set to context with name period

SusanneCalderon:
Fixes https://github.com/adempiere/adempiere/issues/3004

### DIFF
--- a/org.eevolution.hr_and_payroll/src/main/java/base/org/eevolution/model/MHRProcess.java
+++ b/org.eevolution.hr_and_payroll/src/main/java/base/org/eevolution/model/MHRProcess.java
@@ -976,7 +976,9 @@ public class MHRProcess extends X_HR_Process implements DocAction , DocumentReve
 		scriptCtx.put("_Department", getHR_Department_ID());
 		scriptCtx.put("_From", dateFrom);
 		scriptCtx.put("_To", dateTo);
+		scriptCtx.put("_Period", payrollPeriod.getPeriodNo());
 		scriptCtx.put("_PeriodNo", payrollPeriod.getPeriodNo());
+		scriptCtx.put("_PeriodID", getHR_Period_ID());
 		scriptCtx.put("_HR_Payroll_Value", payroll.getValue());
 		//	Scope
 		scriptCtx.put("SCOPE_PROCESS", HRProcessActionMsg.SCOPE_PROCESS);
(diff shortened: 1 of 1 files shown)


e-Evolution:
@SusanneCalderon , we have here a problem, because from that Payroll functionality was created, we always used _Period, so if you change name, a lot the  payrolls can throw errors calculated payroll because  uses this syntaxis, so I think no is good idea change or rename, but if you  are your preference can create _PeriodNo as an alias too

SusanneCalderon:
@e-Evolution 
Ok, will do it.

SusanneCalderon:
Hi @e-Evolution .
So I switched back to the original code and for further use I created two new alias, _PeriodNo and _PeriodID.

org.eevolution.hr_and_payroll/src/main/java/base/org/eevolution/model/MHRProcess.java
```diff
@@ -977,6 +977,8 @@ private void createMovements() throws Exception
 		scriptCtx.put("_From", dateFrom);
 		scriptCtx.put("_To", dateTo);
 		scriptCtx.put("_Period", payrollPeriod.getPeriodNo());
+		scriptCtx.put("_PeriodNo", payrollPeriod.getPeriodNo());
```
yamelsenih: Hi @SusanneCalderon I suggest change it according with #761
The **_Period** is representative of **PeriodNo** and for old compatibility is better keep, but you can add the follow keys I think:

Example:
**_HR_Period_ID**: Period Id

The problem was that in beginning the keys was created without a good standard definition but I think that we should keep old implementation wihtout problems

SusanneCalderon: Hi @yamelsenih 
Theese are the keys now for period values:

		scriptCtx.put("_Period", payrollPeriod.getPeriodNo());
		scriptCtx.put("_PeriodNo", payrollPeriod.getPeriodNo());
		scriptCtx.put("_PeriodID", getHR_Period_ID());

yamelsenih (changes requested):
@SusanneCalderon please change it for merge

e-Evolution:
@SusanneCalderon  please only change the variable to _HR_Period_ID vs _PeriodID because is more clear that we use the payroll period id and not other period

PR #3011, opened 2020-02-12
Add clean definition for delete entities

yamelsenih:
This pull request allows define more than one table for delete using Clean Definition window, see gif
Current ogv

### DIFF
--- a/base/src/org/spin/model/I_AD_CleanDefinition.java
+++ b/base/src/org/spin/model/I_AD_CleanDefinition.java
@@ -1,0 +1,175 @@
+/******************************************************************************
+ * Product: ADempiere ERP & CRM Smart Business Solution                       *
+ * Copyright (C) 2006-2017 ADempiere Foundation, All Rights Reserved.         *
+ * This program is free software, you can redistribute it and/or modify it    *
+ * under the terms version 2 of the GNU General Public License as published   *
+ * or (at your option) any later version.										*
+ * by the Free Software Foundation. This program is distributed in the hope   *
+ * that it will be useful, but WITHOUT ANY WARRANTY, without even the implied *
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.           *
+ * See the GNU General Public License for more details.                       *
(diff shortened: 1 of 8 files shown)


marcalwestf:
@yamelsenih  nice feature. Seems that is saves to go to the database and delete there....

When trying to review, I notice the XML file for creating table, window and menu for _Clean Definition_ table is missing.

In the commit it's also non-existent.
See image:
Selection_353

yamelsenih:
Thanks Mario, I will push xml 

Best regards

yamelsenih:
Hi @marcalwestf this XMKL is done

marcalwestf:
hi @yamelsenih  , the functionality is OK.

I  have a couple of observations:
1.- XML file
here also, change the Release No to 394 and add the PR to the Description.

2.- The second is a mere question:
If I define in window _Clean Definition_, tab _Clean Definition Table_ that the contents of table _M_Requisition_ have to be deleted:
Selection_364

but if in the form _Delete Entities_ I just select the requisition lines:
Selection_363

then requisition lines and requisitions are deleted anyway.
I would expect in this case that if I select requisition lines, only the requisition lines are deleted, not the requisitions.

My question is: does it matter whether I just select the underneath lines or not? 

It seems that only the Clean Definition is considered.

yamelsenih:
hi @marcalwestf 

I  have a couple of observations:

> 1.- XML file
> here also, change the Release No to 394 and add the PR to the Description.
> 
Added
> 2.- The second is a mere question:
> If I define in window _Clean Definition_, tab _Clean Definition Table_ that the contents of table _M_Requisition_ have to be deleted:
> Selection_364
> 
> but if in the form _Delete Entities_ I just select the requisition lines:
> Selection_363
> 
> then requisition lines and requisitions are deleted anyway.
> I would expect in this case that if I select requisition lines, only the requisition lines are deleted, not the requisitions.
> 
> My question is: does it matter whether I just select the underneath lines or not? 
> 
> It seems that only the Clean Definition is considered.

By default all tables and childs added to definition will be cleaned this in ok.

PR #3029, opened 2020-02-25
Bugfix/#requisition document action

yamelsenih:
Currently exists a little bug for Requisition, the problem is that always show prepare and close instead real document action from document sttus. Note the follow gif for current behavior:

Current ogv

**After change:**

Current1 ogv


SusanneCalderon:
Hi @yamelsenih 
I tested it without problems.

PR #3030, opened 2020-02-25
Fixed error with last month day for commission totally paid

yamelsenih:
This pull request resolve a problem with commission for calculation based on invoices totally paid, the problem is that search criteria is based on date ranges **>** and **<** but the correct search criteria should be **>=** and **<=**


SusanneCalderon:
Hi @yamelsenih 
you are right. As between is >=,<= datetrx must use the same criteria.

Issue #1455, opened 2017-11-06
Add Sales Region to Order and Invoice


yamelsenih:
Now exist a Sales Region definition for SO management, it is defined on Sales Region window.

For now the sales region is getted from BP contact but sometimes exist a Sales Region Re-Definition, it cause that all history of sales from a Sales Region can be changed.

My proposal is rounded the Sales Region in Sales Order and Invoice, it allow that a change in BP Location don't change sales history.
![screenshot_20171105_232211](https://user-images.githubusercontent.com/2333092/32424096-b9f4e814-c280-11e7-825b-e1de7e62f36d.png)

e-Evolution:
@yamelsenih adding the sales region and sales order is only to have history sales region? , because currently account engine fill this account dimension using the location.

yamelsenih:
Hi @e-Evolution yes, this is only for keep a sales history from invoice document, the current problem is that  if you change sales region on BP Location the sales history is changed

PR #3031, opened 2020-02-25
Add support to Sales region for Sales Order and Invoice

yamelsenih:
Fixes: #1455 

Screenshot_20200225_151152

Current ogv


SusanneCalderon:
Hi @yamelsenih 
seems a good point for me. I tested it without problems

PR #3042, opened 2020-03-01
Add Invoice Order support

yamelsenih:
Currently exists many Sub-Document Type for handle order:
- Standard = **SO**: Standard order for generate invoices and shipment
- Quotation = **OB**: Quotation (with reserve)
- Proposal = **ON**: Proposal (without reserve)
- Prepay = **PR**: Validate prepay before complete
- POS = **WR**: Generate Invoice and shipment after complete
- Warehouse = **WP**: Used for make a picking
- OnCredit = **WI**: Credit Order
- RMA = **RM**: Return Order Authorization

This pull request allows add a new Sub Document Base for Order **invoiced automatically** used for generate invoiced after complete order, is nice because you can generate invoice and collect it after complete.

The warehouse manager can get order for deliver it. This order is very usefull for Retail where is invoiced and collect a product for a POS but the deliver is make from other warehouse.

The follow is the value:
- Pre-Invoiced = **PI**
Screenshot_20200301_132354

### DIFF
--- a/base/src/org/compiere/model/I_C_DocType.java
+++ b/base/src/org/compiere/model/I_C_DocType.java
@@ -22,7 +22,7 @@ import org.compiere.util.KeyNamePair;
 
 /** Generated Interface for C_DocType
  *  @author Adempiere (generated) 
- *  @version Release 3.9.2
+ *  @version Release 3.9.3
  */
 public interface I_C_DocType 
 {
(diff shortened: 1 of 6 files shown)


marcalwestf:
@yamelsenih  just two suggestions:
- add the link to this pull request to the XML file, so it can be later tracked
- if possible, add to Garden World a Document Type _Pre-Invoiced Order_ like the one you describe so people can try it out; otherwise it can be forgotten or go unnoticed.

These are not mandatory changes, i.e. this pull request can be accepted without them; on the other hand they are "nice to have".

Thanks!!

yamelsenih:
Hi @marcalwestf  offcourse, no problem I will to add it

Best regards

yamelsenih:
Hi @marcalwestf this change is done, Thanks

e-Evolution:
@yamelsenih I think that a best name this feature is "Invoice Order", because the pre is previos mean

Issue #2721, opened 2019-08-24
[Feature Request] Add new Functionality for Replenish Browser


yamelsenih:
## Feature request
I want to add a new Smart browser for replenish from Product Replenish configuration

![Screenshot_20190823_211039](https://user-images.githubusercontent.com/2333092/63630608-77890b80-c5ea-11e9-9f4e-90d24fd28dae.png)

#### Describe the solution you'd like
Is nice to have a way for watch all product to replenish instead a static report
#### Describe alternatives you've considered

None
#### Additional context

See a gif
![Current ogv](https://user-images.githubusercontent.com/2333092/63630614-a7381380-c5ea-11e9-9808-902387017465.gif)

mckayERP:
The changes to ReplenishReport.java have broken the ReplenishReport.  See lines 698 to 700:

    if(!isMandatoryBusinessPartner) {
        localWhere.append(" AND ").append(" AND C_BPartner_ID > 0");
    }

yamelsenih:
Hi @mckayERP can you provide a print screen or step by step for replicate it?

Best regard.

yamelsenih:
Hi @mckayERP here complete error:

![Replenish_Error](https://user-images.githubusercontent.com/2333092/76990938-25f5f480-691f-11ea-81d2-46c92dde1961.gif)


```SQL
or at or near "AND" Position: 312, SQL=SELECT QtyToOrder,C_DocType_ID,Updated,ReplenishType,QtyOnHand,Order_Pack,QtyReserved,QtyOrdered,ReplenishmentCreate,M_WarehouseSource_ID,Order_Min,Level_Min,Level_Max,AD_Org_ID,AD_Client_ID,AD_PInstance_ID,M_Warehouse_ID,M_Product_ID,UpdatedBy,UUID,C_BPartner_ID FROM T_Replenish WHERE (AD_PInstance_ID=? AND AND C_BPartner_ID > 0 AND M_WarehouseSource_ID IS NULL) ORDER BY M_Warehouse_ID, M_WarehouseSource_ID, C_BPartner_ID
```

mckayERP:
Thanks.  It occurred with document types other than PO.

PR #3049, opened 2020-03-18
Fixed error with AND condition for replenish report

yamelsenih:
This pull request resolve a issue with a AND condition unnecessary for query of replenish report

Fixes: #2721 

Replenish_Error

```SQL
or at or near "AND" Position: 312, SQL=SELECT QtyToOrder,C_DocType_ID,Updated,ReplenishType,QtyOnHand,Order_Pack,QtyReserved,QtyOrdered,ReplenishmentCreate,M_WarehouseSource_ID,Order_Min,Level_Min,Level_Max,AD_Org_ID,AD_Client_ID,AD_PInstance_ID,M_Warehouse_ID,M_Product_ID,UpdatedBy,UUID,C_BPartner_ID FROM T_Replenish WHERE (AD_PInstance_ID=? AND AND C_BPartner_ID > 0 AND M_WarehouseSource_ID IS NULL) ORDER BY M_Warehouse_ID, M_WarehouseSource_ID, C_BPartner_ID
```

### DIFF
--- a/base/src/org/compiere/process/ReplenishReport.java
+++ b/base/src/org/compiere/process/ReplenishReport.java
@@ -696,7 +696,7 @@ public class ReplenishReport extends ReplenishReportAbstract {
 		//	For Standard Process
 		StringBuffer localWhere = new StringBuffer("AD_PInstance_ID=?");
 		if(!isMandatoryBusinessPartner) {
-			localWhere.append(" AND ").append(" AND C_BPartner_ID > 0");
+			localWhere.append(" AND ").append(" C_BPartner_ID > 0");
 		}
 		if (!Util.isEmpty(where)) {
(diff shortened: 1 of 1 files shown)


base/src/org/compiere/process/ReplenishReport.java
```diff
@@ -696,7 +696,7 @@ private void createDO() throws Exception
 		//	For Standard Process
 		StringBuffer localWhere = new StringBuffer("AD_PInstance_ID=?");
 		if(!isMandatoryBusinessPartner) {
```
mckayERP: Why does this if statement have an "!"  as in !isMandatoryBusinessParnter.  Shouldn't the test add the C_BPartner_ID > 0 when the Business Partner is mandatory?

PR #3064, opened 2020-03-21
Allows export Packout from ZK web UI

yamelsenih:
This pull request allows export packout from ZK UI, see gif
Export_Packout_from_Web

### DIFF
--- a/base/src/org/adempiere/pipo/PackOut.java
+++ b/base/src/org/adempiere/pipo/PackOut.java
@@ -25,6 +25,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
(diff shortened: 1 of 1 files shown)


marcalwestf:
Hi @yamelsenih , just tested this feature.
- I run it from the IDE (Eclipse)
- Without this feature, the packout file is stored locally (if it is on the server, there is no easy way to get it).
- With this feature applied, I notice that there is a dialog for selecting the location on the laptop/PC where the packout file should be copied, which is OK:
Selection_452
- problem is that it appears 8 times(!!): I have to type in 8 times the name in order to get it copied.

I don't know it it is because I am testing on the IDE ....
But even then it should ask only once.
Perhaps you have any hint.

yamelsenih:
Hi @marcalwestf I test it with a production env and is ok, can you test with aADempiere install?

Best regard

PR #3067, opened 2020-03-21
Allows generate allocation from payment and credit memo from a payment

yamelsenih:
The current behavior is that from a collect can call only invoices (no credit memo) and from a payment only can call AP invoice (no credit memo).

This pull request allows allocate a payment with a AR Credit Memo and Collect with AP Credit Memo from payment.

Note: The Allocation Form allows all it operations.

### DIFF
--- a/base/src/org/compiere/model/MPayment.java
+++ b/base/src/org/compiere/model/MPayment.java
@@ -1385,22 +1385,6 @@ public final class MPayment extends X_C_Payment
 			//	Set
 			isSOTrx = isSOTrxAsString.equals("Y");
 		}	//	now Charge
-		else {
-			if (allocations.length > 0) {
-				for (MPaymentAllocate allocationValue : allocations) {
-					String documentBaseType = DB.getSQLValueString(get_TrxName(), "SELECT dt.DocBaseType "
-							+ "FROM C_Invoice i "
-							+ "INNER JOIN C_DocType dt ON(i.C_DocType_ID = dt.C_DocType_ID) "
-							+ "WHERE i.C_Invoice_ID = ?", allocationValue.getC_Invoice_ID());
-					//	Validate with invoice
-					isSOTrx = documentBaseType.equals(MDocType.DOCBASETYPE_ARInvoice) || documentBaseType.equals(MDocType.DOCBASETYPE_APCreditMemo);
-					if(isSOTrx != isReceipt() 
-							&& !getPayAmt().equals(Env.ZERO)) {
-						return false;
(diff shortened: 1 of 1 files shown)


marcalwestf:
Hi @yamelsenih  I think the new functionality collides with an existing scenario. I try to explain the scenario which right now works and these changes invalidate.

The following scenario is a common situation: you want to offset the payment of a Provider Invoice with a Credit Note you received from the Provider.

Steps to reproduce the scenario:
- Completed an Invoice from a Provider (**API**)
- Completed a Credit Note from a Provider (**APC**)
- Create a Payment to the Provider (**isReceipt=false**)
- Execute the _Complete_ action.

The result it that due to the Payment API <-> APC combination described above, the program classifies the scenario as isSOTrx == true due to the appearance of **APC**:
Selection_773

Then, in the following line, this isSOTrx (in our case ==true) is compared with the isRecepit value of the document (as stated before ==false):
Selection_774

If the comparison yields false, the verification is rejected:
Selection_775

As a follow-up, the MPayment.prepare() is rejected and the message "_Inconsistent Document Type and Invoice"_ appears:
Selection_776

This very common scenario would not work anymore.
Looking forward to your comments.

yamelsenih:
Hi @marcalwestf I don't sure if it worked on previous version. Always I use Payment Allocation as process berfore any payment.
 Said you add a payment with two Invoice to pay?

yamelsenih:
Just remove a unnecessary validation because already exist a validation for allocate amounts

marcalwestf:
@yamelsenih I will test it.

marcalwestf:
hi @yamelsenih to be sure of your question 

> Said you add a payment with two Invoice to pay?

, I just tested the behavior on Adempiere 393.
There, it is possible to create a _Provider Credit Note_ and refer to it on a _Payment_.  On _prepareIt()_, the method _MPayment.verifyDocType()_ delivers true.

Doing the same with this PR's changes causes _MPayment.verifyDocType()_ to deliver false and yields the error message displayed above.

This means, this funcionality would not be accesible anymore.

Perhaps you could do some changes it so this is possible?

marcalwestf (approved):
- Created and completed both a Provider Invoice and a Provider Credit Memo.
- Payment allocation is generated for both Provider Invoice and Provider Credit Memo.
- Checked the allocations in both Payments and Invoices. Result: Ok.
- The validation in MPayment which invalidated the completion was eliminated
- Payment Allocation takes care of validation.

Test passed.

PR #3072, opened 2020-03-21
Allows more tender types for payments

yamelsenih:
This pull request just add more tender type for payment, see:
Screenshot_20200321_003216


marcalwestf:
@yamelsenih  I loaded and applied the XML file.
I can see that in windows _Cash_ and _Payment_ the Bank was added and also additional tender types:
Selection_462
This is OK and cmplies with the pull request.

I notice that you can select a _Bank Account_ which doesn't belong to the selected bank.
Perhaps the Bank Account should get a **Dynamic Validation**?
Otherwise the user could select a bank account different from the bank or vice versa.

yamelsenih:
Hi Mario the bank is used for reference of collect bank from POS.

PR #3073, opened 2020-03-21
A new process for create RMA from Invoice instead delivery

yamelsenih:
This pull request allows define a process for create Return Material Order from a invoice using **Ref_InOutLine_ID** from Invoice, also allows create Return order for POS when it are not delivered.

The use, just calling from source code with role authorization

### DIFF
--- a/base/src/org/compiere/model/I_C_OrderLine.java
+++ b/base/src/org/compiere/model/I_C_OrderLine.java
@@ -22,7 +22,7 @@ import org.compiere.util.KeyNamePair;
 
 /** Generated Interface for C_OrderLine
  *  @author Adempiere (generated) 
- *  @version Release 3.9.2
+ *  @version Release 3.9.3
  */
 public interface I_C_OrderLine 
 {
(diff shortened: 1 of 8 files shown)


marcalwestf:
@yamelsenih I think the problems come because the project cannot be compiled.
The model is missing the method _MOrderLine.setRef_InvoiceLine(invoiceLine)_.

1.- Compiler error in process _OrderRMACreateFromInvoice_
Selection_458

2.- No column _Ref_InvoiceLine_ in table _C_OrderLine_:
Selection_457

yamelsenih:
Hi Mario, I apologize, a class is missing from it. Let me resolve it.

Thanks

yamelsenih:
Hi @marcalwestf this problem was resolved

marcalwestf:
hi @yamelsenih  I have two more observations:
1. The field _Referenced Invoice Line_ is created in the _RMA Line_ tab of both windows _Return Material Order (Customer)_ and _Return Material Order (Vendor)_. 
This is OK.
But, the field _Referenced Shipment Line_ is still mandatory. I suppose it should now be elegible to either take _Referenced Shipment Line_ or _Referenced Invoice Line_:
Selection_470

2. The process _org.spin.process.OrderRMACreateFromInvoice_ is created and the commit contains the source code of the process.
This is also OK.
But I am missing a button or a menu entry where I can call the process, because as of now, there is no way to call the process..

yamelsenih:
Hi @marcalwestf is rigth, the field IN Out Reference should not be mandatory from window, I will change it now.

The process is a definition for be used from POS or reference from source code, maybe can be used as Smart Browse for create from on Return material but for now jus I use it from POS on the next pull request

yamelsenih:
Hi @marcalwestf this process can be used from POS for generate returns based on invoices for orders type **Invoice Order** that can be invoiced before delivery

marcalwestf (approved):
Test passed

PR #3074, opened 2020-03-21
remove old implementation for Generate InOut process

yamelsenih:
Just improve source code and remove old implementation for Selection of process and save by saveEx


marcalwestf:
I execute the form _Generate Shipments (manual)_:
Selection_469

I have the following error when executing the process :
Selection_467

With an exception:
Selection_468

This exception is thrown because Process Instance's Client_ID is null:
Selection_466
Selection_464

Due to the fact that all other processes work well, I might suppose that the form _Generate Shipments (manual)_ somehow misses to set AD_Client_ID at PInstance .

yamelsenih:
Hi @marcalwestf this was solved yet, please remember apply merge as "**Squash and merge**" option

e-Evolution:
this change introducing other issues

@yamelsenih @marcalwestf  The error was introducing #3074 ,the problem is that replace the manual parameter creation by auto generate parameter class for the process, the problem is that system use a internal parameter 'Selection' that is read but not is declared in process , this cause the InvoiceGen generate all of lines.

@marcalwestf , the testing not was very simple , the test need to do with multiples lines

e-Evolution:
@yamelsenih , I got this error in my testing 

===========> InOutGenerate.process: java.lang.NullPointerException [68]
java.lang.NullPointerException
	at org.compiere.process.InOutGenerate.doIt(InOutGenerate.java:109)
	at org.compiere.process.SvrProcess.process(SvrProcess.java:177)
	at org.compiere.process.SvrProcess.startProcess(SvrProcess.java:130)
	at org.adempiere.util.ProcessUtil.startJavaProcess(ProcessUtil.java:160)
	at org.adempiere.util.ProcessUtil.startJavaProcess(ProcessUtil.java:105)
	at org.compiere.apps.ProcessCtl.startProcess(ProcessCtl.java:625)
	at org.compiere.apps.ProcessCtl.run(ProcessCtl.java:374)
	at org.adempiere.webui.apps.form.WGenForm.runProcess(WGenForm.java:296)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.zkoss.zk.ui.impl.EventProcessor.process0(EventProcessor.java:209)
	at org.zkoss.zk.ui.impl.EventProcessor.process(EventProcessor.java:141)
	at org.zkoss.zk.ui.impl.EventProcessingThreadImpl.process0(EventProcessingThreadImpl.java:519)
	at org.zkoss.zk.ui.impl.EventProcessingThreadImpl.run(EventProcessingThreadImpl.java:446)

homebeaver:
Hi @yamelsenih , @marcalwestf 

> Test passed. !

-- the Query to find orders in `line 126 InOutGenerate.doIt()` will produce 
```sql
SELECT ... FROM C_Order
WHERE DocStatus='CO'
  AND IsSOTrx='Y'
  AND EXISTS(SELECT 1 FROM C_DocType dt WHERE dt.C_DocType_ID = C_Order.C_DocType_ID AND DocBaseType='SOO' AND DocSubTypeSO NOT IN ('ON','OB','WR'))
  AND IsDropShip='N'
  AND DeliveryRule<>'M'
...
```

In my opinion there should be an additional "` AND isdelivered='N' `" to avoid creating inouts for delivered orders.

regards EUGen

e-Evolution:
@homebeaver , do you testing with multiple records from Generate Invoices (Manual ) or Generate Shipments (Manual)? when you trying only select a row, the system creates an invoice or shipments for all records without being selected

homebeaver:
Hi @e-Evolution 
for me that is not the point, because we call `InOutGenerate ` from another "macro" process.
The point is: orders which are fully delivered should not be candidate for a shipment.
wdyt?

e-Evolution:
@homebeaver Well, the point is it does a peer review test, and it should not report that the test passed when it doesn't work when there are multiple records.

About not allowing to deliver more than ordered, it is already possible with the correct configuration, that is, complete delivery rule, complete line or available, the system has a view with which it validates which orders are candidates for delivery according to the delivery rule. 

on the other hand, if reality surpasses theory, it is possible to change the delivery rule to forced, which allows a product to be delivered even when there is no inventory available, so I think it was well thought out and it is better than the system supports the most business cases

PR #3078, opened 2020-03-21
fixed error with conversion rate for bank statement from payment

yamelsenih:
The problem:
The convrsion rate is finded but never is allocated to payment amount:
```Java
maybeCurrencyRate.ifPresent(currencyRate -> paymentAmount.multiply(currencyRate)
					.setScale(currency.getStdPrecision(), BigDecimal.ROUND_HALF_UP));
```

then a payment with a distint currency of bank account never is converted, this pull request change it


marcalwestf:
@yamelsenih  one question: should it be on line 193 **payment.getC_Currency_ID()** instead of **bankAccount.getC_Currency_ID()** ?
The code seems to get a conversion rate from the same currency:
Selection_473

yamelsenih:
Hi @marcalwestf please taste it. hahah
:+1:

Issue #3085, opened 2020-03-25
[Feature Request] Add Asset Assignment


yamelsenih:
## Feature request
Create a way for add support to asset assignment, some like tools for employees or BP rental, is nice to have a Smart Browser for allocate a Asset like tool or security implement to employee. Also can be applied to customer for rent a asset.

#### Describe the solution you'd like
The fixed asset have a tab for financial information that can be implement to handle functionality.
![Screenshot_20200325_130252](https://user-images.githubusercontent.com/2333092/77564342-f30cad00-6e98-11ea-8b29-22b06087c297.png)

#### Describe alternatives you've considered
- A Smart Browser for allocate and deallocate asset with filters like Asset group, Employee, and others.
- Aso I think that is nice if is used the **Deliver Assets** Process, see:
![Screenshot_20200325_134404](https://user-images.githubusercontent.com/2333092/77568235-b479f100-6e9e-11ea-83cf-aa00aba6b36b.png)
![Screenshot_20200325_134343](https://user-images.githubusercontent.com/2333092/77568277-c52a6700-6e9e-11ea-9013-d6ffba0d956e.png)

#### Additional context
It is very useful for maintenance and production enterprises.

## Main Proposal
- Just make changes for allows deliver a asset with optional send mail
- Improve source code for best practices
- Enable tab "**Asset Delivery**"
- Add a Smart Browser for generate **Asset Delivery**
- Add a Smart Browser for return delivery (it is useful when the employee return tool or other asset to main warehouse)

yamelsenih:
Here is the complete functionality integrated:
![Deliver_Asset](https://user-images.githubusercontent.com/2333092/77675813-cffaff80-6f63-11ea-8de4-dbc15c576cb4.gif)

PR #3087, opened 2020-03-26
Add support to Deliver Asset Browser

yamelsenih:
Fixes: #3085

This pull request allows deliver many assets from browser. See Gif:
Deliver_Asset

Best regards

### DIFF
--- a/base/src/org/compiere/model/I_A_Asset_Delivery.java
+++ b/base/src/org/compiere/model/I_A_Asset_Delivery.java
@@ -22,7 +22,7 @@ import org.compiere.util.KeyNamePair;
 
 /** Generated Interface for A_Asset_Delivery
  *  @author Adempiere (generated) 
- *  @version Release 3.9.2
+ *  @version Release 3.9.3
  */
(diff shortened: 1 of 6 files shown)


marcalwestf:
One question @yamelsenih : in _I_A_Asset_Delivery_, line 106 I see
`public I_C_ValidCombination getAD_User() throws RuntimeException;`

In  the original file, I find this:
`public org.compiere.model.I_AD_User getAD_User() throws RuntimeException;`

I wonder if the change is OK.

yamelsenih:
Hi Mario let me see it.

Thanks.

marcalwestf:
1. XMLs loaded and applied
2. Smart Browser executed, query parameters set
3. Search and select Assets
4. Process executed (in single step)

Result: Delivery created OK.
Selection_001

marcalwestf:
@yamelsenih so the funcionality is actually OK.
Depending on your answer, it can be accepted immediately.

yamelsenih:
Hi @marcalwestf the reference was changed to Account instead search, I fix it yet.

Thanks

PR #3092, opened 2020-03-31
Add support to Optional Export Format for Table and Documents of

yamelsenih:
Just a simple change for Replication Strategy that allows select a export format by default on Table and Document, see image:
Screenshot_20200331_111755
Screenshot_20200331_111806

- By default and for old compatibility if a export format is not setted then the search is by Client / Table / Version
- If a export format is setted then the search is by Export Format

### DIFF
--- a/base/src/org/adempiere/model/ExportModelValidator.java
+++ b/base/src/org/adempiere/model/ExportModelValidator.java
@@ -31,6 +31,7 @@ package org.adempiere.model;
 
 import org.adempiere.process.rpl.exp.ExportHelper;
 import org.compiere.model.MClient;
+import org.compiere.model.MReplicationDocument;
 import org.compiere.model.MReplicationStrategy;
 import org.compiere.model.MReplicationTable;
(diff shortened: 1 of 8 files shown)


trifonnt:
Hi Yamel,

how do you handle searching of Export Format during import of XML file?

In XML file you do not know which is the Export Format, you know the root of the XML element.

I reference the case when one table for example C_BPartner has more than one export Format.

Kind regards,
Trifon

yamelsenih:
Hi @trifonnt for import format is used Value of Export Format exported:

```Java
String EXP_Format_Value = null;
		EXP_Format_Value = rootElement.getNodeName();
		log.info("EXP_Format_Value = " + EXP_Format_Value);
```

I think that it should not a problem because **Value** is a index for it

yamelsenih:
Best regards

trifonnt:
Hi @yamelsenih,

yes, this will work.

Kind regards,
Trifon

PR #3095, opened 2020-04-01
Add Validation Cost Element by Landed Cost Type

carlosaparadam:
When applied landed costs on Invoice Vendor, this set a default Cost Element that will could be not an Landed Cost type. Additionally if generated Landed Cost from the Smart Browse (Create Landed Cost From Receipt) this don't filter by Cost Element, merging Landed Cost with Material Cost type.

**Landed Cost Window:**
LandedCostType_Window

**Landed Cost Window Proposal:**
LandedCostType_Window_After

**Smart Browse (Create Landed Cost From Receipt):**
LandedCost-SB

**Smart Browse (Create Landed Cost From Receipt) Proposal:**
LandedCost-SB-After

**Additionaly, add code validation before save:**
```
if (getM_CostElement_ID()!=0
	&& !MCostElement.COSTELEMENTTYPE_LandedCost.equals(getM_CostElement().getCostElementType())) {
	log.saveError("Error", Msg.parseTranslation(getCtx(), 
		"@Invalid@ @M_CostElement_ID@"));
	return false;
}
```

### DIFF
--- a/base/src/org/compiere/model/MLandedCost.java
+++ b/base/src/org/compiere/model/MLandedCost.java
@@ -131,6 +131,13 @@ public class MLandedCost extends X_C_LandedCost
 				"@NotFound@ @M_Product_ID@ | @M_InOut_ID@ | @M_InOutLine_ID@"));
 			return false;
 		}
+		
+		if (getM_CostElement_ID()!=0
+				&& !MCostElement.COSTELEMENTTYPE_LandedCost.equals(getM_CostElement().getCostElementType())) {
+			log.saveError("Error", Msg.parseTranslation(getCtx(), 
+					"@Invalid@ @M_CostElement_ID@"));
+				return false;
+		}
 		//	No Product if Line entered
 		if (getM_InOutLine_ID() != 0 && getM_Product_ID() != 0)
 			setM_Product_ID(0);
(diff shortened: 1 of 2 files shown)


SusanneCalderon:
@carlosaparadam 
We are working with Average Costing and distribute the landed costs on material costs. The distribution of landed costs on a different cost element can lead to erroneous costs when a landed cost is not applied always. 

Example: a product which is sometimes imported (causes import costs) sometimes not. If the landed costs are distributed on an extra cost element the average costs can be too high and the quantities of the 2 cost elements may not match.

For simple installations it is easier to understand the average cost on one single cost element and also matches with legacy implementations. 

The usage of more cost elements makes the import  of a valuated inventory which provides one single cost per product extremely difficult.

Adempiere should not restrict the possibilities how the costing is calculated and how your costing system is used. It should be the decision of the implementer if the application uses one or more cost elements. If the decision is made to use only material costs it must be possible to distribute landed costs on this cost element.

carlosaparadam:
hi @SusanneCalderon, i understand your point, however this don't match with the window validation where has a validation for only landed cost type applied. I purposed this validation thinking that was  forgotten on smart browse and because in our experiences occurs that the user by error select an material cost type and when their want see by separated we finished changing this value and generating costing for the product, we could controlling this by ASP, but i think so this should be unified if don't have a validation on smart browse don't should be on the window. 

I wait for your comments, regards and thanks for reviewed this case and shared your point of view.

marcalwestf (commented):
The legacy functionality is that you can use the cost element "Material" and apply to it any landed cost  to calculate the average cost.

The changes requested would disrupt this functionality.

marcalwestf:
The undelying issue of displaying the same Cost Elements on both Window and Smart Brwoser should be matter of a new issue and subsequent pull request.

For now, the pull request will be closed.

PR #3096, opened 2020-04-02
Fixed error with process builder when flag "IsManagedTransaction" is false

yamelsenih:
The problem: currently the **ProcessBuilder** class have a method for unhandle transaction from it, the flag is named **withoutBatchMode()** and the current behavior is not implemented correctly.

**What is the correct behavior?**
When a process builder is runned with **withoutBatchMode()** method called, it will should run the process without transaction and allows that transaction can generate and closed inside process.

### DIFF
--- a/base/src/org/eevolution/service/dsl/ProcessBuilder.java
+++ b/base/src/org/eevolution/service/dsl/ProcessBuilder.java
@@ -64,7 +64,7 @@ public class ProcessBuilder {
     private ASyncProcess parent;
     private List<Integer> selectedRecordsIds;
     private int tableSelectionId;
-    private boolean isBatch = false;
+    private boolean isBatch = true;
     private boolean isPrintPreview = false;
     private String reportExportFormat = null;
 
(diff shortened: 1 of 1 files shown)


e-Evolution:
@yamelsenih This flag was created to manage if the transaction is closed or is handled by the developer, this allows cascading of multiple processes, preventing a sub process from closing the transaction of the parent process.

So this behavior should not be changed.

The is batch is a legacy functionality of Adempiere and is controlled by the SvrProcess class and is intended for another approach.

Please for backward compatibility do not change the current operation, if you need to implement something additional for is Batch it can be done but I think it was intended for another purpose

yamelsenih:
> @yamelsenih This flag was created to manage if the transaction is closed or is handled by the developer, this allows cascading of multiple processes, preventing a sub process from closing the transaction of the parent process.
> 
> So this behavior should not be changed.
> 
> The is batch is a legacy functionality of Adempiere and is controlled by the SvrProcess class and is intended for another approach.
> 
> Please for backward compatibility do not change the current operation, if you need to implement something additional for is Batch it can be done but I think it was intended for another purpose

Hi @e-Evolution you are right, the flag is **isBatch** instead, I change it and worked with old compatibility.

Please take a time for test it with a implementation, I maked a test with real data and is ok now

yamelsenih:
Maybe can be implemented with other way

PR #3107, opened 2020-04-17
Add support to calculate payroll process for all organization

yamelsenih:
This pull request allows define a payroll process for calculate for all employees of all organization, just defina a payroll process as **_*_** and is getted all employees on payroll.
Best regards


e-Evolution:
@yamelsenih  first of all you don't need to register a document as * to process payroll for multiple employees in different organizations, since this depends on which organization registered in the employee's data.

image

On the other hand, this breaks with the standard ADempiere logic where all documents that generate transactions cannot be registered with *.

This may completely affect intercompany accounting functionality.

PR #3119, opened 2020-04-25
Add process from cache for Server Process Control

yamelsenih:
This pull request just change definition for **ServerProcessCtl** for get process from cache instead requery, previously existed it for **ProcessCtl**

It improve performance for server...


e-Evolution:
@yamelsenih , please can refectory name for obsolete practice to use m_ as prefix in local variable 
as m_parent , m_trx this way I can merge

kind regards
Victor

yamelsenih:
Off course, let me make my wish hhaha

yamelsenih:
Hi @e-Evolution it is done

rcapecce:
Hi @yamelsenih & @e-Evolution, I am telling you what is happening to us in the POS functionality of Openup when I generate an Invoice from the POS.

We with our POS version are obtaining a Duplicate Key in the T_Selection table in the CPOS.generateShipment method (https://github.com/OpenupSolutions/adempiere/blob/opup/retail/org.adempiere.pos/src/main/java/base/org/adempiere/pos/service/CPOS.java#L1315)

This run the process "M_InOut_Generate (manual)" (AD_Process_ID = 99) and on ProcessBuilder L304 and L305 try to insert into table T_Selection 2 times and get a Duplicate Key constraint (https://github.com/adempiere/adempiere/blob/develop/base/src/org/eevolution/service/dsl/ProcessBuilder.java#L304-L305)

Here inside the first method (https://github.com/adempiere/adempiere/blob/develop/base/src/org/eevolution/service/dsl/ProcessBuilder.java#L200-L208) there is a "TODO" that indicates to delete the duplicated functionality between the indicated classes.

Checking on this commit, I see that this functionality was added in the ServerProcessCtl class and this is where you insert the T_Selection lines a second time and get the constraint error (37def67#diff-26346dac0f85f417c6cd6c7380f77094R198-R201 (scroll to the selected lines, the page doesn't jump automatically))

We removed these lines (ServerProcessCtl # L198-L201) and only made a single insert in the T_Selection table allowing to create the invoice from the POS

Issue #3131, opened 2020-06-03
Error calculating parent tax on sales order with warehouse document type


e-Evolution:
## Bug report
Error calculating parent tax on sales order with warehouse document type.
The result of the incorrect calculation is that the parent tax that is included in the order when it should only exist is the child tax.

This is caused by the fact that the calculation of the sales order line does not validate the parent and child taxes.

#### Steps to reproduce

1. Crate a Tax Parent  with two Tax Children 
2. Create a Sales Order with document type Order Credit
3. Adding a Sales Order Line with a product and use the Tax Parent
4. Complete Sales Order
5. Check the Sales Order Tax , here show the Tax Parent and Tax Children
#### Screenshot or Gif

![image](https://user-images.githubusercontent.com/786968/83687548-86591180-a5b1-11ea-9f21-5a462aabcb35.png)

#### Expected behavior
Only should show the children tax based on tax parent.

![image](https://user-images.githubusercontent.com/786968/83687444-632e6200-a5b1-11ea-9da3-adfc643193c9.png)
#### Other relevant information
- Your OS: MacOS
- Java version: Java 8
- ADempiere version:
- Browser safari
- DB Provider: PG
- DB version: 12.2

PR #3132, opened 2020-06-04
#3131 Error calculating parent tax on sales order with warehouse docu…

e-Evolution:
…ment type https://github.com/adempiere/adempiere/issues/3131

- Remove duplicated code with same funcionality


base/src/org/compiere/model/MOrderLine.java
```diff
@@ -1009,7 +1009,7 @@ protected boolean afterDelete (boolean success)
 	 * 
 	 * @author teo_sarca [ 1583825 ]
 	 */
-	private boolean updateOrderTax(boolean oldTax) {
+	/*private boolean updateOrderTax(boolean oldTax) {
```
yamelsenih, line 1012: If it is't util please remove complete

PR #3136, opened 2020-06-06
Removing JBoss Wildfly 19 obstacles

marekmosiewicz:
This pull request removes just things which mess with newest JBoss. In tag lib there are some tags <import> which seems to be not strict schema. Removing them (and one <br/> in comment) makes it possible to deploy zkwebui on new JBoss. You need just to add Adempiere.jar and AdempiereSLib into WEB-INF/lib of zkwebui.war and set property PropertyFile pointing do Adempiere.properties of your Adempiere home. When starting JBoss (you need to download it and install) with -DPropertyFile=<Adempiere.properties location> you can deploy zkwebui.war from Jboss console (port 9990)
This pull request does not contain new ZK9 or replace default JBoss. It is just clean to make it work with new  JBoss

I have tested that removing <import> tags does not makes known problems. It deploys on JBoss and Tomcat and seems to work correctly

### DIFF
--- a/zkwebui/WEB-INF/tld/web/core.dsp.tld
+++ b/zkwebui/WEB-INF/tld/web/core.dsp.tld
@@ -92,17 +92,6 @@ Copyright (C) 2005 Potix Corporation. All Rights Reserved.
 		<tag-class>org.zkoss.web.servlet.dsp.action.When</tag-class>
 	</tag>
 
-	<!--         -->
-	<!-- Imports -->
-	<!--         -->
-	<import>
-		<import-name>ServletFns</import-name>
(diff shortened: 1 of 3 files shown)


yamelsenih:
Hi @marekmosiewicz is nice it. I will work with adempiere based on gradle and I think that can be a good idea start with little changes like this

marekmosiewicz:
> Hi @marekmosiewicz is nice it. I will work with adempiere based on gradle and I think that can be a good idea start with little changes like this

Gradle script (https://github.com/marekmosiewicz/adempiere-meta) currently uses build scripts from corresponding projects (it invokes ant for adempiere, own gradle scripts for  adempiere-gRPC-Server and npm for adempiere-vue)

Issue #3141, opened 2020-06-17
Error when Generate Shipment and Invoice and selection multiples rows


e-Evolution:
## Bug report
Error when Generate Shipment and Invoice and selection multiples rows

#### Steps to reproduce
1.-  Create 2 Sales Order Standard and Immediate Invoice Rule
2.- Entry the form Generate  Invoice (Manual)
3.- Selected one row the multiples rows that are shown
4.-Then system generate an invoice for all the rows, even when they were not selected

#### Screenshot or Gif
![image](https://user-images.githubusercontent.com/786968/84852731-2b003800-b023-11ea-90ed-409f813ef62c.png)

#### Expected behavior
The system should only generate an invoice for row selected

#### Other relevant information
- Your OS: MacOS
- Java version: Java 8
- ADempiere version: 3.9.3
- Browser safar- DBProvider:PG
- DB version:12

PR #3142, opened 2020-06-17
Error when Generate Shipment and Invoice and selection multiples rows

e-Evolution:
solved  #3141


e-Evolution:
@yamelsenih @marcalwestf  The error was introducing #3074 ,the problem is that replace the manual parameter creation by auto generate parameter class for the process, the problem is that system use a internal parameter 'Selection' that is read but not is declared in process , this cause the InvoiceGen generate all of lines.

yamelsenih:
A simple test:
```Java
SELECT AD_Process_ID, AD_Process_Para_ID FROM AD_Process_Para WHERE ColumnName = 'Selection';
 ad_process_id | ad_process_para_id 
---------------+--------------------
(0 rows)

```

Is ok because Selection is not used as parameters for process

rcapecce:
Hi @e-Evolution, I am telling you what is happening to us in the POS functionality of Openup when I generate an Invoice from the POS in case it has something to do with what you are commenting on.

We with our POS version are obtaining a Duplicate Key in the T_Selection table in the CPOS.generateShipment method (https://github.com/OpenupSolutions/adempiere/blob/opup/retail/org.adempiere.pos/src/main/java/base/org/adempiere/pos/service/CPOS.java#L1315)

This run the process "M_InOut_Generate (manual)" (AD_Process_ID = 99) and on ProcessBuilder L304 and L305 try to insert into table T_Selection 2 times and get a Duplicate Key constraint (https://github.com/adempiere/adempiere/blob/develop/base/src/org/eevolution/service/dsl/ProcessBuilder.java#L304-L305)

Here inside the first method (https://github.com/adempiere/adempiere/blob/develop/base/src/org/eevolution/service/dsl/ProcessBuilder.java#L200-L208) there is a "TODO" that indicates to delete the duplicated functionality between the indicated classes.

Checking on this commit, I see that this functionality was added in the ServerProcessCtl class and this is where you insert the T_Selection lines a second time and get the constraint error (https://github.com/adempiere/adempiere/commit/37def67ce30015b2ca11c2db3135303b4ef1c13f#diff-26346dac0f85f417c6cd6c7380f77094R198-R201 (scroll to the selected lines, the page doesn't jump automatically))

We removed these lines (ServerProcessCtl # L198-L201) and only made a single insert in the T_Selection table allowing to create the invoice from the POS

PR #3149, opened 2020-06-22
Fixed error with not updateables columns for PO.copyValues method

yamelsenih:
Just a little change but very important. This pull request allows validate if a column is updateable before overwrite it in a copy


e-Evolution:
@yamelsenih  I consider that this criterion is not correct, you may want the value of a column to be copied and not updated and that is ok, this happens because the column can be set only if it has not been inserted once inserted in the base data can no longer be changed.

if what is required is that a copied record that has not been saved allows updating which is another valid possibility, I recommend the configuration (allows copy ->  not updateable, but it is a new record) then  allows setting a value because it has not been inserted

So if you adding the isNew validation would work to any configuration

yamelsenih:
The standard is ok

PR #3151, opened 2020-06-23
Add cascade constraints for ASP tables

yamelsenih:
This pull request allows remove dependences of ASP from tables.


marcalwestf:
@yamelsenih  can you explain what is the actual error?
I see there are lots of constraints re-defined to additionally become _ON UPDATE CASCADE ON DELETE CASCADE_ so mainly if you delete one record you delete all dependencies, which may be correct. The actual constraint definition takes the default (_ON DELETE NO ACTION_).

I suppose the error you want to avoid is to leave no "orphans" i.e. records pointing to non-existing entities?
Thanks

marcalwestf:
@yamelsenih  as I don't have replication instances, what would be the easiest way to test some of these new constraints?

e-Evolution:
@yamelsenih @marcalwestf  please not merge any XML script to created constraints because this process is executing when the new database seed is generated and the process including the constraints creations , so no is necessary include as XML script

marcalwestf:
@e-Evolution  what do you suggest to be done here Víctor?
Would this PR be postponed, closed, or replaced by actions taken when preparing the release?
If so, how can it be documented so it is not forgotten?
Thank you.

marcalwestf:
@e-Evolution  as I understand Víctor, this PR would be considered at the end, just before the 3.9.4 release, right?

e-Evolution:
this PR was include with release 3.9.4

PR #3155, opened 2020-07-02
Add new Smart Browser for Copy Window from Role

yamelsenih:
Just a little change, now you can copy windows from other role... Very easy, see the gif:
Peek 2020-07-01 22-26

### DIFF
--- a/base/src/org/spin/process/CopyWindowFromRole.java
+++ b/base/src/org/spin/process/CopyWindowFromRole.java
@@ -1,0 +1,135 @@
+/******************************************************************************
+ * Product: ADempiere ERP & CRM Smart Business Solution                       *
+ * Copyright (C) 2006-2017 ADempiere Foundation, All Rights Reserved.         *
+ * This program is free software, you can redistribute it and/or modify it    *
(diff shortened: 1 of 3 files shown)


e-Evolution:
@yamelsenih please when use the Query api , create an private or public method to get the data , is more legible the code that mix all in a method 

https://github.com/adempiere/adempiere/pull/3155/files#diff-5ff9ca8f1e2f1f78ef5e5f587f3ae954R64

With this change can to make merge

kind regards

yamelsenih:
Hi Victor I am using query API. See it

A exists is a ANSI SQL used.

See the code new Query()

e-Evolution:
@yamelsenih , yes I saw but my comment is other, is good practique encapsulate a query that returns a collection in a method with  with a business name, like you did with isExistsWindow or isExistsProcess, instead of putting all codeine main logic method

yamelsenih:
Ummm, good, I can add a method for retur collection bases on current query
instead

yamelsenih:
Hey this is ready to merge

PR #3094, opened 2020-03-31
Fix Landed Costs Accounting, when Vendor invoice is in another currency

carlosaparadam:
Currently, when you create a vendor invoice with associated costs in a different currency of purchase receipt, this generate an difference on costing amount and accounting amount.

Show this example:

**_Invoice Vendor info_**
Invoice-Header
Invoice Line

|**_Invoice Currency:_** EUR|
|:---|
|**_Currency Rate:_** 1.249|
|**_Costing Precision:_** 4|
|**_Invoice Amount Allocation:_** 987.63|

**_Invoice Landed Cost Allocation_**
LandedCost

|**Amount**|**Quantity Receipt**|
|---:|---:|
|123.45 €|27,653.00|
|123.45 €|30,000.00|
|123.45 €|30,000.00|
|123.45 €|29,986.00|
|123.45 €|26,780.00|
|123.45 €|28,632.00|
|123.45 €|6,949.00|
|123.48 €|30,000.00|
|**987.63€**|**210.000,00**|

**_Cost Generated_**
When generated costing at completing Invoice, this set current cost price   (costing amount / quantity receipt ), rounding to costing precision and then multiply by currency rate, rounding to costing precision, example: 123.45 / 27,653.00 rounding to 4 decimals =  0.0045 and the result is multiply by currency rate (1.249) and rounding to 4 decimals = 0.0056. Then the amount is calculated, multipling current cost price (0.0056) by quantity (27,653.00) = 154.8568

CostGenerated

|**Landed Cost Amount**|**Receipt Quantity**|**Current Cost Price**|**Amount**|
|---:|---:|---:|---:|
|123.45 €|27,653.00|0.0056 $|154.8568$|
|123.45 €|30,000.00|0.0051 $|153.0000$|
|123.45 €|30,000.00|0.0051 $|153.0000$|
|123.45 €|29,986.00|0.0051 $|152.9286$|
|123.45 €|26,780.00|0.0057 $|152.6460$|
|123.45 €|28,632.00|0.0054 $|154.6128$|
|123.45 €|6,949.00|0.0222 $|154.2678$|
|123.48 €|30,000.00|0.0051 $|153.0000$|
|**987.63€**|**210.000,00**|**--**|**1,228.312$**|

**_Accounting_**
[...]

PR #3157, opened 2020-07-02
#3094 Fix Landed Costs Accounting, when Vendor invoice is in another …

e-Evolution:
…currency https://github.com/adempiere/adempiere/pull/3094

- Implement rounding based on the price list precession


e-Evolution:
I found 3 problems in the functional part and here my proposal
solution :

1.- the date that the model was taking for the additional cost was the date of receipt of materials when it should take the invoice date, this generated a difference since the cost calculated the average layer based on the invoice date, so two different types of changes were used.
2.- The additional costs were not verified and the conversion was not made, that is why I used the getPriceActual method to bring the unit cost in base or company currency price, then looking at the Carlos code only simplify it
3.- There is a behavioral requirement that I have not had before, which is the calculation of unit costs with high pressure, this seems to happen only in very specific cases, after trying some approaches, including changing the precision of the currency, I found it more appropriate to be able to define it in a more granular way, so implement the precision-based rounding defined in the price list, since it seems to me that you can define a list with higher precision only for the additional costs they have this requirement.

You can tell me if these changes meet your requirements.

carlosaparadam:
hi @e-Evolution, I tested it and it is ok.

PR #3188, opened 2020-08-01
Merge pull request #3184 from erpcya/bugfix/#RPL-UUID-created-is-not-rep...

Hayesie88:
licated

Fixed error when a po is save using saveReplica instead save


yamelsenih:
hI @Hayesie88 this was merged here: https://github.com/adempiere/adempiere/pull/3184

PR #3221, opened 2020-09-04
#3217 MAttributeSet method excludeEntry() is misused  https://github.…

### LINKED ISSUE #3217: MAttributeSet method excludeEntry() is misused
The method is defined to use AD_Table_ID but it is called with a mix of table and column ids.  

This is a cut and paste of the references for MAttributeSet.excludeEntry().  Not the various calls.

-  base/src/org/compiere/model/MAttributeSet.java
--  attributeSet.excludeEntry(**tableId**, isSOTrx)
- base/src/org/eevolution/service/ScanBar.java
-- attributeSet.excludeEntry(**getTableLine().getAD_Table_ID()**, Env.isSOTrx(product.getCtx()))
- client/src/org/compiere/grid/ed/VPAttribute.java
-- mas.excludeEntry(**m_AD_Column_ID**, Env.isSOTrx(Env.getCtx(), m_WindowNo))
- zkwebui/WEB-INF/src/org/adempiere/webui/editor/WPAttributeEditor.java
-- mas.excludeEntry(**m_AD_Column_ID**, Env.isSOTrx(Env.getCtx(), m_WindowNo))

e-Evolution:
…com/adempiere/adempiere/issues/3217

- Fixed #3217 the parameter to previous incorrect refractory

### DIFF
--- a/client/src/org/compiere/grid/ed/VPAttribute.java
+++ b/client/src/org/compiere/grid/ed/VPAttribute.java
@@ -16,7 +16,6 @@
  *****************************************************************************/
 package org.compiere.grid.ed;
 
-import java.awt.AWTEvent;
 import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Component;
@@ -32,7 +31,6 @@ import java.beans.PropertyVetoException;
 import java.util.logging.Level;
 
 import javax.swing.JComponent;
-import javax.swing.JFrame;
 import javax.swing.JPopupMenu;
 import javax.swing.JTextField;
 import javax.swing.LookAndFeel;
@@ -46,6 +44,7 @@ import org.compiere.apps.search.InfoProduct;
 import org.compiere.model.GridField;
 import org.compiere.model.GridTab;
 import org.compiere.model.MAttributeSet;
+import org.compiere.model.MColumn;
 import org.compiere.model.MPAttributeLookup;
 import org.compiere.model.MProduct;
 import org.compiere.swing.CButton;
@@ -89,14 +88,14 @@ public class VPAttribute extends JComponent
 	{
 		/**
 		 *	Constructor
-		 *  @param adaptee adaptee
+		 *  @param attribute attribute
 		 */
-		VPAttribute_mouseAdapter(VPAttribute adaptee)
+		VPAttribute_mouseAdapter(VPAttribute attribute)
 		{
-			m_adaptee = adaptee;
+			this.attribute = attribute;
 		}	//	VPAttribute_mouseAdapter
 
-		private VPAttribute m_adaptee;
+		private VPAttribute attribute;
 
 		/**
 		 *	Mouse Listener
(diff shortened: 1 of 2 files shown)


client/src/org/compiere/grid/ed/VPAttribute.java
```diff
@@ -475,7 +475,7 @@ public void actionPerformed(ActionEvent e)
 			if (M_AttributeSet_ID != 0)
 			{
 				MAttributeSet mas = MAttributeSet.get(Env.getCtx(), M_AttributeSet_ID);
-				exclude = mas.excludeEntry(m_AD_Column_ID, Env.isSOTrx(Env.getCtx(), m_WindowNo));
+				exclude = mas.excludeEntry(product.get_Table_ID(), Env.isSOTrx(Env.getCtx(), m_WindowNo));
```
mckayERP, line 478: Suggest MColumn.getTable_ID(Env.getCtx(), m_AD_Column_ID, null) instead.  Product.get_Table_ID() will return the product table rather than the table associated with the column.

zkwebui/WEB-INF/src/org/adempiere/webui/editor/WPAttributeEditor.java
```diff
@@ -279,7 +279,7 @@ private void cmd_dialog()
 			if (M_AttributeSet_ID != 0)
 			{
 				MAttributeSet mas = MAttributeSet.get(Env.getCtx(), M_AttributeSet_ID);
-				exclude = mas.excludeEntry(m_AD_Column_ID, Env.isSOTrx(Env.getCtx(), m_WindowNo));
+				exclude = mas.excludeEntry(product.get_Table_ID(), Env.isSOTrx(Env.getCtx(), m_WindowNo));
```
mckayERP, line 282: Same.  Use MColumn.getTable_ID(Env.getCtx(), m_AD_Column_ID, null).

mckayERP (commented):
See comments in code.

client/src/org/compiere/grid/ed/VPAttribute.java
yamelsenih: Hi, maybe is better use I_M_Product.Table_ID constant

yamelsenih: Hi, maybe is better use I_M_Product.Table_ID constant

e-Evolution: I reviewed and implement other approaches, so that method singer change and now receipt only the table id

On the other hand, set the same logic that ZK UI to(boolean productWindow ).

Test executed :

1.- Product Window TShirt click in Attribute Set Instance field and show the Attribute Dialog
2.-Adding to Attribute Set Fertilizer Lot exclude the M_InOutLine table with IsSOTrx = 'N'
3.-Material Receipt Window add a line for Fertilizer#50 and click  Attribute Set Instance field and the  Attribute Dialog is not  open, this is ok
4.-Shipment Window add a line for Fertilizer#50 and click  Attribute Set Instance field and the  Attribute Dialog is open, this is ok

mckayERP: Still think the table is not the product table.  It should be the table of the tab using the VPAttribute dialog.  It is is used to check excludes by table, so it would need to be the C_OrderLine table on an order or the M_InOutLine table on a shipment receipt.

e-Evolution: Hi @mckayERP, you can see that this code only check if the column is the product window 

https://github.com/adempiere/adempiere/pull/3221/files#diff-31087b85b9e8a767119764d848dc08f7R474
Here you can see that now use the column id to solve of table

https://github.com/adempiere/adempiere/pull/3221/files#diff-31087b85b9e8a767119764d848dc08f7R485

marcalwestf:
as of now, **all calls** to _MAttributeSet.excludeEntry()_ are implemented using _table_ID_:
Selection_881

So the original argument has been tackled.

Now, the changes can be tested.

marcalwestf (approved):
- define an Attribute Set Exclusion (e.g. for Order Line)
- create an Order Line
- the Exclusion is applied

Test passed.

PR #3222, opened 2020-09-10
Fixed error with Sales Orders with cash as payment rule

yamelsenih:
## Step by Step:
- Create a Sales Order POS
- Complete it

A error is generated: No Cash Book

- Change System configurator flag "CASH_AS_PAYMENT" to Y
- Try one more time

A error is generated: No Cash Book

- Create a Bank As Cash
- Add a default account (same organization)
- Try one more time

A error is generated: No Cash Book

## What is the problem?
Always is validate Cash Journal instead

### DIFF
--- a/base/src/org/compiere/model/MInvoice.java
+++ b/base/src/org/compiere/model/MInvoice.java
@@ -1367,7 +1367,7 @@ public class MInvoice extends X_C_Invoice implements DocAction , DocumentReversa
 		}
 		//	No Cash Book
 		if (PAYMENTRULE_Cash.equals(getPaymentRule())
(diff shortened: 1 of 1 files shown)


marcalwestf (approved):
- Created a Sales Order as explained
- Complete this Sales Order
- The following documents were automatically created & correctly completed: Invoice, Shipment, Payment, Allocation.
- Bank account taken was the default defined in window Cash Journal Definition

Result: OK

marcalwestf:
Comment: this works OK when there is only **one** Cash Account and it is defined as default.
It is missing an enhancement for working with more than one Cash Accounts in a Cash Journal.

A suggestion would be to add a field in the document type the cash account to be taken for this document type can be defined.

PR #3232, opened 2020-10-01
Feature/#pf add overwrite definition column

yamelsenih:
This pull request allows overwrite a definition of report withou create a new column inside table, see how is the setup:

## Show original report:
Screenshot_20201001_164718

## Set flag Overwrite reference to true:
Screenshot_20201001_164128
Screenshot_20201001_164141

## See new report
Screenshot_20201001_164732

### DIFF
--- a/base/src/org/adempiere/print/export/PrintDataExcelExporter.java
+++ b/base/src/org/adempiere/print/export/PrintDataExcelExporter.java
@@ -73,7 +73,7 @@ extends AbstractExcelExporter
 			return null;
 
 		if (AD_Column_ID > 0)
-			obj = m_printData.getNode(Integer.valueOf(AD_Column_ID));
+			obj = m_printData.getNodeFromPrintFormatItem(item.getAD_PrintFormatItem_ID());
 		if (obj != null && obj instanceof PrintDataElement) {
 			return (PrintDataElement)obj;
 		}
--- a/base/src/org/compiere/model/I_AD_PrintFormatItem.java
+++ b/base/src/org/compiere/model/I_AD_PrintFormatItem.java
@@ -22,7 +22,7 @@ import org.compiere.util.KeyNamePair;
 
 /** Generated Interface for AD_PrintFormatItem
  *  @author Adempiere (generated) 
- *  @version Release 3.9.2
+ *  @version Release 3.9.3
  */
 public interface I_AD_PrintFormatItem 
 {
@@ -122,19 +122,6 @@ public interface I_AD_PrintFormatItem
 
 	public org.compiere.model.I_AD_PrintFormat getAD_PrintFormatChild() throws RuntimeException;
 
-    /** Column name AD_PrintFormatItem_ID */
-    public static final String COLUMNNAME_AD_PrintFormatItem_ID = "AD_PrintFormatItem_ID";
-
-	/** Set Print Format Item.
-	  * Item/Column in the Print format
-	  */
-	public void setAD_PrintFormatItem_ID (int AD_PrintFormatItem_ID);
-
-	/** Get Print Format Item.
-	  * Item/Column in the Print format
-	  */
-	public int getAD_PrintFormatItem_ID();
-
     /** Column name AD_PrintFormat_ID */
     public static final String COLUMNNAME_AD_PrintFormat_ID = "AD_PrintFormat_ID";
 
@@ -150,6 +137,19 @@ public interface I_AD_PrintFormatItem
 
 	public org.compiere.model.I_AD_PrintFormat getAD_PrintFormat() throws RuntimeException;
 
+    /** Column name AD_PrintFormatItem_ID */
+    public static final String COLUMNNAME_AD_PrintFormatItem_ID = "AD_PrintFormatItem_ID";
+
+	/** Set Print Format Item.
+	  * Item/Column in the Print format
+	  */
+	public void setAD_PrintFormatItem_ID (int AD_PrintFormatItem_ID);
+
+	/** Get Print Format Item.
+	  * Item/Column in the Print format
+	  */
+	public int getAD_PrintFormatItem_ID();
+
     /** Column name AD_PrintGraph_ID */
     public static final String COLUMNNAME_AD_PrintGraph_ID = "AD_PrintGraph_ID";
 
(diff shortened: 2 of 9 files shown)


e-Evolution:
@yamelsenih Yamel

I think this functionality is redundant, since today you can do that in the table or change the report view with the reference you want, so for me this functionality duplicates an existing functionality.

image

image


On the other hand, the most critical and serious part is that it opens a back door for any user with the minimum access profile, for example Garden User to add any query that promises any information from the database, which is very delicate.

That is the reason that any definition of how and that is shown are made from the application dictionary with which you require access from the System Administrator to allow creating a reference, modifying a table, or creating a view.

This feature bypasses this security mechanism for information access and more because any user can copy and modify a printing format, even with the minimum profile.

in summary this functionality has a lot these disadvantages

1.- Duplicity since there is a functionality that allows solving the same requirement without exposing security.

2.- Not using the definition of the column for virtual columns or the report view reflects total ignorance of the current operation of the application dictionary and security in adempiere since any information would be exposed.

3.- It opens a back door by allowing a user to add a query from a printing format with the minimum level of security, opening a back door that exposes the data of the database for any user.

4.- It breaks the security consistency since currently only users with a system administrator profile have the possibility to change and modify any definition that is shown and how it is shown in any part of the application

yamelsenih:
Hi @e-Evolution my reply

>1.- Duplicity since there is a functionality that allows solving the same requirement without exposing security
I don't sure, when you have many and distinct verticals is need allows many changes without change core functionalities

>2.- Not using the definition of the column for virtual columns or the report view reflects total ignorance of the current operation of the application dictionary and security in adempiere since any information would be exposed.
I can hide column

>3.- It opens a back door by allowing a user to add a query from a printing format with the minimum level of security, opening a back door that exposes the data of the database for any user.
Resolved with previous response

>4.- It breaks the security consistency since currently only users with a system administrator profile have the possibility to change and modify any definition that is shown and how it is shown in any part of the application
Resolved with previous response

e-Evolution:
> Hi @e-Evolution my reply
> 
> > 1.- Duplicity since there is a functionality that allows solving the same requirement without exposing security
> > I don't sure, when you have many and distinct verticals is need allows many changes without change core functionalities

So the solution is to create a new view or report view for your vertical, and not duplicate existing functionality

> > 2.- Not using the definition of the column for virtual columns or the report view reflects total ignorance of the current operation of the application dictionary and security in adempiere since any information would be exposed.
> > I can hide column

You simply should not include this possibility, it should not be allowed to expose the database data by an user from a print format in any case without this being allowed by a system administrator,

> > 3.- It opens a back door by allowing a user to add a query from a printing format with the minimum level of security, opening a back door that exposes the data of the database for any user.
> > Resolved with previous response
> 
> > 4.- It breaks the security consistency since currently only users with a system administrator profile have the possibility to change and modify any definition that is shown and how it is shown in any part of the application
> > Resolved with previous response

piracio:
Hi, I think we have a few problems here.
1) SQL Injection ( try on your lab select name, password from ad_user)
2) Subqueries are performance killers. (trust me I am a DBA Oracle/Postgresql for VLDB)
3) what problem you want to resolve, perhaps the issue here is to learn how to use the current functionality

yamelsenih:
> Hi, I think we have a few problems here.
> 
> 1. SQL Injection ( try on your lab select name, password from ad_user)
> 2. Subqueries are performance killers. (trust me I am a DBA Oracle/Postgresql for VLDB)
> 3. what problem you want to resolve, perhaps the issue here is to learn how to use the current functionality

Hi @piracio here my reply
> 1. SQL Injection ( try on your lab select name, password from ad_user)
All new columns are hidden and is not a problem because if a implementer want implement it is own your risk
> 2. Subqueries are performance killers. (trust me I am a DBA Oracle/Postgresql for VLDB)
This not apply because already exist a definition using this model, see ti:
Screenshot_20201205_185848
Is realy that a sub-query can be killer but is not a reason for reject it, also you can see that exist things like **DISTINTS** (Used for Smart Browser) or **SELECT** inside a **IN** clause but they never will be removed from ADempiere
> 3. what problem you want to resolve, perhaps the issue here is to learn how to use the current functionality
Just I want implement a way more extended that can allow add more information for any print format, now just is possible from Report View but you can have manu customer with one Server side using differents tenants and they can have many requirements distints


I just want improve functionality for this, The security level is keeped because all columns added are hidden by default, if some want implement it then need change dictionary before implement

e-Evolution:
Hello @yamelsenih  Yamel, @piracio Horacio, Smart browsers define how the model is shown in the UI, but they do not define the model, they work with the same paradigm of windows, so that information can be displayed in a window it has to be first defined in the model.

A Smart browser cannot directly add an SQL to show any data, for this you have to make a Smart View. This is where you define how to access or combine the models to access the data (a dynamic query that you do not need to create a view in the database).

Print formats work in a similar way to a window or smart browser, their responsibility is to define how the data is displayed, but they do not define the model, this responsibility is defined in a table or a report view, as well as the smart browsers. Print formats are only responsible for presenting the information and not for defining the template.

Thus, in good best practice Design patterns, the Single-responsibility principle https://en.wikipedia.org/wiki/Single-responsibility_principle,  defines that a component or function has only one responsibility, in this case, it includes the responsibility to define the model or how it is combined with the tables and not how is show

The responsibility of how the information is displayed is of the windows, smart browser, and printing formats.

The moment you add the possibility to define how the model is accessed and combined in a window, smart browser, or print format, you are violating the principle of responsibility, and then all this turns into chaos.

Just because someone doesn't want to take the time and hassle to define in the table or smart view. What is the problem with the current approach, have you any argument vs?

On the other hand, The components of the application dictionary are subordinate to the system administrator role, which requires a user to have this role in order to make any changes or combinations of the model.

that does not happen for smart browsers or printing formats, since any user today can modify a smart browser or a printing format in a standard way

For this reason, even though this feature may be attractive, it is not worth breaking the liability principle and opening a back door, even if it is at your own risk, for anyone to steal or extract data.

today exists a way to define the model that meets any need and good design, only need setting the place right

So on the other hand, if you want to take the risk, you can do it with your fork.

kind regards

marcalwestf:
@yamelsenih  , @e-Evolution  what do we do here?
Is there a possibility between not accepting and doing some changes to accept it?

Please, write some suggestions here.

yamelsenih:
Hi @marcalwestf my change just allows create virtual fields for print formats and if you accepts ti just change the print format engine for allows it, however not represent a problem and security bug because all columns added was keeped as not displayed and the only way for use this functionality is that a implementor enable this.

I think that it can represent a risk if is enabled by default but you can apply all XML's and can watch the print format item that not have changes because all columns is not displayed

Best regards

e-Evolution:
Hi @marcalwestf Mario , @yamelsenih Yamel!

As I said before, it is not good that ADempiere has hidden back doors it is a bad practice in any system where it is required to have the secure information that is the case in a business application.

On the other hand, the system already has a way to create calculated fields safely using the system administrator.

The only interesting thing about this functionality is the possibility of overloading the reference to show a different value to the one configured within the column of a table in a similar way as it happens now with windows, but this is sure because to do this the reference even id can only be built as a System Administrator.

If the possibility of extracting information by any user of the system is removed and only the possibility of overwriting the reference is left, this improvement can be accepted.

adaxa-steven:
Hi All,
I agree with Victor ... too much risk of exposing information that is not supposed to be visible to some role or user.
There are any many times I would like to have added a column twice in a print format so on the second instance I could display record_id or Value instead of what was displayed by default.. but its not worth the risk ... just have to put up with the small amount of work for System Administrtor role to add a virtual column to the table just so we can print it.
My 2 cents...
regards
steven

yamelsenih:
Thanks, no problem. I can propose a better solution on next pull request.

Best regards

Issue #3236, opened 2020-10-12
[Bug Report] Workflow Engine


yamelsenih:
## Bug report
Workflow approval throw a error

#### Steps to reproduce
1. Login on Garden World as GardenUser
2. Create a Purchase Requisition
3. Add a line with amount less of 500
4. Complete it: Note that is suspended instead completed (Waiting for Approval)
5. Login with GardenAdmin
6. Go to Workflow approval
7. Try approve the WF: Thoew a error, see log:

```Java
===========> WFActivity.cmd_OK: Please Approve Requisition [18]
org.adempiere.exceptions.AdempiereException:  Transaction Workflow * Not found *
	at org.compiere.wf.MWFProcess.getWorkflowProcessTransaction(MWFProcess.java:615)
	at org.compiere.wf.MWFActivity.run(MWFActivity.java:756)
	at org.compiere.wf.MWFProcess.startNext(MWFProcess.java:341)
	at org.compiere.wf.MWFProcess.checkActivities(MWFProcess.java:256)
	at org.compiere.wf.MWFActivity.setWFState(MWFActivity.java:282)
	at org.compiere.wf.MWFActivity.setUserChoice(MWFActivity.java:1240)
	at org.compiere.apps.wf.WFActivity.cmd_OK(WFActivity.java:779)
	at org.compiere.apps.wf.WFActivity.actionPerformed(WFActivity.java:642)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
	at java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:289)
	at java.awt.Component.processMouseEvent(Component.java:6535)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
	at java.awt.Component.processEvent(Component.java:6300)
	at java.awt.Container.processEvent(Container.java:2236)
	at java.awt.Component.dispatchEventImpl(Component.java:4891)
	at java.awt.Container.dispatchEventImpl(Container.java:2294)
[...]

PR #3237, opened 2020-10-12
Fixed error with transaction for Workflow approval

yamelsenih:
Fixes: #3236


marcalwestf (approved):
- steps were done as described in issue #3237 without pull request changes
- result: an exception was thrown
- same steps done with pull request changes
- result: workflow could be approved and document completed without throwing an exception

Test passed.

PR #3238, opened 2020-10-19
Add support to Web-Store integration and use a implementation with

yamelsenih:
This pull request have all related to minimal use case for Web Store Production use based on [Store-Front](https://github.com/DivanteLtd/storefront-api)

Based on my issue put over project https://github.com/DivanteLtd/storefront-api/issues/95 this have model for minimal use of store:

## Web Store Window
  - Add Cache Server Provider: use for get all connection parameters of external cache used for product catalog, tax, attributes and others
  - Template BPartner: The same business logic of POS it is used as template for new business partner
  - Customer Role: a default role for allows login with a user after register inside ADempiere
Screenshot_20201019_153019

## Web Store Definition
A folder with all need for setup a complete store using Payment Methods, Delivery Via Rules and Product Group
Screenshot_20201019_153525

## Basket: 
A simple basket for save the user selection from store. A basket can be linked to many packages depending of delivery via rule and freight cost rule
Screenshot_20201019_153637
## Package: 
Many packages can be generated automatically from a basket for define freight cost and picking strategy
## Product Group: 
A classification used for web store without affect the general classification of product, also it have a hierarchy
Screenshot_20201019_154303
## Payment Method: 
Used for define distints payment methods:
  - Cash On Delivery
  - Debit and other
Screenshot_20201019_154512
## Export Catalog to Cache Server Provider: 
A batch process for export all catalog from ADempiere to cache server provider
Screenshot_20201019_154804

## Business Partner changes: 
Just add three columns for define bulling and shipping location by default and name 2 used for Store
Screenshot_20201019_155059

Note: The default implementation is done for Elastic Search

This pull request just create the required fields and setup for use ADempiere with a Web Store

For more info please read this document: https://docs.storefrontapi.com/guide/general/introduction.html


e-Evolution:
@yamelsenih Here my feedback about the this integration.

About the new field into the ADempiere Web Store
The Cache Server Provider is OK 
The Customer Role is Ok
The Templete Business Partner OK

Basket new entity model, I'm not sure if another model is required for the Basket, since we already have the sales order, you would lose the integration of discounts, prices, promotions, kit management, etc. .

image


On the other hand, synchronizing the Basket with the sales order is impractical if you are already doing a native integration, the current web store has direct integration with the sales order, it is preferable to increase the functionality of the sales order to support the Basket.

On the other hand ADempiere already has a way to manage the packages, we do not require a new form, if necessary we must increase the current functionality not duplicate it.

image


In conclusion Basket new Entity not is necessary 
In conclusion Packages new Entity not is necessary 


The Product Group:

I consider that adding another data entity to group the categories of the products for the web store is to duplicate the models, this is not correct since the integration with the current product functionality would be lost to forecast management, price management, promotions, commission payment, reporting sales, tax management.

On the other hand, ADempiere already has a way to segregate the products by different taxonomies and a way to identify which products can be sold through the electronic commerce platform

image

The field Seld-Service is use to separate the product that are show in web store ,and this exist now in Product Category and Product Info 

So it is unnecessary to add another model when Adempiere already has it contemplated

image


So it is unnecessary to add another model when ADempiere already has it considered.

Here you can enable the characteristics of the product and if the product can be sold through a web store, it is not necessary to further complicate the model.


Export Catalog to Cache Server Provider: I think that is ok that exist un process that export to cache server , but I think that should be more transparent , when the server is up and a web store is enable the synchronization should be automatic , and not in manual processor the export.

In conclusion , is ok to export process , but should be automatic when server is startup.




In conclusion the Product Group new Entity not is necessary 

Payment Method: currently ADempiere only support this using the Payment Rule or Tender Type , so create a new entity that are link the tender type is ok , I think that is necessary this new entity.

In conclusion the Payment Method new Entity is Ok


Business Partner changes: this change not is necessary because we have the flags that indicate if an address is to ship to or billing to , I not considered that is necessary adding default flags 

Sin título

Business Partner changes: this change is unnecessary because we have the flags that indicate if an address is to ship to or invoice, So adding a new flag to standard is only duplicate the model

Issue #3239, opened 2020-10-26
When the MRequest send notices, this mail caches the previous mail context


rcapecce:
## Bug report
When the MRequest is saved, this send a notification, this notification may be has a lot of context vaiables. If I send a request notification when in the Request hasn't a value in the field (the field is null) and in the previous notification, the field had a value, the notification will send that outdated value.

#### Steps to reproduce
1. Open a completed customer invoices and make a new Request (from the airplane in the ribbon)
2. The (not yet saved) Request will have the Invoice, complete the mandatory fields and save
3. This send a notification template with the Invoice Number
4. Open a Media Project and make a new Request
5. The Request will have the Project but **not the Invoice**, complete the mandatory fields and save
6. The template will be loaded with the project value and the outdated Invoice value (in this Request the field was null).

#### The template used
The template used is something like this
MailText2 field:

```
Assigned to: @AD_User.Name@
Date Next Action: @DateNextAction@
Project @C_Project.Value@
Sales Order @C_Order.DocumentNo@
Invoice @C_Invoice.DocumentNo@
```

#### Expected behavior
The MMailText needs a clear option between Request

PR #3240, opened 2020-10-26
Adding a clear method to empty the previous context in the MMailText

rcapecce:
Fixes: #3239

### DIFF
--- a/org.adempiere.request/src/main/java/org/compiere/model/MMailText.java
+++ b/org.adempiere.request/src/main/java/org/compiere/model/MMailText.java
@@ -458,4 +458,11 @@ public class MMailText extends X_R_MailText
 	public String toString() {
 		return "MMailText [getMailHeader()=" + getMailHeader() + ", getR_MailText_ID()=" + getR_MailText_ID() + "]";
 	}
+
+	/**
+	 * Clear the entityMap from the previous context
+	 */
+	public void clear() {
+		entityMap.clear();
+	}
 }	//	MMailText
--- a/org.adempiere.request/src/main/java/org/compiere/model/MRequest.java
+++ b/org.adempiere.request/src/main/java/org/compiere/model/MRequest.java
@@ -1044,6 +1044,7 @@ public class MRequest extends X_R_Request
 		if(!Util.isEmpty(eventType)) {
 			MMailText mailText = MRNoticeTemplate.getMailTemplate(getCtx(), MRNoticeTemplate.TEMPLATETYPE_Request, eventType);
 			if(mailText != null) {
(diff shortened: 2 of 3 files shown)


e-Evolution:
@rcapecce @eduardolopezuy  I can understand the problem and the solution is close, but I consider that the map with the instances should be cleaned for an execution cycle in a defined context after sending the email, in this way we are sure not to have
memory leak or have a map cleaned before sending the mail.

Therefore, in all dependencies where MMailText.setPO is called, the map must be cleaned up after sending the mail.

kind regards
Victor

e-Evolution:
Reviewing this issue again and the solution that arises is not complete.

The error is caused by an incorrect definition of the cache here my more detailed analysis:

The messages are brought from the run templates but they are in a cache that could be loaded at some point.
 https://github.com/adempiere/adempiere/blob/e3f880b6b8608f618791dec2dc205c9a74e78467/org.adempiere.request/src/main/java/org/compiere/model/MRequest.java#L1045

On the other hand, as this is an instance that is already in cache or in memory, when you add an entity to add the answer of that entity using this method

https://github.com/adempiere/adempiere/blob/e3f880b6b8608f618791dec2dc205c9a74e78467/org.adempiere.request/src/main/java/org/compiere/model/MMailText.java#L310

The entityMap can now also bring information from past or accumulated executions, so this cache is poorly defined since it can bring stale information.


While adding these lines of code may fix the problem for this case, it doesn't prevent it from happening in more places.


mailText.clear (); // This can be a cached object with a previous context that may be messing up the data

It seems to me that using a cache to obtain an MMailText without guaranteeing that the context is cleaned is a bad design of the cache itself since entityMap is done in an execution cycle when sending an email from MRequest.


perhaps it is more appropriate to generate an instance each time the MMailText is required rather than explicitly controlling this and letting the garbage collector do its job.

Since it is not clear when it is correct or not to clear the cache, this should be when the MMailText instance is created

I suggest that you review it together with @yamelsenih  Yamel since he was the one who implemented this improvement and look for another approach since the current one does not have a correct design.

kind regards
Victor

rcapecce:
Hello @e-Evolution and @yamelsenih.
I fix this PR with your comment, moving the clear template method to the MRNoticeTemplate.getMailTemplate(...)

This is working the same as the previous commit, clearing the cache for the following uses.

I leave here the tests that I did before and after the PR.

In the first test you can see how a value is set in the R_Group_ID field and the template takes it, but when it is set to null, the template remembers the previous value.

Peek 14-04-2021 17-35

In the second test (with the source of this pull request) you can see how the value is set in the R_Group_ID field and the template takes it, and, when it is set to null, the template set to null the value too.

Peek 14-04-2021 17-42

PR #3251, opened 2020-11-19
Add Third Party Access support for Token based on default Token

yamelsenih:
## What is the improve?
This pull request allows define a **Token Type** for **Third Party Access** very similar to GirHub / GitLab and other serious applications. 

With this change can exists interfaces for login based on temporal token instead **user** and **password**

## Here a reference of scope for this: 
- Gihub Token: https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token
- https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html
- MD5 by default
- Process for Generate Token
- Process For Revoke Token

## Menu location
Screenshot_20201118_202450

## A example of use:
Peek 2020-11-18 20-37

### DIFF
--- a/base/src/org/spin/model/I_AD_Token.java
+++ b/base/src/org/spin/model/I_AD_Token.java
@@ -23,7 +23,7 @@ import org.compiere.util.KeyNamePair;
 
 /** Generated Interface for AD_Token
  *  @author Adempiere (generated) 
- *  @version Release 3.9.2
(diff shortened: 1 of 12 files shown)


e-Evolution (commented):
@yamelsenih can please move the Query to method this way we can reuse the query for other functionality

PR #3252, opened 2020-11-19
Fixed Memmory Consumption for HTML reports

yamelsenih:
## The problem
when a reporte is generated as **html** format it is loaded completely in memory as **html** object before write to file, a complete file like generated from **Fact Acct Detail** can have many records and a example for a little customer can load at least **1.5 GB**

This pull reques just write the file instead load it completely in memory.

A little change but save many RAM use.

TODO: Ins nice to have pagination for **PrintData** object, it can help to improve speed and memory use for each report running


marcalwestf (approved):
- code checked
- exported as HTML file

Test passed.

PR #3253, opened 2020-11-19
Just resize image for reduce memory consumption

yamelsenih:
Resize image before load for improve memory consumption


marcalwestf (approved):
- loaded a very heavy image into a report
- run the report
- the image was loaded correctly in the report
- the image was downsized accordingly to the print element defined
- the resulting document was lighter than otherwise

Test passed.

PR #3254, opened 2020-11-19
Add support to Corporate Branding Image for Organization Info

yamelsenih:
Minor change for add price checking image for organization:
Screenshot_20201119_104415

This can be used for load some like this: https://docs.erpya.com/es/4.0.0/adempiere/quote-to-invoice/consultation-of-product-prices.html

### DIFF
--- a/base/src/org/compiere/model/I_AD_Org.java
+++ b/base/src/org/compiere/model/I_AD_Org.java
@@ -22,7 +22,7 @@ import org.compiere.util.KeyNamePair;
 
 /** Generated Interface for AD_Org
  *  @author Adempiere (generated) 
- *  @version Release 3.9.2
+ *  @version Release 3.9.3
  */
 public interface I_AD_Org 
 {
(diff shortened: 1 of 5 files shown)


e-Evolution:
@yamelsenih  I think it is not the correct approach to add an image with a very specific context, the correct approach would be to use the approach of how images are now stored for the system. as it is currently done, this should be one more or the ones you require.

image

yamelsenih:
Screenshot_20201119_144434

e-Evolution:
@yamelsenih thank you, is right the change of name because as we chat is more a branding corporation

yamelsenih:
Thanks for recomendation

adaxa-steven:
for extra branding options is it worth also adding 'Logo Web Header' to Client info?
image

and phone, fax, bank accounts etc ... which can then be used in the default print format for invoices and POs so that a new client can immediately create an invoice and/or a PO with all the fields that tehy will likely need to have on their official documents?
image

e-Evolution:
@adaxa-steven , we have the contribution of ADAXA in the last version of ADempiere.

image

we have not the last fields Logo web Header and ZK Home page, what is the functionality?

About the Organization Info we have synchronized with ADAXA version :

image

Now @yamelsenih adding a new corporate image logo to show in some functionality like query price in the retail industry is necessary setting by organizatión because each organización can management your corporation branding

adaxa-steven:
Hi Victor,
I should have checked about the extra fields on Org .. sorry for confusion.
I only looked at Yamel's screen shot
image

the last image gets displayed here..
image

it allows extra branding and is also good for setting a different image in test and dev versions to what is in production ... i think there is also code there to allow you to change the screen background colour to make it clear that you are in a test/dev system.

I can't explain the zk Home page well. I believe it is used so that a person logging into the system can be directed to the desired location by some proxy server if you are running a multi tenant type system with each Org being a separate company. 

regards
steven

PR #3259, opened 2020-11-22
Add support to tender type for bank transfer

yamelsenih:
## Feature
Just add support to tender type (optional) for bank transfer

Screenshot_20201121_223542


adaxa-steven:
is there some reason not to allow the amount to be specified in the receiving bank account?
any exchange diff or fees can then be expensed to clear out any remaining balance on the transfer still in the bank transfer account.
image

PR #3261, opened 2020-11-23
Add ID for Product Price

yamelsenih:
Just add **M_ProductPrice_ID** for **M_ProductPrice** table


marcalwestf (commented):
Hi @yamelsenih 
- XML  loaded correctly
- created one entry in PriceList
- the new entry's record ID is 1000000, the same as one record updated with the migration file

Selection_940


This would mean that if all existing M_ProductPrice records were set by the migration file, there would be as many repetitions.

I wonder if it is correct to have two records having the same ID?

e-Evolution:
the same comment that #3262

PR #3262, opened 2020-11-23
Add ID for M_Storage Table

yamelsenih:
Juas add **M_Storage_ID** column for **M_Storage** table

### DIFF
--- a/base/src/org/compiere/model/MStorage.java
+++ b/base/src/org/compiere/model/MStorage.java
@@ -29,13 +29,14 @@ import org.compiere.util.CLogMgt;
 import org.compiere.util.CLogger;
 import org.compiere.util.DB;
 import org.compiere.util.Env;
-import org.compiere.util.Util;
 
 /**
  * 	Inventory Storage Model
  *
  *	@author Jorg Janke
  *	@version $Id: MStorage.java,v 1.3 2006/07/30 00:51:05 jjanke Exp $
+ *	@author Yamel Senih, ysenih@erpya.com, ERPCyA http://www.erpya.com 
+ *	Add support to Storage ID
  */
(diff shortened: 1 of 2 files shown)


mckayERP (changes requested):
Not sure you should remove the parent keys without making those same fields mandatory.  There is no sense in creating storage records without a product, locator or attribute set instance.    I'd be OK with the change is theses mandatory settings were made.

mckayERP (approved):

marcalwestf (commented):
I always wondered why M_Storage (and M_Cost by the way) nerver had from the beginning an ID. 

Of course, if Product, ASI and Locator are the PK, an extra ID becomes at best redundant.

My question here @yamelsenih is if 

- M_Storage_ID will be unique
- in case M_Storage_ID would be unique, will there be any potential conflict when somebody uses the ID instead of the compounf PK?

e-Evolution:
@marcalwestf  Tables with multiple keys are really a Value Object https://martinfowler.com/bliki/ValueObject.html, https://deviq.com/domain-driven-design/value-object , they cannot be considered as an entity https://deviq.com/domain-driven-design/entity, since they are composed of the combination of multiple keys.

This is the same for data like M_Storage, M_Location, C_Location, etc.

So adding an ID can be confusing, on the other hand handling the queries with unique keys is a bit complex, 


Will be  know, @yamelsenih  What is the intention of adding an ID when there is already a UUID that is a surrogate key designed for the sole purpose of replication?

mckayERP:
Whether its a constraint on the table (parent keys) or some beforeSave logic, the storage records should be restricted to a single records per product, ASI, locator combination.  This is a business logic constraint - the immutable value object if you will.

PR #3281, opened 2020-12-14
Feature/#3208 add junit jupiter

mckayERP:
Add junit5 juniper support to Adempiere tests and build process.  The test can be developed in accordance with standard practice in junit juniper and run in the IDE and during the build process.   Two types of tests are possible: unit tests which do not require any interaction with a database and integration tests which do.  For integration tests, the IDE must be able to launch the application - meaning the properties files and database must be setup and the database has an import of the current seed data.

For places where the integration tests are not possible, for example during the first build without an installed version of the application, the file test.properties in the util_dev directory can be used to control which tests are performed as part of the build.

A test execution listener is added to provide a summary of the test output.  Detailed test reports can be found in base/test/test_reports following build tests.  These reports are compatible with the junit report reader in the IDE.

This pull requests only updates the base build process.  As tests are added to other modules, the build process in those modules should be updated with the equivalent changes.

Existing tests in the software were updated to junit5 and work with the build process.  Both unit and integration tests exist.  The one test not modified was the email test as this referenced a particular email server.  In this case, the test was commented out.

I have a number of older contributions to make that include junit tests and I'd like to include the tests as part of those contributions.  I hope this change is acceptable as it would allow me to make the contributions easily and give the team greater confidence in the contributed code.


mckayERP:
The travis build is failing because the Ant version used is old and doesn't have the junitlauncher required to run the tests.  We may need to update it.  From stackoverflow https://stackoverflow.com/questions/31392245/what-version-of-ant-comes-with-travis-ci

```
 before_install:
    - wget --no-check-certificate https://www.apache.org/dist/ant/binaries/apache-ant-1.10.9-bin.tar.gz
    - tar -xzvf apache-ant-1.10.9-bin.tar.gz
    - export PATH=`pwd`/apache-ant-1.10.9/bin:$PATH
    - echo $(ant -version)
```

mckayERP:
Updated the travis.yml to install the correct version of Ant.  Now there is an error in the directory path which is different in travis than in my development environment.  I'll have to find the correct way to specify the path in the junitlauncher and try again.

e-Evolution:
Hi Mike, great job.

I think like you, libraries like many of the ones we have are obsolete so updating is an excellent idea.

Review the contribution and I have the following questions.

1.- This contribution resolves the conflict between libraries because there are currently three versions of junit:
./tools/lib/junit.jar
./zkwebui/WEB-INF/lib/junit.jar
./org.adempiere.webservice/WEB-INF/lib/junit-3.8.1.jar

Can we have a single version that does not generate conflicts?

On the other hand, I see in your contribution modifications are made to handle the cache and transactions for some classes in the core

base / src / org / compiere / model / MPeriod.java

Can you isolate this change in another pull request, since I understand that you are solving a concept other than the test environment?

A good practice is not to mix within a pull request solution for two different problems, a pull request solves only one problem

mckayERP:
Hi Victor, 

Will do.  I was planning on making a few pull requests related to this but I guess I left too much in the utilities.  I'll pull that out and get rid of the other Junit libraries. 

Thanks for the comments.

mckayERP:
@e-Evolution That should do it.  There were a lot of tests hidden in the main code which I've pulled out into test source packages.  The tests are split into integration tests which have a IT_<ClassName> pattern to their names and unit tests which are typically Test<ClassName>.  The integration tests are not run during the travis checks but can be easily during local builds by changing one property in the utils-dev/test.properties file.   The tests that exist here are far from complete but its a good start.  I did have to make a few changes to some model classes so model.get() methods will find records created in the same transaction.  Let me know if there are any comments.

e-Evolution:
@mckayERP , I tried build the project using ./utils_dev/RUN_build.sh and got this errors, may be need update ant libraries, when I use last version ant from OS this work, please can you fix? :

clean.tests:
   [delete] Deleting directory /Users/e-Evolution/Develop/ADempiere/develop/base/test/build
   [delete] Deleting directory /Users/e-Evolution/Develop/ADempiere/develop/base/test/test_results

compile.test:
    [mkdir] Created dir: /Users/e-Evolution/Develop/ADempiere/develop/base/test/build
    [javac] Compiling 66 source files to /Users/e-Evolution/Develop/ADempiere/develop/base/test/build

alertIfUnitTestSkipped:

unit-tests:
    [mkdir] Created dir: /Users/e-Evolution/Develop/ADempiere/develop/base/test/test_results

BUILD FAILED
/Users/e-Evolution/Develop/ADempiere/develop/utils_dev/build.xml:21: The following error occurred while executing this line:
/Users/e-Evolution/Develop/ADempiere/develop/base/build.xml:218: Problem: failed to create task or type junitlauncher
Cause: the class org.apache.tools.ant.taskdefs.optional.junitlauncher.confined.JUnitLauncherTask was not found.
        This looks like one of Ant's optional components.
Action: Check that the appropriate optional JAR exists in
        -/Users/e-Evolution/Develop/ADempiere/develop/utils_dev/./lib
        -/Users/e-Evolution/.ant/lib
        -a directory added on the command line with the -lib argument

Do not panic, this is a common problem.
The commonest cause is a missing JAR.

This is not a bug; it is a configuration problem

e-Evolution:
@mckayERP  Can we use a single directory org.adempiere.test/src/test/java/org/adempiere/test to put here all ADempiere tests, instead of creating a directory for each directory package? 

What do you think?

mckayERP:
@e-Evolution Thanks for the feedback - I'll see what the configuration issue is.  

For the directories there are a number of strategies but the basic tenet is that the tests require package accesses so they have to be in the same package as the class under test.

We can

1.  Include tests in the same source folder as the class under test and change the naming scheme so that it is always

    className
    classNameIT  (for integration tests)
    classNameTest (for unit tests)

2. Use the IDE which allows us to specify that a source folder contains "tests".  I chose this approach.

In the first case, the test code needs to be excluded from the jars so the Ant builds would have to ignore class files that end with the name test.  It has the advantage that the test and the class under test are in the same place and appear in alphabetical order.

In the second case, the test code is easily excluded from the jars.   I used a test source folder per module as I thought the individual modules could still be opened as projects so it was better to include the tests within their own project.   The tests for a class would be in the test src directory under the same package.  I used the name scheme IT_ClassName for integration tests and TestClassName for unit tests.

If we want to completely abandon the sub-project approach, we could move all the tests to a single test source directory. On the other hand, the sub-project approach still has merit - especially if we move to maven or some other dependency management system.  Then a developer could work within a single project without having to "open" everything.

mckayERP:
For the error you experienced, the junitlauncher requires the latest versions of ANT - at least 1.10.6 - but I'm using 1.10.8 and that is what I set the travis build to use as well.

mckayERP:
@e-Evolution 
Hi Victor, I missed the RUN_Build scripts.  Hopefully the last commit will fix it.  I added the necessary ANT libraries and updated the build scripts accordingly.  RUN_Build.bat and RUN_Build2.bat work on my machine.  Can you try the .sh versions?

Any further thoughts on the directories?

Thanks.

mckayERP:
@e-Evolution @yamelsenih @marcalwestf 
Note - the target of this PR is a feature branch.  Can we create the branch in the Adempiere repo?  I have a number of other changes that build off of it.

e-Evolution:
@mckayERP Hi Mike!

I am waiting to resolve some merges like #3308 that were made where I disagree with the changes before doing merge that is already tested.

It is necessary to try to merge the changes in the order that the pull requests were created to avoid resending the changes since they will be out of sync with the last changes.

So as long as we can resolve the discrepancies to make the merge of these changes

kind regards
Victor

mckayERP:
@e-Evolution 
Ok, Thanks.  This is a feature branch so should be safe until merged with development.  Its up to date with development branch as of two days ago and I'll keep it in sync as development changes so an eventual merge will be easy.

e-Evolution:
@mckayERP great work , thank a lot.

PR #3285, opened 2020-12-17
Fixed error with Reversed document when it is copied from window

yamelsenih:
# Step by Step
- Create a Payment
- Reverse created payment
- Copy the reversed payment and complete it

# What is the bug? 
Sometimes the **Reversal_ID** is flagged as **IsAllowsCopy** on true for window, then is copied for new payment, it is a error because some validator have **Reversal_ID** as value for determinate if is a voided document or not

# What is the consequenses
Is a document type is flagged as **IsReversedWithOriginalAcct** as true then the account fact engine sear a reversal document for create fact and is wrong

# A sample
Screenshot_20201216_221825

Peek 2020-12-16 22-17

### DIFF
--- a/base/src/org/compiere/model/M_Element.java
+++ b/base/src/org/compiere/model/M_Element.java
@@ -346,22 +346,37 @@ public class M_Element extends X_AD_Element
 		//	its must be change for dynamic dictionary query
 		return columnName.equals("AD_Client_ID")
 			//
-			|| columnName.startsWith("Created") || columnName.startsWith("Updated")
-			|| columnName.equals("EntityType") || columnName.equals("DocumentNo")
-			|| columnName.equals("Processed") || columnName.equals("IsSelfService")
-			|| columnName.equals("DocAction") || columnName.equals("DocStatus")
-			|| columnName.equals("Posted") || columnName.equals("IsReconciled")
+			|| columnName.startsWith("Created") 
+			|| columnName.startsWith("Updated")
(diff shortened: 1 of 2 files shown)


e-Evolution:
@yamelsenih Hi Yamel, this is only part of the solution, is necessary also the change in all fields in all window

e-Evolution:
@yamelsenih I reviewed and found that all these fields are not synchronized with the columns, so that   even when the column IsAllowCopy is setting as 'N', the field is setting as 'Y'
```
SELECT w.Name AS "Window", t.Name AS "Tab" , f.Name AS "Field" , f.IsAllowCopy , c.ColumnName , c.IsAllowCopy  FROM AD_Field f 
INNER JOIN AD_Column c ON (f.AD_Column_ID = c.AD_Column_ID)
INNER JOIN AD_Tab t ON (f.AD_Tab_ID =  t.AD_Tab_ID)
INNER JOIN AD_Window w ON (t.AD_Window_ID = w.AD_Window_ID)
WHERE f.IsAllowCopy <> c.IsAllowCopy AND c.IsAllowCopy = 'N';
```

yamelsenih:
Hi @e-Evolution the query apply changes for fields and columns you can check it here: https://github.com/adempiere/adempiere/pull/3285/commits/3a0e49da0b29844871f3854d87040a38f2ea1bab#diff-01ba76f25c38ad40688529bd75485974316f16c3fc53725c1e7c6bd83490dce4R5

PR #3286, opened 2020-12-17
Add backward compatibility for iReport parameters

yamelsenih:
# Current problem
When a report based on iReport have **AD_Org_ID** parameter exist a error
0014_Listado_Productors_Activos_Formato_Sencillo_393

```XML
	<parameter name="AD_Org_ID" class="java.math.BigDecimal">
		<defaultValueExpression><![CDATA[]]></defaultValueExpression>
	</parameter>
```

### DIFF
--- a/JasperReports/src/org/compiere/report/ReportStarter.java
+++ b/JasperReports/src/org/compiere/report/ReportStarter.java
@@ -39,6 +39,7 @@ import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.PropertyResourceBundle;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


JasperReports/src/org/compiere/report/ReportStarter.java
```diff
@@ -519,26 +520,38 @@ else if(
 
             if (Record_ID > 0) {
             	params.put("RECORD_ID", new Integer(Record_ID));
-            	params.put("Record_ID", new Integer(Record_ID));
+            	if(!Optional.ofNullable(params.get("Record_ID")).isPresent()) {
```
mckayERP, line 523: Consider .OrElse() to remove the "if not" statement as in

    Optional.ofNullable(params.get("Record_ID")).orElse(
        params.put("Record_ID", new Integer(Record_ID)));

You could also extract these repeated patterns to a method as so

    protected void verifyParameterExistsOrAddDefault(
            HashMap<String, Object> params, final String parameterName,
            int defaultValue) {

        Optional.ofNullable(params.get(parameterName)).orElse(
            params.put(parameterName, new Integer(defaultValue)));

    }

Then it would look like

            if (Record_ID > 0) {
            	params.put("RECORD_ID", new Integer(Record_ID));
            	verifyParameterExistsOrAddDefault(params, "Record_ID", Record_ID);
            }

mckayERP (approved):

marcalwestf (approved):
Test OK.
Acceptance depending whether @yamelsenih considers it worthwhile do changes according to suggestion of @mckayERP

PR #3287, opened 2021-01-02
Scheduler Add support to frequency type:

yamelsenih:
This pull request add support to the follows frequencies to Scheduler
- Second
- Does Not Repeat
- Weekly
- Monthly
- Yearly
Also allows define custom Date Next Run for a Schedule

A example for only a run of schedule
Screenshot_20210102_001441

### DIFF
--- a/serverRoot/src/main/server/org/compiere/server/AdempiereServer.java
+++ b/serverRoot/src/main/server/org/compiere/server/AdempiereServer.java
@@ -206,7 +206,7 @@ public abstract class AdempiereServer extends Thread
 	 * Validate if is fixed time
 	 * @return
 	 */
-	private boolean isValidForRun() {
+	protected boolean isValidForRun() {
 		if(Optional.ofNullable(p_model.getFrequencyType()).orElse(MScheduler.FREQUENCYTYPE_Day).equals(MScheduler.FREQUENCYTYPE_DoesNotRepeat)
 				&& Optional.ofNullable(p_model.getDateLastRun()).isPresent()) {
 			return false;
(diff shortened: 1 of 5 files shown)


serverRoot/src/main/server/org/compiere/server/AdempiereServer.java
```diff
@@ -198,7 +201,19 @@ public void runNow()
 		//
 		log.fine(getName() + ": " + getStatistics());
 	}	//	runNow
-
+
+	/**
+	 * Validate if is fixed time
+	 * @return
+	 */
+	private boolean isValidForRun() {
```
mckayERP, line 209: For testing - if the junit upgrades gets approved - consider making this a protected method.  It is difficult to test private methods.

serverRoot/src/main/server/org/compiere/server/AdempiereServerMgr.java
```diff
@@ -557,7 +564,11 @@ public String getServerCount()
 	 */
 	public Timestamp getStartTime()
 	{
-		return m_start;
+		return startTime;
 	}	//	getStartTime
 
+	public static void main(String[] args) {
```
mckayERP, line 570: Why a main method?

mckayERP (commented):
OK - but why add the main() method?

serverRoot/src/main/server/org/compiere/server/AdempiereServer.java
yamelsenih: You are right, I will go

serverRoot/src/main/server/org/compiere/server/AdempiereServerMgr.java
yamelsenih: Just go to run with current IDE. Maybe can be removed

PR #3288, opened 2021-01-02
Just add transaction name for query based on POInfoColumn

yamelsenih:
Currently the transaction name is not setted when a POInfoColumn is called from PO class then when a field is referenced to Table the query for search Reference Value information is running without main transaction, this problem generate many active transaction like the follows:

```SQL

 3037 | Select AD_Reference_ID FROM AD_Column WHERE AD_Column_ID=$1
```

### DIFF
--- a/base/src/org/compiere/model/POInfoColumn.java
+++ b/base/src/org/compiere/model/POInfoColumn.java
@@ -50,7 +50,7 @@ public class POInfoColumn implements Serializable
 	 *	@param columnDescription Column Description
 	 *	@param isKey true if key
 	 *	@param isParent true if parent
-	 *	@param ad_Reference_Value_ID reference value
+	 *	@param referencevalueId reference value
 	 *	@param validationCode sql validation code
 	 *	@param fieldLength Field Length
 	 * 	@param valueMin minimal value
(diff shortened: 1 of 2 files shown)


base/src/org/compiere/model/POInfoColumn.java
```diff
@@ -64,9 +64,9 @@ public POInfoColumn (int ad_Column_ID, String columnName, String columnSQL, int
 		boolean isMandatory, boolean isUpdateable, String defaultLogic,
 		String columnLabel, String columnDescription,
 		boolean isKey, boolean isParent,
-		int ad_Reference_Value_ID, String validationCode,
+		int referencevalueId, String validationCode,
 		int fieldLength, String valueMin, String valueMax,
-		boolean isTranslated, boolean isEncrypted, boolean isAllowLogging, boolean isAllowCopy)
+		boolean isTranslated, boolean isEncrypted, boolean isAllowLogging, boolean isAllowCopy, String transactionName)
```
mckayERP, line 69: This is a public constructor - should it be overloaded?  It is only called from the loadInfo method above to the risk is small.

mckayERP (approved):
OK if the change to the public constructor is low risk - I think it is as it is only called in one place and is very low level.

marcalwestf (approved):
- tested classs POInfo and POInfoColumn
- trxName is passed correctly.
Test passed.

e-Evolution:
@marcalwestf , HI Mario @yamelsenih Yamel should first implement the advise of @mckayERP before the merge, move a public method not is good if he need adding suport the transaction then should create a new method with new signer and overload,is a good practice in development.

Would implement this and merge again

PR #3289, opened 2021-01-02
Just Implement Cache for get Table ID based on table name

yamelsenih:
This pull equest implement cache for **MTable.getTable_ID(String tableName)** method

### DIFF
--- a/base/src/org/compiere/model/MTable.java
+++ b/base/src/org/compiere/model/MTable.java
@@ -26,6 +26,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.logging.Level;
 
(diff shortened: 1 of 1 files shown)


base/src/org/compiere/model/MTable.java
```diff
@@ -876,22 +878,13 @@ public String getSQLCreate()
...
-			rs = pstmt.executeQuery();
-			if (rs.next())
-				retValue = rs.getInt(1);
-		} catch (Exception e) {
-			s_log.log(Level.SEVERE, SQL, e);
-			retValue = -1;
-		} finally {
-			DB.close(rs, pstmt);
+		Integer tableId = s_tableNameCache.get(tableName);
+		if(Optional.ofNullable(tableId).orElse(new Integer(0)) > 0) {
```
mckayERP, line 882: If using the Optional pattern, you could use .OrElseGet and do it all in one line

    public static int getTable_ID(String tableName) {

        return Optional.ofNullable(s_tableNameCache.get(tableName))
                .orElseGet(() -> 
                    getTable_IDAndAddToCache(tableName)
                );

    }

    private static Integer getTable_IDAndAddToCache(String tableName) {

        int retValue = DB.getSQLValue(null,
                "SELECT AD_Table_ID FROM AD_Table WHERE tablename = ?",
                tableName);
        s_tableNameCache.put(tableName, retValue);
        return retValue;

    }

Would the transaction name ever be required here?

mckayERP (approved):
Added a suggestion but its not required.  Good improvement.

base/src/org/compiere/model/MTable.java
e-Evolution: @mckayERP is right is not same orElse that  orElseGet to runtime the first is evaluate and second only if get None

yamelsenih:
Hi @mckayERP already exists your recommendations

mckayERP:
I'm OK with it.  Thanks.

Issue #3292, opened 2021-01-04
Error on applying migration script serial 5910.


mckayERP:
When applying the migration script 5910, I receive an error that the i.DocBaseType columns does not exist.  I think this is caused by duplicate sequence numbers and duplicate SQL updates.  There are two SQL update statemens and each is duplicated.  All four have the sequence number 140.  One query adds the RV_Invoice view with the new column i.DocBaseType.  The other adds RV_InvoiceLine view which uses the first view in the from clause.  With the same sequence number, the second update may be run before the first causing the error.

The script is also associated with version 3.9.3.  I think this should be 3.9.4.

PR #3293, opened 2021-01-04
3292 - fixes to migration script 5910 - remove duplicate steps and set the correct sequence.

mckayERP:
The original script had four steps with the same sequence number.  Two were duplicates.  The two originals needed to be applied in order but with the same sequence number, the order of application was unspecified - resulting in errors.

In this PR, I've deleted the duplicate entries and set the correct order.

I've also adjusted the Release No to 3.9.4

The correct entries are:

(New Lines 323-324)
    <Step DBType="Postgres" Parse="N" SeqNo="140" StepType="SQL">
      <Comments>RV_C_INVOICE</Comments>
      ...

and
(New Lines 403-404)
    <Step DBType="Postgres" Parse="N" SeqNo="145" StepType="SQL">
      <Comments>RV_C_INVOICELINE</Comments>
      ....


mckayERP:
Note - while the steps were duplicated, they were all for the Postgres database type.  Perhaps these should be changed to All database types?  As it is, the views will only be added to Postgres.

mckayERP:
Fixes #3292

marcalwestf (approved):
Test OK.

PR #3297, opened 2021-01-04
Change the order of steps to avoid an error when the views are added before the fields they require.

### LINKED ISSUE #3296: Migration script serial 06550 is causing errors.
Relates to #3241 

The order of steps in the migration script is wrong.  There are four Steps 0 which add views which references fields that do not exist yet.

mckayERP:
Fixes #3296.

There were four steps with sequence number zero that added views which depended on the fields added in later steps.  I moved these four to the end and the migration worked fine.


marcalwestf (approved):
- XML loaded 
- XML correctly executed
Test passed.

PR #3299, opened 2021-01-04
3298 Add Release No to migration script serial 06490

### LINKED ISSUE #3298: Migration Script 06490 Add Parameter Sales Region in Open Item does not have a release No.
The release number is blank so the migration script is loaded multiple times when RUN_MigrateXML is run multiple times.  It should be loaded only once.

mckayERP:
Fixes #3298 

Simply adds a release number to migration script 06490 Add Parameter Sales Region in Open Item.


yamelsenih:
Thanks @mckayERP happy new year

PR #3304, opened 2021-01-05
Add support to functionality Allow Negative Posting from Accounting

carlosaparadam:
Hi to all, this pull request allow to control that reversed documents with document type checked with Reversal with Original Accounting.

Currently all reversed documents with the document type with this check has been posted with negative amounts, without consider the setting for allow posting negative in Accounting Schema.

see this example:
ControlNegativePosting


marcalwestf (approved):
Changes tested by @SusanneCalderon .

PR #3305, opened 2021-01-05
Cleaning Landed Cost Allocation when don't have landed cost definition

carlosaparadam:
Currently, when you have an invoice and generate landed cost but you want revert the landed costs generated, is not possible because if deleted the landed cost definition don't deleted the landed costs allocation records. It causes landed costs to be generated in a document even without you want generated landed costs for the document. 

Currently, the method validate that landed cost definition and after clean the landed costs allocation
Screenshot_20210105_190325

My purpose is that clean the landed costs definition and after validate landed costs definition
Screenshot_20210105_190049

Current functionality:
CurrentLandedCostsAllocation

Purposed functionality:
PurpusedLandedCostsAllocation

### DIFF
--- a/base/src/org/compiere/model/MInvoiceLine.java
+++ b/base/src/org/compiere/model/MInvoiceLine.java
@@ -991,9 +991,6 @@ public class MInvoiceLine extends X_C_InvoiceLine implements DocumentReversalLin
 			getParent().setPosted(false);
 			getParent().saveEx();
 		}
-		MLandedCost[] lcs = MLandedCost.getLandedCosts(this);
-		if (lcs.length == 0)
-			return "";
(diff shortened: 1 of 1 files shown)


marcalwestf:
@carlosaparadam a very good catch! We came across this weird behavior and wondered how to get rid of it. 
Now I see you solved it in a simple way.

marcalwestf (approved):
- Open a _Purchase Invoice_ with _Landed costs_ and _Allocations_
- Delete the _Landed Costs_ of an _Invoice Line_
- _Prepare_ the _Invoice_
- The _Landed Costs allocations_ don't appear anymore.

Test passed.

PR #3306, opened 2021-01-06
Fixed error with payment selection for Over Under Payment

yamelsenih:
## Step by Step
- Create a AP Invoice with USD 2000
- Create a Payment for created invoice with 1500, set flag over un der
amount to true
- Print payment created
- Create a new payment selection
- Search previous invoice created from SB "Create from Invoice"

## Expected behavior
The Invoice shoul be showed with:
  - Grand Total: USD 2000
  - Open Amount: USD 500
  - Payment Amount: 500

## Bug
The invoice is missing

## What is the problem?
The Create From Invoice (SB) not show a invoice after add to payment
selection and don't allows create a new payment selection for a invoice
with a previous under payment

This Pull Request resolve it

### DIFF
--- a/migration/393lts-394lts/07230_Fixed_error_with_Over_Under_Amount_for_Payment_Selection.xml
+++ b/migration/393lts-394lts/07230_Fixed_error_with_Over_Under_Amount_for_Payment_Selection.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<Migrations>
+  <Migration EntityType="D" Name="Fixed error with Over Under Amount for Payment Selection" ReleaseNo="3.9.4" SeqNo="7230">
+    <Step SeqNo="10" StepType="AD">
+      <PO AD_Table_ID="53224" Action="U" Record_ID="50054" Table="AD_Browse">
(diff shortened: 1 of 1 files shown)


marcalwestf:
@yamelsenih  this error was already notified to us.

marcalwestf (approved):
- XML loaded and executed -> OK
- _Purchase Invoice_ created and completed -> OK
- Complete _Payment_ with under payment amount -> OK
- Create and save Payment Selection -> OK
- Open SB _Create from Invoice_ -> OK
- _Invoice_ appears in SB with open amount -> OK
- Select _Invoice_ and type in the desired amount -> OK
- Check _Payment Selection Line_ with amount created -> OK

Test passed

e-Evolution:
@yamelsenih , @marcalwestf 

657 / 5000
Resultados de traducción
I think this does not solve the problem completely, since the SQL statement is validating if there is no previous payment selection, which is possible in the following case.

An invoice that was paid with a partiality which will have a payment selection registered, and could be in the case that the next period requires paying another part of the invoice, this invoice would be excluded since it has a previous payment selection, because limits functionality to full payments only.

It seems to me that partial payments to an invoice should be supported in different periods.

@marcalwestf  Mario, did you test the case for partial payments to the same invoice?

PR #3308, opened 2021-01-08
Improve source code dependency

yamelsenih:
Just improve location for source code, currently the base directory
depend of request and project source folder

This pull request just move the **model classes** from
**org.adempiere.request** and **org.adempiere.project** to base source
folder

The callouts and model validator was kept back


marcalwestf (approved):
- checked files correctly moved
- compiler and clean project OK
- ANT build OK
- run under Swing OK
- run under ZK OK

Test passed.

e-Evolution:
Let me disagree with this change.

In the past I took a few hours to start the process of separating the code into directories that are focused on business functionality, in this case request and projects functionality initially, but the idea is to do this for the rest of the functionality, as has been done or increased the functionality of other functionalities (Manufacturing, Cashflow , Fleet , etc).

The objective of separating all the related classes by bounded context or business context https://martinfowler.com/bliki/BoundedContext.html , is to identify the dependencies to start after that a process of decoupling the code that today is redundant, a deficiency of the legacy code that we have in ADempiere, which for any application designed with modern philosophies like DDD are not correct, due to the high code coupling.

@yamelsenih , @marcalwestf  Therefore these changes first of all take the hours invested on my part to identify and separate the request and project management classes, we are practically going back to what we had before, which is to have all the code agglomerated in a single directory now called base instead of identifying classes and code by business process.

In this way, not only does it go against what was being done, but we also return to a concept of monolithic construction and management of the code, an approach that my company and I , we do not want to travel this path.

So I ask that this change be reversed, since it does not really represent an improvement, from our perspective this change makes us go back and leads us to the monolithic and coupled construction of the code that is a state from which we should leave.

I discussed this with @yamelsenih Yamel in private chat and I know the background of why he requires this, since he is trying to remove the dependency on Ant as a construction tool to pass to Gradle which is good, but on the other hand if Gradle does not allow to generate a binary or jar based on multiple source code directories, I don't think it's the right tool.

e-Evolution:
Talking with @yamelsenih and @marcalwestf  Yamel , Mario we decided by mutual agreement to revert to find a solution and refactor the code that is now coupled to a programming based on interfaces

PR #3310, opened 2021-01-12
Add tab for Social Media to Business Partner

yamelsenih:
Screenshot_20210112_162751


marcalwestf (approved):
XML loaded and executed.
Window BP opened and tab displayed.

Tst passed.

Issue #3321, opened 2021-01-19
Multiple processes using the process builder can cause conflicts.


mckayERP:
If creating multiple instances of the ProcessBuilder, and then running the processes from the instances after all the instances are created causes an error.

See the conversations in gitter on 16 and 17 Jan 2021.

mckayERP:
Merged.

PR #3322, opened 2021-01-19
Change the internal reference in the ProcessBuilder from a static field to a non-static field.

mckayERP:
See the comments in Gitter on 16 and 17 Jan 2021.

The ProcessBuilder uses a static reference to the builder instance.  When multiple process are using the ProcessBuilder class, this static reference can end up pointing at the last instance created.

Fixes #3221


mckayERP:
Fixes #3321.  Used wrong reference above.

PR #3336, opened 2021-01-24
Fix/#3329 client acct processor

### LINKED ISSUE #3329: Refactor the Accounting Server and Client Acct Processor
There is a lot of duplicated code in the Accounting Server and the Client Acct Processor.  The postSession method is almost a complete duplication.  These classes could be refactored to use a common class that performs the postSession.  Some of the code should be moved to the Doc class where the knowledge of posting, locking and unlocking documents should reside.  The posting methods in the accounting server and processor don't need to know this.

mckayERP:
Fixes #3329

This PR provides refactoring of the Client Acct Processor and the Acct Processor which use a lot of duplicated code.  The code was split into methods that were allocated to the classes responsible for the information managed by the methods.  Tests were added for all changes and the tests passed on my system.  Now all postings are managed by the SessionPoster class.  The Acct Processors only have simple calls to the SessionPoster to perform the posting actions.


marcalwestf:
hi @mckayERP this is a very intrusive PR, which touches a very sensitive part of the system. 
Before accepted, it will need to undergo comprehensive positive and (mostly) negative tests. 
These tests **must be**  accomplished by as many contributors as possible. This cannot be accepted by one contributor only.

Do you have a step-by-step test strategy to start with?
I'd suggest you describe how you tested bottom-up so we can replicate the tests and then we would add our own cases.
Thanks.

mckayERP:
Hi Mario @marcalwestf 

One of the benefits of having unit and integration tests is that the testing is incorporated with the pull request.  The commits include both unit and integration tests.  The integration test for the Client Acct Processor was written before the refactoring and I was checking every few minutes that the code changes didn't fail the tests.  The only addition I made to that test was to add an invoice that would fail posting and to verify that it was handled correctly.  The other tests were developed in a test driven design approach as I refactored the code - write a failing test first, then write the code to pass the tests, then refactor to clean up, then write or change the test so it fails again and keep going until done.  The tests provide nearly 100% coverage of the new and refactored code.

I appreciate that the code is sensitive.  Nothing relating to how the documents are posted was changed - so everything from Doc.post() on down remains the same.  

As a strategy to review, I'd start with the tests to see that they cover all routes through the code.  The unit tests look at individual methods.  Integration tests require the database and look at the bigger process.  The tests use the Garden World data and reset the accounting and then run the Client Acct Processor and ensure that all documents were posted.  I didn't do the same for the AcctProcessor as it was reduced to a trivial process.  The meat of the code is in the SessionPoster and that is well tested.

As a final test, I built and ran the server and reset the Garden World accounting and then ran the AcctProcess from the Server Manager to verify that all the documents were posted.  Then I created an invoice and tried again to ensure that it was posted on the next run.

So start by loading the branch in your IDE and running the tests.  In Eclipse, I can do this by clicking within a test file and hitting Alt-Shift X and then T to run the test or Alt-Shift D and then T to debug the test.   I use the EclEmma Eclipse plugin to check coverage and ensure that nothing important was missed.

You can also run the tests during the build process.  See the file test.properties in the utils_dev directory and turn on the integration tests during the ant builds.  

If you have a particular test case, let me know and I can add it to the tests.  With the test written, it takes just a few seconds to run through the all the related tests.  This gives a lot of confidence that any changes made in the future haven't broken anything.

Hope you and Susanne are doing well.  Stay safe!

Mike

marcalwestf:
thanks, @mckayERP for your thorough and swift explanation.
We'll follow your instructions and keep you informed.

I find it very instructive and useful as a showcase that you first developed the tests before developing the actual functionality.

e-Evolution:
@mckayERP great and good changes to improbamente the code, but we need are sure that refactories all work rightly  .

This code writing also proposes to write the following functional tests:

- Immediate accounting with automatic period from the server
- Non-immediate accounting with automatic period from the server
- Immediate accounting with automatic period from the client
- Non-immediate accounting with automatic period from the Client
- Send for posting with posting error
- Submit to post with automatic period
- Render to post with zero cost entries
- Accounting reset accounting with automatic period
- It is important to validate that when the accounting is generated automatically, the accounting posting is done with the same database transaction with which the document is completed.
- Validate that not execute Client Accounting Processor when exist running the same process same time, this may be cause account facts duplicate functionality



- Immediate accounting with no automatic period from the server
- Non-immediate accounting without automatic period from the server
- Immediate accounting without automatic period from the client
- Non-immediate accounting without automatic period from the Client
- Send for posting with posting error without automatic period
- Resend to post without automatic period
- Render to account with zero cost movements without automatic period
- Accounting reset accounting without automatic period
- Validate the correct operation of applying by document and that the re-accounting works correctly.
- Validate that can not execute the posting or same document and that generate duplicate account
- Validate that no duplicate accounting when Accounting Processor is run an web server and startup other web server same process same time, this may be cause account fact duplicate functionality

- Validate the order by in that documento are posted (DateAcct , ProcessedOn)
- Validate not rest accounting to financial period closed  
- Validate that manual reposting when financial period closed

mckayERP:
@e-Evolution That is a good list.  I'll try to structure it into some tests and add the results to the PR.

mckayERP:
More work is still required but here is an interim look at the results.

I've taken Victor's list and used it to structure a test in IT_DocumentPosting.java.  This doesn't cover all the tests but it does look at the cases where:

A document is completed when 
- Client Accounting is Immediate, Queue or Disabled
- Automatic Period Control is enabled or disabled
- Posting in an open period or a closed period

For a document that is already posted
- Client Accounting enabled (I or Q) or disabled (D)

The tests where setup to follow the given, when, then pattern and nested to reduce the amount of setup required.  The output of the test shows the following:

image

 There is a small issue with  where I put the file which I will fix shortly.

mckayERP:
To get full coverage of the tests, I added tests of the document engine.  I had to make  a few changes to the document engine code to make it testable.  The tests provide near 100% coverage.  (The DocumentEngine could stand a full refactoring.)

Next tests will look at the multi-threading where there are multiple accounting processors running - all trying to post the same document.

More work required to provide full coverage of Victor's test sets.

mckayERP:
I've been looking at the multi-threading issue.  A possible error case occurs when a client or account processor on one machine tries to post a document (not forced and not reposted) while a user on another machine attempts to post the same document with force=true.  If the first machine captures the lock (processing='Y'), the second machine will still capture the lock and attempt to post the document.  There is a chance that duplicate accounting will be generated if both computers check and verify no accounting entries exist before either posts the new entries.  This is a race condition.  Any ideas how should I test this? Trying to recreate this specific timing issue is nearly impossible.

The only suggestion I would have is to only use force=true in in Doc.post() is particular situations - for example when the Doc is stuck in the locked state.  Right now, force=true is the default when the DocumentEngine.post() method is used.  Changing this to force=false would not prevent the user from posting a document but it would prevent the race condition.

e-Evolution:
@mckayERP  Hi Mike!

I did the most common tests for accounting

- Immediate Posting, ok
- Posting in Queue, ok
- Repost by document, ok
- Reconcile forcefully, ok
- Post Client Accounting Processor when executing single way, ok
- Post Client Accounting Processor in parallel, get error posting

To replicate the behavior.

1.- Change CLIENT_ACCOUNTING to I

image

2.- Execute the process Client Accounting Processor in two browsers at the same time. I got errors

image

image

Although the accounting is not duplicated, the documents already posted are marked with a posting error.

I don't know if there can be a way to avoid executing a second process while the first does not finish to avoid some strange behavior.

On the other hand, the multi-process behavior happens, when you start two application servers to the same database and execute the accounting process run now, this generates duplicate accounting but it was an error in the current version.

Maybe if we think of a solution we can solve both cases.

kind regards
Victor

mckayERP:
Thanks again Victor,

This is a tough one for sure.   I'll have a look and see if I can come up
with some ideas.

Michael McKay
President, mckayERP <http://mckayerp.com>
+1 (613) 297-4483

<http://mckayerp.com>

e-Evolution:
@mckayERP can please resolve the conflict , this way I can make merge

mckayERP:
@e-Evolution Merged with develop and made the changes necessary to resolve conflicts.  Integration tested with the current seed.

Note that the Integration tests IT_ClientAcctProcessor will fail with the current seed util PR #3377 is approved and merged.

Issue #2921, opened 2019-12-30
The Generate Outbound (Sales Order/Distribution) the Weight and Volumen show incorrect info 


e-Evolution:
## Bug report
The Generate Outbound (Sales Order/Distribution) the Weight and Volumen show incorrect info.

#### Steps to reproduce
1.- Open The Generate Outbound (Sales Order/Distribution) and check the Weight and Volumen

#### Screenshot or Gif

![image](https://user-images.githubusercontent.com/786968/71603623-5c438500-2b23-11ea-8dc8-9af7b6468602.png)

#### Expected behavior Weight and Volumen based on product Weight and Volumen sum 
the column should show the 

#### Other relevant information
- Your OS: MacOS
- Java version: 1.8
- ADempiere version: 3.9.3
- Browser safari
- DB Provider: PG
- DB version: 12

EdwinBetanc0urt:
Hi @e-Evolution Victor, from what I understand you are fixing something that is currently wrong, in that case shouldn't it point to the `hotfix/3.9.4.001` branch? Since only new features should point to the `develop` branch.

Second the build gave error, that was failing in branches extending from `hotfix/3.9.4.001`, after the pull request https://github.com/adempiere/adempiere/pull/4081 and before the pull request https://github.com/adempiere/adempiere/pull/4222 (where it was corrected), I suggest you update to correct the build.

PR #3338, opened 2021-01-25
#2921 The Generate Outbound (Sales Order/Distribution) the Weight and…

e-Evolution:
… Volumen show incorrect info https://github.com/adempiere/adempiere/issues/2921

- Add Delivery Rule and Delivery Via
- Add the Attribute Set Instance Id
- Add Calculate Delivery to to multiples Attribute Set Instance Id

### DIFF
--- a/org.eevolution.warehouse/src/main/java/base/org/spin/form/OutBoundOrder.java
+++ b/org.eevolution.warehouse/src/main/java/base/org/spin/form/OutBoundOrder.java
@@ -23,20 +23,36 @@ import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.text.DecimalFormat;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 import java.util.Vector;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 
 import org.adempiere.exceptions.AdempiereException;
 import org.adempiere.exceptions.DocTypeNotFoundException;
 import org.compiere.minigrid.IMiniTable;
+import org.compiere.model.MBPartner;
 import org.compiere.model.MDocType;
+import org.compiere.model.MLocator;
+import org.compiere.model.MOrder;
 import org.compiere.model.MOrderLine;
+import org.compiere.model.MOrg;
 import org.compiere.model.MProduct;
+import org.compiere.model.MQuery;
 import org.compiere.model.MRefList;
 import org.compiere.model.MRole;
+import org.compiere.model.MStorage;
+import org.compiere.model.MTable;
 import org.compiere.model.MUOM;
+import org.compiere.model.MUser;
+import org.compiere.model.MWarehouse;
+import org.compiere.model.PrintInfo;
 import org.compiere.model.X_C_Order;
+import org.compiere.print.MPrintFormat;
+import org.compiere.print.ReportCtl;
+import org.compiere.print.ReportEngine;
 import org.compiere.util.CLogger;
 import org.compiere.util.DB;
 import org.compiere.util.DisplayType;
@@ -45,7 +61,9 @@ import org.compiere.util.KeyNamePair;
 import org.compiere.util.Msg;
 import org.compiere.util.Util;
 import org.compiere.util.ValueNamePair;
+import org.eevolution.exceptions.NoBPartnerLinkedforOrgException;
 import org.eevolution.model.I_DD_Order;
+import org.eevolution.model.MDDOrder;
 import org.eevolution.model.MDDOrderLine;
 import org.eevolution.model.MWMInOutBound;
 import org.eevolution.model.MWMInOutBoundLine;
@@ -63,24 +81,26 @@ public class OutBoundOrder {
 	public static CLogger log = CLogger.getCLogger(OutBoundOrder.class);
 	
 	public final int SELECT 					= 0;
-	public final int ORDER 						= 2;
+	public final int ORDER 						= 1;
 	/**	Lines									*/
 	public final int OL_WAREHOUSE 				= 1;
 	public final int ORDER_LINE 				= 2;
 	public final int OL_PRODUCT 				= 3;
 	public final int OL_UOM 					= 4;
-	public final int OL_QTY_ON_HAND 			= 5;
-	public final int OL_QTY 					= 6;
-	public final int OL_WEIGHT 					= 7;
-	public final int OL_VOLUME 					= 8;
-	public final int OL_SEQNO 					= 9;
-	public final int OL_QTY_ORDERED 			= 10;
-	public final int OL_UOM_CONVERSION 			= 11;
-	public final int OL_QTY_RESERVERD 			= 12;
-	public final int OL_QTY_INVOICED 			= 13;
-	public final int OL_QTY_DELIVERED 			= 14;
-	public final int OL_QTY_IN_TRANSIT 			= 15;
-	public final int OL_DELIVERY_RULE 			= 16;
+	public final int OL_ASI 					= 5;
+	public final int OL_QTY_ORDERED 			= 6;
+	public final int OL_UOM_CONVERSION 			= 7;
+	public final int OL_QTY_ON_HAND 			= 8;
+	public final int OL_QTY 					= 9;
+	public final int OL_WEIGHT 					= 10;
+	public final int OL_VOLUME 					= 11;
+	public final int OL_SEQNO 					= 12;
+	public final int OL_QTY_RESERVERD 			= 13;
+	public final int OL_QTY_INVOICED 			= 14;
+	public final int OL_QTY_DELIVERED 			= 15;
+	public final int OL_QTY_IN_TRANSIT 			= 16;
+	public final int OL_DELIVERY_RULE 			= 17;
+	public final int OL_DELIVERY_VIA_RULE 		= 18;
 	/**	Warehouse and Product					*/
 	public final int SW_PRODUCT 				= 0;
 	public final int SW_UOM 					= 1;
@@ -106,10 +126,12 @@ public class OutBoundOrder {
 	protected int 				salesRepId = 0;
 	/**	Warehouse			*/
 	protected int 				warehouseId = 0;
+	/**	Locator			*/
+	protected int 				locatorId = 0;
 	/**	Operation Type		*/
 	protected String 			movementType = null;
 	/**	Document Action		*/
-	protected String 			documentAction = null;
+	//protected String 			documentAction = null;
 	/**	Document Type 		*/
 	protected int 				docTypeId = 0;
 	/**	Document Type Target*/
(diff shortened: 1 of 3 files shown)


marcalwestf:
Hi @e-Evolution Victor, I tried to test this PR first on Swing, but there is no V class, so I wonder if you forgot to include it:
Selection_983

When trying to open the Smart Browser _Generate Outbound (Sales Order/Distribution Order)_, I get a _Class Not Found_ Exception:
Selection_984

I will continue to test on ZK in the meantime.

marcalwestf:
One remark: when an Order is selected, lines are displayed with a weight and volume calculation which depends on the quantity. **This is OK**:
Selection_986

when the lines are selected, weight and volume are rounded (because the unit is _Each_), **which is correct**.
Selection_987

The only improvement here would be to round also correctly **before** the lines are selected.

yamelsenih:
Hi @marcalwestf @e-Evolution already exist other PR for this and my test was failed, I want to test it before merge in ADempiere

marcalwestf:
... I created a product with an UOM=kg, fractional weight and volume; then completed a Sales Order with a fractional quantity.

The Weight and Volume of the selected line are nevertheless rounded, which I think is an error because volume, as well as weight, should in these cases be fractions.
Selection_988

marcalwestf:
@yamelsenih ok. 
Which PR is about? Thanks.

marcalwestf:
concerning my initial comment that the V class is missing: it was so from the beginning.

marcalwestf:
..waiting for @yamelsenih Yamel to carry out tests before accepting this PR.

yamelsenih:
Testing ti now

yamelsenih:
Testing it

org.eevolution.warehouse/src/main/java/base/org/spin/form/OutBoundOrder.java
```diff
@@ -166,19 +188,21 @@
...
 		Vector<Vector<Object>> data = new Vector<Vector<Object>>();
 		StringBuffer sql = null;
 		if (movementType.equals(I_DD_Order.Table_Name)) {
 			//Query for Material Movement
 			sql = new StringBuffer("SELECT " +
-					"wr.Name Warehouse, ord.DD_Order_ID, ord.DocumentNo, " +	//	1..3
-					"ord.DateOrdered, ord.DatePromised, reg.Name, cit.Name, sr.Name SalesRep, " +	//	4..8
-					"cp.Name Partner, bploc.Name, " +	//	9..10
-					"loc.Address1, loc.Address2, loc.Address3, loc.Address4, ord.C_BPartner_Location_ID, ord.Weight, ord.Volume " +	//	11..17
+					"ord.DD_Order_ID, ord.DocumentNo, ord.Description , cp.Name Partner, bploc.Name, wr.Name Warehouse, " +
```
yamelsenih, line 200: I don't sure if this is correct, maybe is a change for a customer or own change, the Weigh and Volume is showed from Order to select instead of Quantoty to deliver.

I think that this change can be part of a ADempiere Labs, Customizxation for your customer otr a Location

org.eevolution.warehouse/src/main/java/base/org/spin/form/OutBoundOrder.java
```diff
@@ -236,10 +260,12 @@
 			// role security
 		} else {//Query for Sales Order
 			sql = new StringBuffer("SELECT " +
-					"wr.Name Warehouse, ord.C_Order_ID, ord.DocumentNo, " +	//	1..3
-					"ord.DateOrdered, ord.DatePromised, reg.Name, cit.Name, sr.Name SalesRep, " +	//	4..8
-					"cp.Name Partner, bploc.Name, " +	//	9..10
-					"loc.Address1, loc.Address2, loc.Address3, loc.Address4, ord.C_BPartner_Location_ID, ord.Weight, ord.Volume " +	//	11..17
+					"ord.C_Order_ID, ord.DocumentNo, ord.Description , cp.Name Partner, bploc.Name, wr.Name Warehouse , " +
```
yamelsenih, line 263: I don't sure if this is correct, maybe is a change for a customer or own change, the Weigh and Volume is showed from Order to select instead of Quantoty to deliver.

I think that this change can be part of a ADempiere Labs, Customizxation for your customer otr a Location

org.eevolution.warehouse/src/main/java/base/org/spin/form/OutBoundOrder.java
```diff
@@ -1114,12 +1343,12 @@ public String validStock(IMiniTable stockTable) {
 			//	Get Values
 			String product 				= ((KeyNamePair) stockTable.getValueAt(i, SW_PRODUCT)).getName();
 			String warehouse 			= ((KeyNamePair) stockTable.getValueAt(i, SW_WAREHOUSE)).getName();
-			BigDecimal m_QtyOnHand 		= ((BigDecimal) stockTable.getValueAt(i, SW_QTY_ON_HAND));
-			BigDecimal m_QtyInTransit 	= ((BigDecimal) stockTable.getValueAt(i, SW_QTY_IN_TRANSIT));
-			BigDecimal m_QtySet 		= ((BigDecimal) stockTable.getValueAt(i, SW_QTY_SET));
-			BigDecimal m_QtyAvailable 	= ((BigDecimal) stockTable.getValueAt(i, SW_QTY_AVAILABLE));
+			BigDecimal qtyOnHand 		= ((BigDecimal) stockTable.getValueAt(i, SW_QTY_ON_HAND));
```
yamelsenih, line 1346: Please make this code more atomic, the refactory can be change in other pull request as Source code Improve

org.eevolution.warehouse/src/main/java/base/org/spin/form/OutBoundOrder.java
```diff
@@ -1131,9 +1360,9 @@ public String validStock(IMiniTable stockTable) {
 				.append(product)
 				.append("[")
 				.append("@M_Warehouse_ID@=").append(warehouse)
-				.append(" @QtyAvailable@=").append(m_QtyOnHand.subtract(m_QtyInTransit).doubleValue())
-				.append(" @QtyToDeliver@=").append(m_QtySet.doubleValue())
-				.append(" @PickedQty@=").append(m_QtyAvailable.doubleValue())
+				.append(" @QtyAvailable@=").append(qtyOnHand.subtract(qtyInTransit).doubleValue())
```
yamelsenih, line 1363: Please make this code more atomic, the refactory can be change in other pull request as Source code Improve

org.eevolution.warehouse/src/main/java/base/org/spin/form/OutBoundOrder.java
```diff
@@ -1152,14 +1381,14 @@ public void loadBuffer(IMiniTable orderLineTable) {
 		int rows = orderLineTable.getRowCount();
 		int m_C_OrderLine_ID = 0;
 		BigDecimal qty = Env.ZERO;
-		Integer seqNo = 0;
+		BigDecimal seqNo = BigDecimal.ZERO;
```
yamelsenih, line 1384: Please make this code more atomic, the refactory can be change in other pull request as Source code Improve

yamelsenih (changes requested):

yamelsenih:
I am waiting for resolve all suggestions

e-Evolution:
In any case, I am trying to improve this functionality, although its application is really limited and it cannot be used in an environment where you have multiple people preparing shipments since when calculating picking on memory two people may be cause conflict for the same existence, for what this form has a use limited to one person.

That is why the process-based approach is more reliable since it is done through an atomic transaction so you do not have the problem of conflict in the use of existence.

Maybe @yamelsenih  Yamel is right and we should remove it from the core system and only keep it as customization due to its limited scope or the possibility of causing problems in the picking

yamelsenih:
Hi @e-Evolution I use this funciionality for many customers without problems. Your changes only change the current behavior and I think that could can implent your own changes for your customization. The Generate outbound order was created for generate a outbound based on current sales orders or distribution and inventory. The attributes management is nice and I think that can be included but my test with it have a result very distint.

In other hand. The weight and volume is based on order currently and I think that calculate it I not necessary because already exist a calculation in status bar when is selected a record.

Best regards

e-Evolution:
Well, @yamelsenih  say what you want, even using the functionality that you introduced, does not work in 3.9.3 and I put the steps to replicate the problem.

is clear that exists  a concurrency problem and it was demonstrated step by step with a business case, you can close your eyes or tell me I don't understand

always is easier answering anything than using answering with arguments.

Anyway, @marcalwestf Mario will review the case, I prefer that he review and accept or not,  as he took the time to deploy and review with detail,  and it may even be better to disable this functionality, firstly because there is no implementation for swing and because not work.

yamelsenih:
Hi @e-Evolution I don't agree with merge it. If exist a bug with concurrent then the results of bot (before and after) must be the same and the change proposed is not the same. Also add changes like weight and volume distint of original propose of form. I add all comment in the pull request and I am waiting for each correction or reply 

Best regards

yamelsenih:
Other comment: If you want to solve bug then limit your pull request to only solve it instead.

This add more functionality and buggies. Is not atomic.

> Add Delivery Rule and Delivery Via
>Add the Attribute Set Instance Id
>Add Calculate Delivery to to multiples Attribute Set Instance Id

The. What is the real scope of this pr? Make a fix of bug or also is enhance?

Please make other PR or change this for make it more atomic and test with a real case.

Best regards

marcalwestf:
@e-Evolution  @yamelsenih  Victor and Yamel, how do we here go on?

yamelsenih:
I tasted it and not get the same result for query. I can test another time now

yamelsenih:
Testing it:

First comment:
This form is was created for improve the distribution management and exist columns that was designed for show the information more fast to operator. I saw that all columnas was changed and it is not good.
The main interest for a Logistic Manager is know what is the delivery route.
Before change:
Before_1
After change:
After_1

Recomendation: I am agree with add new columns but follow the original desing you can add they after last instead

yamelsenih:
Second comment:
For Attribute set instance I think tah this change is really important and should be added but a only chanche without change all desing. I agree with add it but is innecessary change all form just for add support to attribute set instance.

yamelsenih:
Third comment (Tech comment):
I created the Generate Outbound Order wwith many care of get the maximum speed for operations, Note that not exist Sub-querys and other bad use of SQL. I can see that the weight and volume is added ad Sub-query and it is very bad for performance of query:

```SQL
(SELECT SUM((ol.QtyOrdered - ol.QtyDelivered) * p.Weight)  AS Weight FROM DD_OrderLine ol INNER JOIN M_Product p ON(p.M_Product_ID = ol.M_Product_ID) WHERE ol.DD_Order_ID = ord.DD_Order_ID) AS Weight
```

Exists many ways (not easy but better like) SELECT AS FROM and other way for use aggregate functions instead it

yamelsenih:
Fourth comment:
The additional columns like weigh and volume already exists as part of form. The just get the weith and volume of Sales / Distribution order (This calculation is created with the model of MOrder and MDDistribution) I think that this is very rounded because already exists a business logic for it and is unnecessary make other calculation here. Also the total weigh and total volume can be showed when you select the lines of Orders, see:
Screenshot_20210710_080707
Screenshot_20210710_080725

yamelsenih:
I suggest create a Pull Request for add support to Attribute set instance and other adding the additional columns

yamelsenih:
Hello @e-Evolution can you update this pull request? or create a new instead?

e-Evolution:
Close open other PR

Issue #3333, opened 2021-01-23
[Feature Request] Inbound/Outbound Definition Window priority is show incorrect way


e-Evolution:
## Feature request
 Inbound/Outbound Definition Window priority is show incorrect way

![image](https://user-images.githubusercontent.com/786968/105614336-5489fa80-5d8e-11eb-949c-149911437577.png)

PR #3339, opened 2021-01-25
#3333  Inbound/Outbound Definition Window priority is show incorrect …

e-Evolution:
…way https://github.com/adempiere/adempiere/issues/3333


marcalwestf (approved):
XML loaded and executed successfully.
Checked changes in Application Dictionary: AD_Client_ID is now read-only; Priority was moved down.
Changes in Window _Inbound/Outbound Definition_:
Selection_982

Test passed.

PR #3347, opened 2021-01-27
Fixed error with receipt without order line or RMA line

yamelsenih:
This pull request validate reference with Order Line or RMA Line from
Receipt line, see the gif
Validate Order Line and  RMA Line

### DIFF
--- a/base/src/org/compiere/model/MInOut.java
+++ b/base/src/org/compiere/model/MInOut.java
@@ -37,7 +37,6 @@ import org.compiere.print.ReportEngine;
 import org.compiere.process.DocAction;
 import org.compiere.process.DocumentReversalEnabled;
 import org.compiere.process.DocumentEngine;
-import org.compiere.util.CLogger;
 import org.compiere.util.DB;
 import org.compiere.util.Env;
 import org.compiere.util.Msg;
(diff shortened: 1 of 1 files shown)


marcalwestf:
hi @yamelsenih Yamel, 
I just tested your changes.

TEST
Tested the functionality on a production database and works. 
This means, your changes behave as described.
This is OK.
**But**:

USE CASE
We noticed some customers have real-life cases such as the following:
1. You complete a _Purchase Order_
2. The provider delivers the products you put in the _Purchase Order_
3. Additionally, the provider delivers one or more products **not included in the PO** because of different reasons (an agreement, transport conditions, bonus, etc.).
4. Thus, there are situations where the _Material Receipt_ **must have**  lines without reference to either PO or RMA lines. 
5. This PR without further changes would break backward compatibility.

SUGGESTION
As the changes do work well, and in order to preserve backward compatibility, I wonder if an entry in table _AD_Sysconfig_ would be recommendable? 
The default value would be not to test for references and it would be changed only if desired.
Then, the customer would decide if it is desired (or not) to have this functionality.

e-Evolution:
Hello @marcalwestf Mario, @yamelsenih Yamel , the only question I have is how the cost is managed, since the business logic for the standard cost and average cost is based on the cost of the purchase order.

I have some questions :

How does a return without reference to a purchase order affect the average purchase order cost or average invoice cost?

How does a return without reference to a purchase order affect the calculation of variation in purchase order price or variation in invoice price?

Can you include the tests of these esenarios?

yamelsenih:
Hi @marcalwestf @e-Evolution the problem is that when a receipt is completed without reference to PO Line, the cost is wrong, I think that make a receipt without PO Line is very bad because it can broken the cost engine and not exists a process for fix it.

A way for manage warehouse in without PO Line can be using **Internal Use** or **Physical Inventory**

yamelsenih:
Maybe this approach is more complicated. I will add other PR with a better definition or just add a model validator for y instance meanwhile is discuss

PR #3257, opened 2020-11-21
#2658 invoice externalization for e-invoicing UBL and CII

homebeaver:
this implements issue #2658 both schemas

- OASIS UBL
and
- UN/CEFACT

Hi Mario @marcalwestf ,
do you remember? A year ago you started to test a PR for this issue.
See you remarks and questions : https://github.com/adempiere/adempiere/pull/2704#issuecomment-530473764
and my answers : https://github.com/adempiere/adempiere/pull/2704#issuecomment-530694867

I updated the description how to test: see https://github.com/adempiere/adempiere/issues/2658#issuecomment-520595450

I stopped then the work on this topic. And now I resume.
Can you please review the new PR?

regards EUGen

Issue #2658, opened 2019-06-26
create e-invoices conforming to OASIS UBL or UN/CEFACT (EU 16931)


homebeaver:
## Feature request
I suggest to enhance the customer invoice with xml-information, which conforms to a standard.

#### which standard?
the most popular xml standards for invoices are
- OASIS UBL
- UN/CEFACT

![grafik](https://user-images.githubusercontent.com/3493126/60188000-e1757680-982e-11e9-84be-5e3cc28263bb.png)

The xml should be included in the readable pdf invoice - see notes about ZUGFeRD in #1567 
- ZUGFeRD is a german implementation for the EU Norm 16931
- in france it is called x-factur
- also OpenPEPOL is an implementation of EU 16931

![grafik](https://user-images.githubusercontent.com/3493126/60189518-4d58de80-9831-11e9-9e1a-6f53a7537c3f.png)

#### example

![grafik](https://user-images.githubusercontent.com/3493126/60190263-a6754200-9832-11e9-8c9c-33207cb5fef3.png)

regards

homebeaver:
here is the example pdf: 
[200002.pdf](https://github.com/adempiere/adempiere/files/3330685/200002.pdf)

homebeaver:
@e-Evolution , @yamelsenih , @marcalwestf , @SusanneCalderon 
what do you think about this feature request?

trifonnt:
I find this functionality very interesting.
In fact I have not seen any other Open Source ERP to have such functionality.

yamelsenih:
I will test it

yamelsenih:
@homebeaver have you a change or pull request for it?

homebeaver:
Hi @yamelsenih - I'am working on it. Let me first explain the implementation idea. See the scatched picture:

![20190711_225944](https://user-images.githubusercontent.com/3493126/61085302-2f45cd80-a430-11e9-8dde-1630445722be.jpg)

to get the resultig pdf - like the 200002.pdf example above we need
- modules to create a UBL invoice , marshalling, etc conforming to EN16931 rules (this is not ADempiere specific - so I'll contribute a jar : the yellow circle)
- then we need mapping modules AD-objects to UBL-pojo's. The generic mapping can be subclass for user needs. Expl: mapping client UoM's to UBL-compatible UoM's : orange in adempiere base
- at last we need a pdf-appender to merge the xml+pdf into a hybrid A3-PDF

[...]

Issue #1400, opened 2017-10-20
Dynamic report export


yamelsenih:
Currently just exist some file export from report (See pic)
![screenshot_20171019_195559](https://user-images.githubusercontent.com/2333092/31799467-3939eb6e-b508-11e7-9928-b4538736abf0.png)
My proposal is add a abstract class for handle standard export and it can be extended for add more format to default list.

It allows create more custom formats for export from report view.

yamelsenih:
For use and documentation you can see: [http://wiki.adempiere.net/Spin_Contribution:_FR:_Export_Format_From_Report](http://wiki.adempiere.net/Spin_Contribution:_FR:_Export_Format_From_Report)

Best Regard

e-Evolution:
This commit break the back compatibility not exit the method 

public boolean createXML (File file)

public boolean createXML (Writer writer)

a new functionality not should break the back compatibility 

Please fix or revert.

yamelsenih:
Hi Victor, I can add this method, but I was assured that all core functionality is the same. No problem I will add to it

e-Evolution:
yes , not should be remove , you can implement new implement but not should remove.

I saw other methods was removed 

public boolean createDelimitedFile (Writer writer, char delimiter, Language language, boolean printHeader)

private void createCSVvalue (StringBuffer sb, char delimiter, String content)

yamelsenih:
You can try use the new custom format, is more easy

yamelsenih:
**private void createCSVvalue (StringBuffer sb, char delimiter, String content)** a private method is not used for other, it should not add for back compatibility.

e-Evolution:
This change continue break the back compatibility.

the original method was with 2 parameters , some customization launch this error org.compiere.print.ReportEngine.createXLSX(Ljava/io/File;Lorg/compiere/util/Language;)V

Again a new functionality not should change public methods 


/**
	 * Create Excel file
	 * @param outFile output file
	 * @param language
	 * @throws Exception if error
	 */
	public void createXLS(File outFile, Language language)
	throws Exception
	{
		PrintDataExcelExporter exp = new PrintDataExcelExporter(getPrintData(), getPrintFormat());
		exp.export(outFile, language);
	}
	
	public void createXLSX(File outFile, Language language)
	throws Exception
	{
		PrintDataExcelExporter exp = new PrintDataExcelExporter(getPrintData(), getPrintFormat(), true);
		exp.export(outFile, language);
	}

[...]

PR #3349, opened 2021-01-30
#2658 invoice externalization in a specifc directory

homebeaver:
Hi @marcalwestf , @e-Evolution 

this PR replaces the #3257 : invoice externalization for e-invoicing UBL and CII

- I moved the package einvoice to an extra subdir as suggested by Mario : https://github.com/adempiere/adempiere/pull/3257#issuecomment-769158910 and requested by Victor : https://github.com/adempiere/adempiere/pull/3257#issuecomment-769196722
- the functionality is unchanged to the tested PR

regards EUGen


homebeaver:
Hi Mario @marcalwestf ,

in https://github.com/adempiere/adempiere/pull/3257#issuecomment-766391813 you wrote:

> Conclusion:
> The changes are not intrusive and are backward-compatible.
> The XMLs generated are correct.
> I haven't tested if the XMLs generated are **semantically correct or if they do conform to the specifications**.
> This PR can be accepted if the remarks on my previous comment are considered.

On `fnfe-mpe.org` there is a web-validator for CII-invoices.
I dropped two xmls generated with AD+this PR. Here is the result:

grafik

PS: the fnfe-mpe validator does not work for UBL. For UBL-xml there is a [cmdline validator](https://github.com/itplr-kosit/validator)

Hope, this helps to commit the PR.
regards

e-Evolution:
Hi  @homebeaver  

This code looks much better and more isolated, less invasive than the previous one, and is built as a separate module.

I still have a question why you have to modify the core classes, with functionality that not is for general solution.

MInvoice.java
ExportFormatXML.java
PO.java

Can you give me more detail of what you are trying to do, maybe I can help you with a better approach.

I implement the electronic invoicing of my country and use the standard export formats of ADempiere reports in XML to convert to a required format using style sheets.

In my case, I added the functionality through a model validator, so as not to put specialized code in the general functionality of the system.

homebeaver:
Hi Victor, @e-Evolution
Thank you for viewing the PR.

I‘ve sketched my idea in feature request #2658 in Jun 2019.
There are some XML Standards representing Business documents. I know 3 of them pretty good:

1. openTrans is document oriented but not wide spreaded
1. UBL by OASIS is wide spreaded and and consists of 62 main business documents
1. UN/CEFACT schema on the other side is process oriented. The Cross Industry Invoice (CII) spec was released Oct 2016
CII is more mature then UBL. In the expectation that UN/CEFACT will release its own XML schemas OASIS declared not to release further versions of UBL, see [link](http://ubl.xml.org/ebXML). The UN/CEFACT Supply Chain Reference Data Model released in 2019 contains Cross Industry Order.

- Both CII and UBL are accepted syntaxes to the European Norm 16931 for electronic invoices.
- The XML produced by Adempiere does not implemet any Standard. It just represents the AD data model.

I add `PO.externalize()` as an abstract method to be implemented in subclass i.e. invoice or order or any other business documents.

The default extrnalization does nothing and `createXML` ramains unchanged to be back comatible, as you claimed in https://github.com/adempiere/adempiere/issues/1400#issuecomment-340230574 . This is the modification in `ExportFormatXML`.

The modification in `MInvoice` ensures no change to createXML and special invoice externalisation only if the client defines the SysConfig key "XML_EINVOICE_SCHEMA_NAME" as specified in issue #2658 mentioned above.

This guarantees minimal side effects and clients/tenants can define own externalisation Schema, 
i.e. Client A can use UBL , Client B CII, Client C None. 
Even own implementations are possible. Client D can use openTrans, but has to implement the externalisation.

I hope the answer is datailed enough
Grüße / regards
EUGen 
FI to @marcalwestf , @trifonnt

marcalwestf:
@homebeaver hi Eugen, I just read the information of the links you provided.
Thanks a lot for the time and effort invested trying to explain the matter and convey your ideas.

@e-Evolution  Victor's  and @yamelsenih Yamel's (valid) objection is that if this PR is about a custom-made or even country-made solution, it should be implemented as a model validator.

Reading the links you provide, it seems that it is not about a custom or a country-only solution, but these are standards indeed, thus worthwhile implementing. 

Perhaps can @trifonnt  Trifon, @e-Evolution Victor and @yamelsenih Yamel cast their opinions.
Thanks!

trifonnt:
@homebeaver Eugen, @marcalwestf Mario, @e-Evolution Victor and @yamelsenih Yamel,

>Reading the links you provide, it seems that it is not about a custom or a country-only solution, but these are standards indeed, thus worthwhile implementing.

I strongly support this Pull Request to be integrated into the core as this is not country specific solution.

Kind regards,
Trifon

yamelsenih:
Hi @trifonnt thanks.

I think that this cuncionality can be integrated using this way for export report: https://wiki.adempiere.net/Spin_Contribution:_FR:_Export_Format_From_Report is very easy implement it and I use this functionality for Tax declaration and other especific format.

Best regards

homebeaver:
Hi, I resoved conflicts on .classpath coming from commit 0a280d2d5c29407dae181beb42dc9e1a5fcc5ae7

marcalwestf:
@homebeaver @trifonnt  hi Eugen, Trifon,
the functionality is OK.

What Victor and Yamel are suggesting is to implement the functionality in question without touching core classes, but with a _Model Validator_, which would do the same.

If somebody doesn't want to use this functionality, he simply has to de-activate the Model Validator and vice-versa. This would bring some advantages like if there is a need for enhancements, improvements, or bug-fixing, they can be done without touching the core classes.

I will move this PR to Release 4.0 so the modifications can be done as suggested and this PR accepted.

yamelsenih:
Hi @homebeaver @trifonnt you can see a implementation here: https://github.com/adempiere/LVE/blob/master/withholding/src/main/java/base/org/erpya/lve/util/exp/ExportFormatXML_ISLR.java

yamelsenih:
Hello @homebeaver  I will to close this pull request. Please if you have any change then a new pull request is welcome.

Issue #3368, opened 2021-02-23
Throws an error when doing partial receipts from an outsourced service PO


iscapnet:
## Bug report

Currently Adempiere throws an error when doing partial receipts from an outsourced service PO during crossover period and the previous period is closed

#### Steps to reproduce

1. Create a manufacturing workflow with an outsource service item, for example PChair, add outsourcing service "Chrome Service"
2. Create a manufacturing order using the BOM and the workflow for a small quantity of Chairs and complete it.
3. Create a document type Activity Control
4. Go to Activity Control Report window and add a new record with the quantity of chairs from the previous MO
5. Hit "Complete" and look at the bottom of the screen for the outsourced service PO
6. Go to the Purchase Order and complete it
7. Go to Material receipt and do a partial receipt, leave the rest pending for the next month
8. Close the current period and open the next period
9. Create a new material receipt with the future date for the remaining Chairs and complete it
10. Go to Activity Control Report window and complete it, then the error shows up

#### Screenshot or Gif
![image](https://user-images.githubusercontent.com/10192498/108788207-ad3ae780-7534-11eb-8975-05dfb0fb6da0.png)

#### Link to minimal reproduction

N/A

#### Expected behavior
An outsourced PO should allow partial material receipts even within multiple disperse periods as long as the originating MO is not closed, this is a very common logistics practice.

#### Other relevant information
- Your OS: Docker/Linux
- Java version:  OpenJDK 64-Bit Server
- ADempiere version: 3.9.3
- Browser Chrome
- Browser Version 88.0.4324.182 
- DB Provider: PostgreSQL
- DB version: 12.2

#### Additional context

N/A

PR #3369, opened 2021-02-23
#3368 Throws an error when doing partial receipts from an outsourced …

e-Evolution:
…service PO https://github.com/adempiere/adempiere/issues/3368


marcalwestf (approved):
Steps performed as described by @iscapnet  Enrique.
Error corrected.
Test passed.

iscapnet:
Thanks @marcalwestf !

PR #3389, opened 2021-03-09
Feature/#3388 average costing

### LINKED ISSUE #3388: Costing Method AverageInvoice, Problem with voided documents
## Bug report
When a document is voided, the cancellation document copies the Cost Detail register from the original document. This calculation is correct when the original and the cancellation document have the same accounting date. When the date of the annulation document differs from the original it is correct to calculate the new average cost from the last cost detail.

## Solution
Considerate the last Costdetail for the calculation.

SusanneCalderon:
Fixes #3388 
This enhancement is needed due to the implementation of reverse with current date


SusanneCalderon:
@yamelsenih , @e-Evolution 
I tested the original code and this pull request is obsolete. I will close it.

PR #3391, opened 2021-03-09
Feature/#3390 rounding problem calculate tax

SusanneCalderon:
This numeric problems occurs when the tax rate is a prime number


SusanneCalderon:
@e-Evolution Hi Victor, I merged and now the check has passed

e-Evolution:
@SusanneCalderon , I check this change in production and got an error when the tax is not calculated based on document level, on the other hand, I think that is wrong the {}  because this defines the scope for tax document level, and is incorrect.

given this strange behavior,  I  need to revert this pull request

e-Evolution:
@SusanneCalderon Hi Susanne

this my suggestion for this bugfix:

```		
		if (taxBaseAmt.scale() > getPrecision())
			taxBaseAmt.setScale(getPrecision(), BigDecimal.ROUND_HALF_UP);
		
		//	Calculate Tax
		if (documentLevel || taxAmt.signum() == 0)
			taxAmt = tax.calculateTax(taxBaseAmt, isTaxIncluded(), getPrecision());

		setTaxAmt(taxAmt);

		//	Set Base
```

PR #3394, opened 2021-03-10
Remove Hard Code for C_Location in Replication Data 

Raul-mz:
Having configured an export format with reference to the c_location table, it did not import the registry.
Example: Create business partner location export format add a line with C_Location_ID type Referenced. Run Test Import Processor.


marcalwestf:
hi @Raul-mz Raúl, can you add please more information as to test your pull request?
I suppose I will need a special file.
Could you provide the file according to the configuration of the screenshot that follows?
Or If the configuration of the screenshot is wrong, do please tell me what is wrong.

Thank you!
Selection_051

marcalwestf (approved):
Code OK.
Test passed.

Issue #3395, opened 2021-03-11
Error in update reserved quantity in case different warehouse and attributeset instance set in shipment


SusanneCalderon:
## Report
In the following case

- Shipment executed from a different warehouse as the sales order
- The attributeset instance is set in the shipement line
the reserved quantity in the warehouse of the sales ordered is add and not subtracted.

#### Steps to reproduce

- Create a standard sales order with warehouse 1
-Create the shipment in process and change from warehouse 1 to warehouse 2
- Select a attribute set instance in the shipment line

#### Screenshot or Gif
![Screenshot 2021-03-11_01](https://user-images.githubusercontent.com/15349639/110808601-c2f52000-8249-11eb-91e7-0e697ae6f7ca.png)
![Screenshot 2021-03-11_02](https://user-images.githubusercontent.com/15349639/110808599-c25c8980-8249-11eb-91ad-3a2a54421800.png)
![Screenshot 2021-03-11_03](https://user-images.githubusercontent.com/15349639/110808596-c25c8980-8249-11eb-9bdc-ef2796ac7219.png)
![Screenshot 2021-03-11_04](https://user-images.githubusercontent.com/15349639/110808595-c1c3f300-8249-11eb-84ca-907e3ae20eea.png)
![Screenshot 2021-03-11_05](https://user-images.githubusercontent.com/15349639/110808588-c1c3f300-8249-11eb-91ee-e2dfab2096c9.png)

#### Solution 
MInOut complete(). Not negate qtySO in line 1441
`
					if (!sameWarehouse) {
						//correct qtyOrdered in warehouse of order
						MWarehouse warehouse = MWarehouse.get(getCtx(), orderLine.getM_Warehouse_ID());
						if (!MStorage.add(getCtx(), orderLine.getM_Warehouse_ID(),
								warehouse.getDefaultLocator().getM_Locator_ID(),
								inOutLine.getM_Product_ID(),
								inOutLine.getM_AttributeSetInstance_ID(), reservationAttributeSetInstance_ID,
								Env.ZERO, QtySO, QtyPO.negate(), get_TrxName()))
							{
								processMsg = "Cannot correct Inventory";
								return DocAction.STATUS_Invalid;
							}
					}
`

e-Evolution:
[...]

PR #3396, opened 2021-03-11
Adempiere #3395 Error in update reserved quantity

SusanneCalderon:
https://github.com/adempiere/adempiere/issues/3395
This error occurs when the delivery is from a different locator and the attributeset instance in the shipment line is set

### DIFF
--- a/base/src/org/compiere/model/MInOut.java
+++ b/base/src/org/compiere/model/MInOut.java
@@ -1438,7 +1438,7 @@ public class MInOut extends X_M_InOut implements DocAction , DocumentReversalEna
 								warehouse.getDefaultLocator().getM_Locator_ID(),
 								inOutLine.getM_Product_ID(),
 								inOutLine.getM_AttributeSetInstance_ID(), reservationAttributeSetInstance_ID,
-								Env.ZERO, QtySO.negate(), QtyPO.negate(), get_TrxName()))
+								Env.ZERO, QtySO, QtyPO.negate(), get_TrxName()))
 							{
 								processMsg = "Cannot correct Inventory";
 								return DocAction.STATUS_Invalid;


e-Evolution:
Hi @SusanneCalderon Susanne first of all, ADempiere is not designed to change the location of a warehouse other than the one where the sales order was generated, this is because Adempiere works with well-established delivery rules that implement an algorithm according to the type of rule, the only rule that Allowing you to do this would be the forced delivery rule and it should even only apply to material from the same warehouse and not a different one.

This problem generally occurs when deliveries are made manually or the delivery rule is not taken into account.

Pino that we should not allow this since a lot of business logic would have to be modified, the recommendation is that the change of the warehouse is made in the sale order or a transfer of material between locations of different warehouses be made.

If all this business logic did not exist there would be no need to add the warehouse in the sales order, but ADempiere works with the warehouse in any inventory movement.

On the other hand, if I remember correctly, the quantity reserved is applied to the default location with instance 0 of each warehouse, so when it is delivered it should be updated in the same way.

SusanneCalderon:
Hola @e-Evolution ,
en realidad no estoy de acuerdo con este punto que no debe ser posible entregar desde otra bodega. Pero voy a cerrar este pull request para retomar el tema mas tarde.

e-Evolution:
Hi @SusanneCalderon  Susanne!

I understand your discrepancy, but it is how the system works in how it calculates the ordered, reserved and updates to existence.

To allow the movement of material from other warehouses, I think that a deeper analysis is required, since Compiere created the inventory movements to attend to these business cases and I created the distribution orders.

In controlled or orderly warehouse workflows it is not good practice for someone else to decide on my warehouse.

even your solution will still have to deal with the problems that the reserved and ordered would cause since they would not be matched correctly

To control the movement between warehouse or supply there is the functionality of:

1.- Inventory Movement between locations
2.- Distribution Orders
3.- Exit orders for collection

On the other hand, we will gladly talk in depth about the subject

and well with pleasure we discussed it thoroughly.

kind regards
Victor

Issue #3407, opened 2021-03-18
Taxdeclaration Consider DateAcct instead of dateInvoiced


SusanneCalderon:
## Feature request
As the tax declaration is a fiscal document, the considered date in the process create Tax Declaration should be dateAcct. 

#### Describe the solution you'd like
Include a process parameter to decide if the accounting date or the invoice date should be considered.

In process CreateTaxDeclaration consider dateacct ot dateInvoiced, depending of the parameter

PR #3408, opened 2021-03-18
Adempiere #3407 Taxdeclaration Consider DateAcct

SusanneCalderon:
Fixes https://github.com/adempiere/adempiere/issues/3407


e-Evolution:
@SusanneCalderon I tested and is ok, but I need you to help me by being consistent using only the first letter as uppercase for fields and only fully uppercase for SQL statements. please change PROCESSED to Processed

On the other hand, if your approach is only to include taxes based on the accounting date, do you should not also validate that the invoice is posted , so should you adding the validating the "Processed"  you should also validate the Posted = 'Y'?

carlosaparadam:
Hi @SusanneCalderon , i think that resolving the comments that write @e-Evolution  it will be fine.

Regards

SusanneCalderon:
@carlosaparadam 
Hi, Carlos, I included a parameter isOnlyPosted and corrected the sql Statement.

e-Evolution:
@SusanneCalderon please fix the description of new elements, now show all without spaces, I think we can be more explicit as Use Account Date to Created of declaration and Validate the use of document posted, also check the translation.

image

I think that the processed field should be always read-only 

image

and finality can merge develop with your branch to avoid any conflict.

SusanneCalderon:
@e-Evolution 
Hi, Victor, I updated the XML.
imagen
imagen

Issue #3392, opened 2021-03-09
Aging Report Enhance the table T_Aging more columns


SusanneCalderon:
## Feature request
Currently the Aging report shows the column till 90 days and >90 days. The financial report of Aging for sales invoices requires more columns, like 90 - 120 days, 120 - 180 days, 180 - 360 days, > 360.
#### Describe the solution you'd like
Enhance the table T_Aging with the columns 
- PastDue90_120		
- PastDue120_180
- PastDue180_360
- PastDue360_plus
- And ajust the class MAging to fill these ranges.

PR #3412, opened 2021-03-19
Adempiere  #3392 Aging Report Enhance the table T_Aging more columns

SusanneCalderon:
Fixes: #3392


e-Evolution:
@SusanneCalderon I saw that adding new columns into the T_Aging but not found the XML script with the modification the table in application dictionary

PR #3414, opened 2021-03-22
Fix distribution by organization dimension

carlosaparadam:
## Problem
When have an Accounting Distribution by Organization Dimension, this don't work.

## Step by Step

- Create a Accounting Distribution with this definition:
Screenshot_20210322_111056
Screenshot_20210322_111141

- find the Vendor Invoice 10000003
Screenshot_20210322_111213

- Post it document and see that the distribution don't work
Screenshot_20210322_111234

## Purpose
Change get Distribution method for comparing with Org_ID

image

## Result After Changes
Screenshot_20210322_111616

### DIFF
--- a/base/src/org/compiere/model/MDistribution.java
+++ b/base/src/org/compiere/model/MDistribution.java
@@ -112,7 +112,7 @@ public class MDistribution extends X_GL_Distribution
 			if (distribution.getC_DocType_ID() != 0 && distribution.getC_DocType_ID() != docTypeId)
 				continue;
 			//	Optional Elements - "non-Any"
-			if (!distribution.isAnyOrg() && distribution.getAD_Org_ID() != orgId)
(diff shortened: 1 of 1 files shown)


marcalwestf (approved):
1. Create on a **standard** Adempiere a _Distribution_ Rule similar to the one described.
   Condition: in the _Distribution Rule_ header be careful that the Organization of the Distribution Order is not the very same as the one for which the rule applies (* is OK)!!!! 
2. Create a _Purchase Invoice_ so the _Distribution Rule_ applies.
3. Complete the _Purchase Invoice_ and run the _Accounting_ process.
4. The _Distribution Order_ **is not executed**.
5. Now open the _Purchase Invoice_ and re-run the _Accounting_ process.
6. The _Distribution Order_ **is executed**.
7. Test passed.

PR #3417, opened 2021-03-30
Fixed error with wrong validation for production

yamelsenih:
## Step for reproduce
- Create a new Production without production batch
- Add product to produce and BOM
- Complete it

Screenshot_20210330_151701

Peek 2021-03-30 15-16

## What is the bug?
When a production is created without order it make a bad validation with
order

## What is the resolution?
Just validati if exist a production batch


marcalwestf (approved):
1. Reproduce case as described without changes (i.e. _Production Light_ without batch) -> error on execution of _prepare_ action.
2. Execute _prepare_ action with this commit  ->  no error.
3. Test passed.

PR #3428, opened 2021-04-10
Fix RoleAccessUpdate to enable update for all clients and all roles.

### LINKED ISSUE #3427: Role Access Update not working as expected
Log in as SuperUser or System under System Administrator role  OR GardenAdmin as GardenAdmin role.

Open the Role Access Update process

Set the Client as System
Set the Role to blank.

Run the process

Expected behaviour: all roles in all clients are updated and the process log lists them.

Actual behaviour: only the system administration role is updated.

Cause:  The addition of the process class structure with the abstract process class added some checks on the parameters.  The role id is an int and as a blank/null, it is changed from null to zero.  But zero is a valid role id for the system administration role and the clause ad_role_id=? is added to the where clause, restricting the list of roles to update to the system administration role.

Fix: either test for null or convert a null to -1 to allow all roles to be returned.

mckayERP:
Fixes #3427

Update RoleAccessUpdate to isolate constructors and static utilities to enable mocking for tests. Clean up the logic and fix the Main method.
In RoleAccessUpdateAbstract, change roleId to a BigDecimal and test for null in the getter.
Add unit and integration tests to show the application across all roles in all clients, all roles in the current client or a single role.

Unit and integration tests are included.

If the role is set, only that role is updated - even for the System Administration role
If the client is set to system, and role is not selected, all roles in all clients are updated
If the client is set other than system and the role is not selected, all roles in the selected client are updated

### DIFF
--- a/base/src/org/compiere/process/RoleAccessUpdate.java
+++ b/base/src/org/compiere/process/RoleAccessUpdate.java
@@ -21,96 +21,102 @@ import java.util.List;
 import java.util.logging.Level;
 
 import org.compiere.Adempiere;
+import org.compiere.model.I_AD_Role;
 import org.compiere.model.MRole;
 import org.compiere.model.Query;
 import org.compiere.util.CLogMgt;
 import org.compiere.util.CLogger;
 import org.compiere.util.Env;
+import org.eevolution.service.dsl.ProcessBuilder;
 
 /**
- *	Update Role Access
- *	
- *  @author Jorg Janke
- *  @version $Id: RoleAccessUpdate.java,v 1.3 2006/07/30 00:51:02 jjanke Exp $
+ * Update Role Access
+ * 
+ * @author Jorg Janke
+ * @version $Id: RoleAccessUpdate.java,v 1.3 2006/07/30 00:51:02 jjanke Exp $
  * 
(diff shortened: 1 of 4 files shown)


marcalwestf (approved):
Tested with Adempiere 393 seed + XMLs up to 6510.
Run process _Role Access Update_ with all 4 different combinations (Client vs Role set/unset).

The process behaves as described.

**Test passed**.

Missing still test on operative database which have more roles than GardenWorld.

marcalwestf:
This was the result of testing _GardenWorld_:
Selection_085

marcalwestf (commented):
Checking process behavior on client system with in total three clients (i.e. two clients in addition to GardenWorld).

The process behaves as described.

marcalwestf:
due to the fact that @yamelsenih  Yamel is a reviewer, I didn't accept this pull request.
Perhaps can Yamel test it and accept it, or tell me when I should do it.

yamelsenih:
Hi @marcalwestf @mckayERP I am testing it
Best regards

yamelsenih:
Hi @marcalwestf I taste it and is Ok for me. I approve to merge.

Thanks @mckayERP

Issue #3430, opened 2021-04-13
[Bug Report] [Bug Report] Transaction blocking error when creating account distribution and immediate posting


e-Evolution:
## Bug report
Transaction blocking error when creating account distribution and immediate posting

===========> Trx.commit: POSave_a2799278-89cf-4e32-b5fb-76330f48f04d [6595]
org.postgresql.util.PSQLException: FATAL: terminating connection due to administrator command
  Where: while locking tuple (108,1) in relation "c_bpartner"
SQL statement "SELECT 1 FROM ONLY "funsalud"."c_bpartner" x WHERE "c_bpartner_id" OPERATOR(pg_catalog.=) $1 FOR KEY SHARE OF x"; State=57P01; ErrorCode=0
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2412)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2125)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:297)
	at org.postgresql.jdbc.PgConnection.executeTransactionCommand(PgConnection.java:739)
	at org.postgresql.jdbc.PgConnection.commit(PgConnection.java:760)
	at com.mchange.v2.c3p0.impl.NewProxyConnection.commit(NewProxyConnection.java:803)
	at org.compiere.util.Trx.commit(Trx.java:336)
	at org.compiere.util.Trx.commit(Trx.java:364)
	at org.compiere.model.PO.save(PO.java:2199)
	at org.compiere.model.MAccount.get(MAccount.java:327)
	at org.compiere.model.MDistributionLine.getAccount(MDistributionLine.java:127)
	at org.compiere.acct.Fact.distribute(Fact.java:747)
	at org.compiere.acct.Doc.postLogic(Doc.java:795)
	at org.compiere.acct.Doc.post(Doc.java:651)
	at org.compiere.acct.Doc.postImmediate(Doc.java:358)
	at org.compiere.process.DocumentEngine.postImmediate(DocumentEngine.java:1290)
	at org.compiere.process.DocumentEngine.postIt(DocumentEngine.java:494)
	at org.compiere.process.DocumentEngine.processIt(DocumentEngine.java:322)
	at org.compiere.process.DocumentEngine.processIt(DocumentEngine.java:258)
[...]

PR #3431, opened 2021-04-13
 #3430 [Bug Report] Transaction blocking error when creating account distribution and immediate posting

e-Evolution:
…rom the account statement and immediate posting https://github.com/adempiere/adempiere/issues/3430


marcalwestf (approved):
- Without changes: case reproduced as described
- With changes: no blocking
- Test passed

PR #3436, opened 2021-04-15
Feature/#ad add grid sequence window custom

yamelsenih:
Just add a tab for grid sequence of ASP on Window customization
Screenshot_20210415_155529


marcalwestf (approved):
- XML loaded and applied.
- Open _Window Customization_. New tab pops up: _FieldGrid Sequence_.
- Create example Window Customization. Tab works.
- Generate Packout and export.

PR #3437, opened 2021-04-16
Fixed error with inconsistency of payment

yamelsenih:
## Step by Step
- Create a Payment and complete: Amount **USD 1500**
- Add payment to bank statement: Account balance **USD 1500**
- Reverse payment
- See the account balance: **USD 1500**
- See Payment Details (parameter Is Reconciled true) report: **USD
1500**

## The problem
After reverse it is created other payment as reversal (it is ok) the
problem is that the original payment and reversal is flagged as un
reconciled.

### What is the result?
- The Account balance is of **USD 1500**
- The Payment Details with flag Is Reconciled as true: **USD 0**

Both must match but is not really

## What is the expected behavior?
Both (Bank Balance and Payment details as reconciled) must be **USD
1500**, the reversed payment will should be flagged as not reconciled
instead and it can be available for reconciled with a new bank
statement.

## Other problem?
the allocate flag is not set after reversed for both payments
nevertheless, the payment allocation is created and completed

## The follows are some images of error:

Bank Balance after reconcile
Screenshot_20210415_200840

Report after reverse payment
Screenshot_20210415_200812

Screenshot_20210415_200648

### DIFF
--- a/base/src/org/compiere/model/MPayment.java
+++ b/base/src/org/compiere/model/MPayment.java
@@ -2228,7 +2228,7 @@ public final class MPayment extends X_C_Payment
 		MPeriod.testPeriodOpen(getCtx(), reversalDate, getC_DocType_ID(), getAD_Org_ID());
 
 		//	Auto Reconcile if not on Bank Statement
-		boolean reconciled = getC_BankStatementLine_ID() == 0; //AZ Goodwill
+		boolean reconciled = getC_BankStatementLine_ID() > 0; //AZ Goodwill
 
 		//	Create Reversal
 		MPayment reversal = new MPayment (getCtx(), 0, get_TrxName());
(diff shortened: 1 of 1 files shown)


marcalwestf (approved):
Test without changes
- Create a new bank + bank account
- Do put an initial amount to this bank account
- Create and complete a payment as described in this pull request.
- Create a Bank Statement and Include the payment in it; complete the bank statement.
- Reverse/void the Payment
- Result -> all documents are OK, only report  _Payment Details_ wrong according to description : isReconciled is wrongly displayed in report.


Test with changes
- Do the very same steps as without changes
- Result: all documents are OK **and also** report  _Payment Details_ **is OK** according to description : isReconciled is correctly displayed in report.

**Test passed**.

Issue #3309, opened 2021-01-11
GardenWorldCleanUp not working as expected


mckayERP:
The GardenWorldCleanUp process, which is run at the end of the migration loader and is available in the System menu, if failing.

It is supposed to move the dates in GardenWorld so that they appear to be more recent.  The intent was to ensure the GardenWorld demo data was somewhat up-to-date.

The process is failing as not all the dates were being updated.  New documents were not being updated as the GardenWorldCleanUp code did not account for them.  As a result, the Garden World calendar kept shifting until dates for some documents no longer fall within existing periods.

The code needs to be refactored so it does not depend on particular table models and can discover the documents that need to change.

marcalwestf:
this PR is a prerequisite to accepting https://github.com/adempiere/adempiere/pull/3434

PR #3442, opened 2021-04-20
Issue #3309 - Refactoring part 2: Move knowledge of the Doc classes to the Doc Classes

mckayERP:
The GardenWorldCleanup code needs to know which document columns represent the DateAcct to use when adjusting dates.  Several documents do not use the standard column name DateAcct.  When docs have been added that do not use the standard, the GardenWorldCleanup may not take these dates into account.

This commit reduces model dependencies in GardenWorldCleanup where the code needs to know which columns in the model to use for accounting dates.  This knowledge should be the responsibility of the Doc classes.  Add methods to the Doc class to determine the column name that acts as DateAcct.  Not all documents use the same
column name for this.  The Doc class provides an implementation for the getDateAcctColumnName which child classes can override if required.  Tests for these methods are provided.  The tests were put in the manufacturing model as the tests depend on the manufacturing document classes.

### DIFF
--- a/base/src/org/compiere/acct/Doc.java
+++ b/base/src/org/compiere/acct/Doc.java
@@ -17,6 +17,8 @@
 package org.compiere.acct;
 
 import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.math.BigDecimal;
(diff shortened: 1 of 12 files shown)


mckayERP:
Local Tests.

image

PR #3443, opened 2021-04-20
Issue #3309 - Refactoring part 3: Add methods to MPeriod to find min and max dates in calendar.

mckayERP:
Required by the GardenWorldCleanup process, these methods provide a way to find the min/max dates in a calendar.

Update MPeriod getC_Calendar_ID method to allow a different client id to be used.  This is necessary when the System account tries to access calendar dates from other clients.  Used in GardenWorldCleanup as the process is run using the system context.


marcalwestf (approved):
Check Code -> OK

Run test.
Result:
Selection_107


Test passed.

PR #3446, opened 2021-04-20
Set default value for centrally maintained menu

yamelsenih:
When a menu is flagged as summary level and also is flagged as centrally
maintained, the translation never is applied for him.

This change only ensure the the summary level is not central maintained

Note: please use squash for merge inside trunk


marcalwestf (approved):
Test:
- checked code ->  OK
- loaded and applied XML -> OK, all menu entries with _isSummary_ where set to _isCentrallyMaintained_=_N_ 
- created and saved a menu **summary** entry -> OK:  _isCentrallyMaintained_=_N_ 
- created and saved a menu **non-summary** entry -> OK:  _isCentrallyMaintained_=_Y_ 

Test passed

PR #3447, opened 2021-04-20
Fixed error with tender type for create payment selection from payment

yamelsenih:
When a payment selection is created automatically from Payment (after print) and the payment is created with Zelle or Mobile Payment, the payment rule of payment selection is created as Check instead **Direct Deposit** This pull request just allows create the payment selection with **Direct Deposit** for payments with tender types **Zelle** or **Mobile Payment**

Note: please use squash for merge inside trunk


marcalwestf (approved):
Tested  -> OK

PR #3450, opened 2021-04-26
Minor change for classpath

yamelsenih:
Just remove bad reference on classpath
Screenshot_20210426_172531


marcalwestf (approved):
Case reproduced:
Selection_121

And with changes: 
Selection_122

-> Test passed

Issue #3266, opened 2020-11-25
Error in calculating quantity ordered or reserved for manufacturing order and distribution order


e-Evolution:
## Bug report

The ordered and reserved quantity are duplicate when prepare it or complete a manufacturing order.

#### Steps to reproduce

1.- Create an Manufacturing Order 
2.- Change the Document status from Draft to In process, here create a first reserved and ordered.
3.- Change the Document status from  In process to complete, here create a second reserved and ordered.

#### Expected behavior
The ordered and reserved should be calculate rightly 

#### Other relevant information
- Your OS: MacOS
- Java version: 1.8
- ADempiere version: 3.9.3
- Browser safari
- DB Provider: PostgreSQL
- DB version: 12.3

yamelsenih:
Hi @e-Evolution this change add a new bug:

Step for reproduce:
- Create a Sales Order
- Add a new Line: Only with a Charge
- Complete it

The follow error is throwed:
```Java
java.lang.NullPointerException
	at org.compiere.model.MOrder.reserveStock(MOrder.java:1558)
	at org.compiere.model.MOrder.prepareIt(MOrder.java:1311)
	at org.compiere.process.DocumentEngine.prepareIt(DocumentEngine.java:411)
	at org.compiere.process.DocumentEngine.processIt(DocumentEngine.java:280)
	at org.compiere.process.DocumentEngine.processIt(DocumentEngine.java:258)
	at org.compiere.model.MOrder.processIt(MOrder.java:1169)
	at org.compiere.wf.MWFActivity.performWork(MWFActivity.java:861)
	at org.compiere.wf.MWFActivity.run(MWFActivity.java:781)
	at org.compiere.wf.MWFProcess.startNext(MWFProcess.java:341)
	at org.compiere.wf.MWFProcess.checkActivities(MWFProcess.java:256)
	at org.compiere.wf.MWFActivity.setWFState(MWFActivity.java:284)
	at org.compiere.wf.MWFActivity.run(MWFActivity.java:784)
	at org.compiere.wf.MWFProcess.startWork(MWFProcess.java:471)
	at org.compiere.wf.MWorkflow.start(MWorkflow.java:731)
	at org.compiere.wf.MWorkflow.startWait(MWorkflow.java:787)
```

PR #3452, opened 2021-05-05
#3266 Error in calculating quantity ordered or reserved for Sales Ord…

e-Evolution:
…er, manufacturing order and distribution order https://github.com/adempiere/adempiere/issues/3266

- Create new method to force same Warehouse for Sales Order Line but SO/PO
- Create new calculate order sizes
- Added new logic for that the Qty Reserved is only setting when the  product is Stocked

### DIFF
--- a/base/src/org/compiere/model/MOrder.java
+++ b/base/src/org/compiere/model/MOrder.java
@@ -1557,6 +1557,7 @@ public class MOrder extends X_C_Order implements DocAction
 		// Calculate Reserve Stock
 		orderLines.forEach(orderLine ->{
 			//	Force same WH for all but SO/PO
+			//	Check/set WH/Org
 			forceSameWarehouseToLine(headerWarehouseId, orderLine);
 			//	Binding
 			BigDecimal target = binding ? orderLine.getQtyOrdered() : Env.ZERO;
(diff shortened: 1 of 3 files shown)


base/src/org/compiere/model/MOrder.java
```diff
@@ -1501,6 +1501,42 @@ public boolean isBinding() {
...
+	}
+
+	/**
+	 * Calculate Sizes
+	 * @param orderLines Order Lines
+	 */
+	private void calculateOrderSizes(List<MOrderLine> orderLines) {
+		List<BigDecimal> volumenList = new ArrayList<>();
+		List<BigDecimal> weightList = new ArrayList<>();
+		orderLines.forEach(orderLine -> {
```
yamelsenih, line 1529: I don't know if is better use set for bigDecimal using a AtomiReference and add inside fiort forEach instead make two loops affter it

yamelsenih (changes requested):

base/src/org/compiere/model/MOrder.java
yamelsenih: Here a suggestion of code more clean without a unnecessary loop:

```Java
private void calculateOrderSizes(List<MOrderLine> orderLines) {
		setVolume(orderLines.stream().filter(orderLine -> orderLine.getM_Product_ID() > 0).map(orderLine -> {
			return Optional.ofNullable(orderLine.getProduct().getVolume()).orElse(Env.ZERO).multiply(Optional.ofNullable(orderLine.getQtyOrdered()).orElse(Env.ZERO));
		}).reduce(BigDecimal.ZERO, BigDecimal::add));
		setVolume(orderLines.stream().filter(orderLine -> orderLine.getM_Product_ID() > 0).map(orderLine -> {
			return Optional.ofNullable(orderLine.getProduct().getWeight()).orElse(Env.ZERO).multiply(Optional.ofNullable(orderLine.getQtyOrdered()).orElse(Env.ZERO));
		}).reduce(BigDecimal.ZERO, BigDecimal::add));
	}
```

PR #3377, opened 2021-03-02
Fix/#3309 fix garden world clean up

mckayERP:
Fixes #3309

The GardenWorldCleanup class was failing and not adjusting the dates of the demo client properly.  Some tables had been added to the system and the GardenWorldCleanup class had not been adjusted.  As a result, the class kept shifting the calendar but not moving these dates.  After a few years, the calendar no longer covered some of the dates and errors were being thrown.

This PR fixes the issue and reduces the fragility of the code by moving the responsibilities to the correct classes.  Specifically, the Doc class is responsible to know which column in a document is used for the DateAcct.  In most cases, this is standard, but there are a few which use other date columns.

Tests are added to make sure the Cleanup process works correctly and the integration tests run it on the current database.

In refactoring the code, some improvements were added to the Msg class - a wrapper for message keys to use in stead of "@" + message + "@", and some utility functions in MPeriod and TimeUtils.

PR #3455, opened 2021-05-07
Add Purchasing Tab to Business Partner

yamelsenih:
Screenshot_20210507_163653

Just a little change for add reference to **M_Product_PO** in Business
Partner when it is flagged as vendor
Here a gif
Purchase


mckayERP:
Dependent on PR #3377 to fix the build issue.

mckayERP (approved):
A good idea.   I recommend approval but I did not check the XML application.

PR #3458, opened 2021-05-18
Add the database function to get the daily salary

Ganzt90:
Add the database function to get the daily salary, based on the Employee (Business Partner), the payroll process and the date. The function returns the amount of the daily salary payroll concept that is defined in the employee's contract.

Payroll_Contract

### DIFF
--- a/db/ddlutils/oracle/functions/DailySalaryToDateByHrProcess.sql
+++ b/db/ddlutils/oracle/functions/DailySalaryToDateByHrProcess.sql
@@ -1,0 +1,53 @@
+CREATE OR REPLACE FUNCTION dailySalaryToDateByHRProcess
+(
+  p_C_BPartner_ID IN NUMBER,
+  p_HR_Process_ID IN NUMBER,
+  DateTo DATE
+)
+/******************************************************************************
+ * Product: Adempiere ERP & CRM Smart Business Solution                       *
+ * This program is free software; you can redistribute it and/or modify it    *
+ * under the terms version 2 of the GNU General Public License as published   *
+ * by the Free Software Foundation. This program is distributed in the hope   *
+ * that it will be useful, but WITHOUT ANY WARRANTY; without even the implied *
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.           *
+ * See the GNU General Public License for more details.                       *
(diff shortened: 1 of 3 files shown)


e-Evolution:
Hi Ganzt90

The field for Daily Salary (Concept) and Monthly Salary (Concept) are ids from Payroll Concept Catalog and it is used to know which concept is used to calculate the Daily Salary and Monthly Salary.

So this is not a value if not the record id, this reason because the field had the (Concept) word.

If for any reason you require the calculated values, I recommend adding a column calculated using the concept to bring the last value calculated from the employee's payroll.

yamelsenih:
Hi @e-Evolution this function allows get montly and daily salary based on payroll process historic instead current montly or daily salary.

The Employee salary is nice for know the current salary but if you want add it to payroll process and get on reports about historic is necessary some way for get it more easy.

This function is used for reports like Employee historic on Payroll Report.

Best regards

marcalwestf:
if I understand it well, this pull request just adds a function.
The function is called with the parameters Business Partner ID, HR ProcessID, and a date.
The date is actually the date up to which the function should search for any given BP in an HR process.

So it is no big issue to have such a function. In the worst case, is it a function one may not need to use.

I will test it on production data.

marcalwestf:
Tested the SQL of the function with operational data:
imagen

It delivers the _Amount_ of a _HR_Movement_ for an employee and a _HR_Process_. 
The _Contract_'s and the _HR_Movement_'s Concepts must match.
The date is to yield only amounts of movements before that date.

In short, this function can be used in certain situations, when a _HR_Movement_'s amount is to be known under certain circumstances. This is not delivered by another function, so it has a legit reason to be put forward.

marcalwestf (approved):
XML loaded and applied.
Function tested against productive data.

IMHO this PR is valid.

Test passed.

marcalwestf:
Build is not successfull.

marcalwestf:
@e-Evolution I think the original description by @Ganzt90  can be misunderstood, as it doesn't convey exactly the semantics of the function.

I think that just analyzing the function and its possibilities, it makes sense to accept this PR.

marcalwestf:
@Ganzt90  could you see to get the checks passed? Thank you.

marcalwestf:
@Ganzt90  could you please remove the conflicts?
Thanks!

PR #3460, opened 2021-05-20
Fixed error with multiplier for RV_C_InvoiceLine view

yamelsenih:
The view **RV_C_InvoiceLine** show all transaction for invoice line
depending of multiplier of document.

## What is the error?
The multiplier only affect to amounts but not for the follow columns:
- Weight
- Volume

## What is the fix?
Just multiply all columns with the multiplier


marcalwestf:
@yamelsenih  I loaded and applied the XML file; it worked well.

I have only one question regarding the meaning of the new negative columns.
I understand that Discounts and Margins should be displayed -same as amount- according to the multiplier because in a report, you can add the values and get the correct sum in the end.

But then, I ask myself what would be the reason to have PriceList, PriceActual, PriceLimit, PriceEntered, Weight, and Volume set also accordingly. To get the multiplication correct? You would not add unit Prices or unit Volumes in a report, would you? 

My question is only to know because these fields are actually only informative. Thanks, Yamel.

e-Evolution:
Hi Yamel  @yamelsenih 

I think it is wrong for any unit value to change based on the type of document or type of transaction, this does not happen anywhere in the system.

I can understand that the result of a total or aggregate amount should apply the factor to change the sign based on type transaction for views and reports ok, but not unit quantities.

So in this change, you should only change the sign for calculated or added fields as happens for credit notes.

marcalwestf (approved):
- Load and apply XML
imagen

- Test View
imagen

- Test reports on View (Daily Invoice, which acts on RV_C_Invoice_Day, which uses **rv_c_invoiceline**
imagen

- The following columns have a multiplier: qtyinvoiced,qtyentered, linenetamt, linelistamt, linelimitamt, linediscountamt,  lineoverlimitamt

marcalwestf:
@e-Evolution hi Victor, as you have some remarks about this PR, please feel free to give the final review.

Issue #3462, opened 2021-05-21
[Bug Report] Total matched invoice qty 


e-Evolution:
## Bug report
Error match invoice duplicates the record for receipt of material and invoice

when reversals were enabled for match invoice or match purchase order the three-way validation purchase order -> material receipt, material order receipt -> invoice now requires recording each match independently for each document compared to the previous approach where a single record was used to match receipt and invoice.

#### Steps to reproduce
1.- Create a new invoice 
2.- Create lines based on the purchase order

#### Screenshot or Gif

<img width="1344" alt="image" src="https://user-images.githubusercontent.com/786968/119193199-e822bd80-ba46-11eb-98c2-f6eb9e5b80ae.png">

<img width="848" alt="image" src="https://user-images.githubusercontent.com/786968/119193620-7eef7a00-ba47-11eb-9b71-995d3ace55ae.png">

#### Expected behavior
Only should be a record for match purchase should be created

#### Other relevant information
- Your OS: macos
- Java version: 1.8
- ADempiere version: 3.9.3
- Browser safari
- DB Provider: PG
- DB version: 11.2

yamelsenih:
Please can you add a step by step for reproduce this bug? 

The above steps is not completed for reproduce.
 I want to reproduce the big completely


Best regards

e-Evolution:
Steps to reproduce

1.- Create a new document Invoice

<img width="1030" alt="image" src="https://user-images.githubusercontent.com/786968/119276826-dfb2ba00-bbe1-11eb-9dbd-a440520f7ff3.png">

2.- Adding line Mulch by 500 Each

<img width="1039" alt="image" src="https://user-images.githubusercontent.com/786968/119276823-d6295200-bbe1-11eb-8b1e-00527275d472.png">

3.- Create a Material Receipt by partial quantity in this case 100 each

<img width="1057" alt="image" src="https://user-images.githubusercontent.com/786968/119276936-71bac280-bbe2-11eb-9cdb-38426d9aba5d.png">


4.- Create other Material Receipt  by 10 each 

<img width="1058" alt="image" src="https://user-images.githubusercontent.com/786968/119276912-4a63f580-bbe2-11eb-98c2-52e637813205.png">

5.- Generate the Invoice lines based on the Purchase Order

<img width="1043" alt="image" src="https://user-images.githubusercontent.com/786968/119277026-ea218380-bbe2-11eb-97b2-f339064f525e.png">


[...]

PR #3464, opened 2021-05-21
#3462 [Bug Report] Total matched invoice qty https://github.com/ademp…

e-Evolution:
…iere/adempiere/issues/3462

### DIFF
--- a/base/src/org/compiere/process/InvoiceCreateFrom.java
+++ b/base/src/org/compiere/process/InvoiceCreateFrom.java
@@ -72,7 +72,7 @@ public class InvoiceCreateFrom extends InvoiceCreateFromAbstract {
 			throw new AdempiereException("@CreateFromType@ @NotFound@");
 
 		//	Loop
-		recordIds.stream().forEach( key -> {
+		recordIds.forEach(key -> {
 			// variable values
 			int productId = getSelectionAsInt(key, "CF_M_Product_ID");
 			int chargeId = getSelectionAsInt(key, "CF_C_Charge_ID");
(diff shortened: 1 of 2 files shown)


base/src/org/compiere/process/InvoiceCreateFrom.java
```diff
@@ -112,22 +112,7 @@ protected String doIt() throws Exception {
 				MOrderLine orderLine = new MOrderLine (getCtx(), key, get_TrxName());
 				//	Set reference
 				referenceId.set(orderLine.getC_Order_ID());
-				//	Set InOut
-				String whereClause = "EXISTS (SELECT 1 "
-						+ "FROM M_InOut io "
-						+ "WHERE io.M_InOut_ID = M_InOutLine.M_InOut_ID "
-						+ "AND io.DocStatus IN ('CO','CL'))";
-				MInOutLine[] inOutLines = MInOutLine.getOfOrderLine(Env.getCtx(), key, whereClause, get_TrxName());
-				log.fine ("Receipt Lines with OrderLine = #" + inOutLines.length);
```
yamelsenih, line 121: This code allows allocate receipt from order when a order have receipt allocated.

How to think you resolve this?

yamelsenih (changes requested):

yamelsenih:
Hi @e-Evolution I think that this change make a bad implemetation for core, the original funcionality allows allocate the receipt with invoice and is a legacy funcionality.

Maybe is better if the original bug is resolved. I can make a local environment for test it tomorrow.

Best regards.

base/src/org/compiere/process/InvoiceCreateFrom.java
e-Evolution: The invoice match with purchase order was designed to validate the invoiced quantity with respect to the ordered quantity, so it should not be associated with a material receipt, for that it is the correspondence of receipt of materials with purchase order.

So these lines of code cause bad behavior.

The logic assigns the first receipt of materials that it finds even when it already has a previous match, in the case of partial or multiple receipts it generates the match for the same receipt of material causing a big problem with the matches, having to fix this manually which implies time and confusion in users

yamelsenih: Please put here a step by step for replicate it. This funcionality is legacy and has very little changes. I want to be sure of it

yamelsenih:
Hi @e-Evolution Please put here a step by step for replicate it. This funcionality is legacy and has very little changes. I want to be sure of it

marcalwestf:
@yamelsenih  Yamel, @e-Evolution  Victor provided the step-by-step reproduction instructions with issue https://github.com/adempiere/adempiere/issues/3462

marcalwestf:
@e-Evolution, @yamelsenih  as I understand this pull request, the error occurs when you:
1.- Complete a Purchase Order.
2.- Create an Invoice (Vendor) and use the Purchase Order to generate an Invoice Line, but not for the whole quantity ordered. 
3.- Complete the Invoice (Vendor).
4.- Stay in the Invoice (Vendor) window and create a Material Receipt. Complete the Material Receipt.
5.- Create another Invoice (Vendor) and use the Purchase Order to get an Invoice Line, but not for the whole quantity ordered. 

-> Here you can see that the Material Receipt Line of step 4 is referenced in the Invoice Line of the second Invoice, which is an error because there has not been a Material Receipt for this Invoice.

This behavior can be reproduced with standard Adempiere and it is suppressed with this Pull request.

marcalwestf:
...the deleted code that @yamelsenih asks to be explained is in my opinion dangerous because it gets the first material receipt line that matches the invoice quantity or the first material receipt line if none matches. 

This means if there is already one material receipt line and the quantity doesn't match, it will be referenced. And this is an error.
So in my opinion is OK to eliminate the reference to MInOutLine.

marcalwestf (approved):
- error was reproduced on standard Adempiere
- error was fixed with changes by this pull request

-> Test passed

marcalwestf:
@e-Evolution  Victor, there are some checks pending
Thanks

@yamelsenih perhaps you can review also the changes and cast your opinion.

yamelsenih:
I agree, just update this branch with last changes for approval from CI

yamelsenih (approved):

PR #3472, opened 2021-05-24
Error in calculating quantity ordered or reserved for manufacturing order and distribution order 

e-Evolution:
Related with #3266

- Remove unnecessary code
- Improvement of the code separating the behavior of calculation of weight and volume of the reservation process.

A good practice is that a function method in the code should be responsible for only one behavior

replace #3452

### DIFF
--- a/base/src/org/compiere/model/MInOut.java
+++ b/base/src/org/compiere/model/MInOut.java
@@ -1459,9 +1459,9 @@ public class MInOut extends X_M_InOut implements DocAction , DocumentReversalEna
 			}	//	stock movement
 
 			//	Correct Order Line
-			if (product != null && orderLine != null && isSOTrx() && !orderLine.getParent().isReturnOrder())		//	other in VMatch.createMatchRecord
+			if (product != null && product.isStocked() && orderLine != null && isSOTrx() && !orderLine.getParent().isReturnOrder())		//	other in VMatch.createMatchRecord
 				orderLine.setQtyReserved(orderLine.getQtyReserved().add(QtySO));
-			else if (product != null && orderLine != null && !isSOTrx() && !orderLine.getParent().isReturnOrder())
+			else if (product != null && product.isStocked() && orderLine != null && !isSOTrx() && !orderLine.getParent().isReturnOrder())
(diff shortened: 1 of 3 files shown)


marcalwestf:
@e-Evolution  hi Victor, this seems to be an important issue for Libero Manufacturing.

Could you please write a step-by-step way to test it?
As I understand from your explanations, the behavior as of now should not change, even if calling _calculateOrderSizes()_ has been moved and its implementation redefined.

Thanks!

marcalwestf:
... and the build conflict solved...

e-Evolution:
> @e-Evolution hi Victor, this seems to be an important issue for Libero Manufacturing.
> 
> Could you please write a step-by-step way to test it?
> As I understand from your explanations, the behavior as of now should not change, even if calling _calculateOrderSizes()_ has been moved and its implementation redefined.
> 
> Thanks!

@marcalwestf Hi Mario This change applies a best practice in code development, based on the SOLID responsibility principle, a method should only have one purpose or responsibility, so the reservation method not should also calculate the total weight and volume.

So this pull request separates the logic of calculating weight and volume from the reservation method to an independent method and calls the method after the quantities were reserved, it also removes commented code that should not be more.

To validate the tests you must follow the following steps:

1.- Find a product and add weight and volume
2.- Create a sales order for the product with volume and weight
3.- Complete the sales order
4.- Validate that the reservation was made correctly in the inventory.
5.- Validate that the weight and volume for the order is correctly calculated

yamelsenih:
Hi @e-Evolution @marcalwestf I approve this, the old PR was very wrong but @e-Evolution change this following my comments.

Best regards

marcalwestf:
@e-Evolution  Victor, could you please solve the conflicts in order to be able to accept the PR?
Thanks

marcalwestf (approved):
- Create a product with Weight and Volume
- Do an Inventory or a Material Receipt on this product
- Complete a Sales Order with this product
- Check the Volume and the Weight in the Sales Order Header
- Do the same with orders having several products with Weight and Volume defined

The Volume and the Weight in the Sales Order Header are according to the Weight/Volume defined and the quantities of the Order Lines.

-> Test passed.

Issue #3473, opened 2021-05-24
[Feature Request] Adding the Weight and Volume field into Sales Order and Purchase Order


e-Evolution:
## Feature request
Currently the Weight and Volume are calculate but this not are show in the Sales Order and Purchase Order

#### Describe the solution you'd like

Sales Order with Weight and Volume  fields
![image](https://user-images.githubusercontent.com/786968/119405127-68932980-bca6-11eb-8cd7-28ee40c7ce16.png)

Purchase Order Sales with Weight and Volume  fields

![image](https://user-images.githubusercontent.com/786968/119405404-d3dcfb80-bca6-11eb-9a33-e9ab6ea6d144.png)

PR #3474, opened 2021-05-24
Adding the Weight and Volume field into Sales Order and Purchase Order  Feature/3473

e-Evolution:
implementing change to #3473


marcalwestf:
... this is only an XML file for displaying Weight and Volume on Purchase Orders and Sales Orders.

e-Evolution:
> ... this is only an XML file for displaying Weight and Volume on Purchase Orders and Sales Orders.

Hello @marcalwestf Mario, if this is correct, this XML is an improvement that adds the total weight and volume in the Purchase and Sales order Window, since it is currently being calculated in the business logic, it seems very convenient to show this information.

PR #3476, opened 2021-05-24
Add support to data type as mapping from entities

yamelsenih:
# What is the problem?
When all data is generated from ADempiere to ElasticSearch, not exist a way for determinate the data type for object and isn necessary run a rebuild based on proxy-adempiere-api. This method is very unuseful because all data type send from ADempiere should be sent with default types like keyword for search or date for sorting based on [ElasticSearch Data Type](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html)


marcalwestf (approved):
Compare changes in source code -> seem OK.

As we are not able to test the impact of these changes in _Vue Store Front_, and the project _Vue Store Front_ is enduring ongoing changes, the changes are accepted on this basis.

PR #3480, opened 2021-05-27
Fixed error with Dependent Entities for Record Access

yamelsenih:
## What is the bug?
When a record is add to record access for exclude and using dependent
entities flag it work fine for windows but not work for reports.

### Setup Role
- Login into ADempiere as **GardenAdmin** role
- Go to Role
- Lookup **GardenWorld Admin**
- Set to true the flags: **Personal Lock** and **Personal Access**
Lock
- Re-Login
### Recreate Bug
- Login into ADempiere as **GardenAdmin** role
- Go to **Document Type**
- Lookup **POS Order** record
- On Tool Bar press **Ctrl** + **Lock Button**
- On Record Access Dialog:
  - Press new record
  - Set flag **Exclude** and **Dependent Entities** to true
  - Press **Ok** button
- Logout
- Note: If you open **Sales Order** window you can see 5 Orders:
  - 3 as **POS Order**
  - 1 as **On Credit Orde**
  - 1 as **Standard Order**
Screenshot_20210527_113652
### Verify Bug
- Login into ADempiere as **Garden User**
- Go to **Sales Order** window, now you can see only:
  - 1 as **On Credit Orde**
  - 1 as **Standard Order**
 Screenshot_20210527_113958
- Go to **Quote-to-Invoice** -> **Sales Orders** -> **Order Detail**
- Set flag **Sales Transaction** to true
- Run report
- Open Print Format and remove **Document Type** column
- Note that you can see 5 **Sales Orders** instead 2

### DIFF
--- a/base/src/org/compiere/model/MRole.java
+++ b/base/src/org/compiere/model/MRole.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.Vector;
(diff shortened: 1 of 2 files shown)


marcalwestf (approved):
- follow lock process as described
- before PR: report displays locked records
- after PR: report **doesn't display** locked records

Test passed

PR #3481, opened 2021-05-28
Bad reference for price list create process

yamelsenih:
When the process "Create Product Price" on Price List Version is runned
using a price list discount schema based on Product PO, the price used
for calculate is the **Price List** instead **Price PO**. This have a
problem because the real price is based on Price PO instead Price List
of Product PO.

### DIFF
--- a/base/src/org/compiere/process/M_PriceList_Create.java
+++ b/base/src/org/compiere/process/M_PriceList_Create.java
@@ -375,8 +375,8 @@ public class M_PriceList_Create extends M_PriceList_CreateAbstract {
 						combinationMap.put(productPurchasing.getM_Product_ID(), 
 								new ProductCombination(productPurchasing.getM_Product_ID(), 
 										conversionRate, 
-										productPurchasing.getPriceList(), 
-										productPurchasing.getPriceList(), 
+										productPurchasing.getPricePO(), 
+										productPurchasing.getPricePO(), 
 										productPurchasing.getPricePO(), 
 										discountSchemaLine));
 					});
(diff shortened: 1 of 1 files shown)


e-Evolution:
@yamelsenih Hi Yamel

I have reviewed this commit and I think that the price should be set as follows:

```
//	Put combination
		combinationMap.put(productPurchasing.getM_Product_ID(), 
				new ProductCombination(productPurchasing.getM_Product_ID(), 
						conversionRate, 
						productPurchasing.getPriceList(), 
						productPurchasing.getPricePO(),
						productPurchasing.getPricePO(), 
						discountSchemaLine));

```

Since in the product purchase screen if the list price exists and the method for ProductCombination asks for the parameters:

```/**
		 * Default constructor
		 * @param productId
		 * @param currencyId
		 * @param priceStandard
		 * @param priceList
		 * @param priceLimit
		 * @param discountSchemaLine
		 */
		public ProductCombination(int productId, BigDecimal conversionRate, BigDecimal priceList, BigDecimal priceStandard, BigDecimal priceLimit, MDiscountSchemaLine discountSchemaLine) {
			this.productId = productId;
			this.conversionRate = conversionRate;
			this.priceList = Optional.ofNullable(priceList).orElse(Env.ZERO);
			this.priceStandard = Optional.ofNullable(priceStandard).orElse(Env.ZERO);
			this.priceLimit = Optional.ofNullable(priceLimit).orElse(Env.ZERO);
			this.discountSchemaLine = discountSchemaLine;
		}

```

image


Since for the limit price there is no field in the product purchases tab, you could use the last PO price as the minimum reference price but I suppose in countries with a lot of inflation it is not a good idea. 

but it would be a criterion to consider

kind regards
Victor

yamelsenih:
Hi @e-Evolution I think that also can be validate if not exist Price List then use price PO also

PR #3482, opened 2021-05-28
Fixed error with Role Access Update for System Administrator Role

yamelsenih:
## Bug
After merge #3428 the **System Administrator** role is wrong after run
process **Update Role Access** as System Administrator.

## What is the problem?
The system **Administrator Role** is was create with **AD_Role_ID = 0**,
when the role is create with **MRole** constructor, by default is
created a new instance instead find the role.

## Step by Step
- Login as **System Administrator**
- Run process **Update Role Access**
- Logout
- Login after
- Try Open **Table and Column** window


yamelsenih:
Hi @mckayERP I will to change for resolve the test for role. If you have any help is nice

mckayERP:
@yamelsenih Good catch.  Yes, I'll have a look.

yamelsenih:
Thanks @mckayERP please, if you can see the test class for resolve this can be so many helping for me

e-Evolution:
@yamelsenih Hi Yamel you can update the branch, since it presents errors in the building

marcalwestf (approved):
Do as described without changes:
image

With changes: the window _Table and Column_ is displayed correctly.
Test passed.

PR #3484, opened 2021-05-31
Remove deprecated methods for Order

yamelsenih:
Just a minor change removing deprecated method for Order business model


marcalwestf (approved):
Created Purchase and Sales Orders.
Execute the following: Copy From, Prepare, Complete, Void, Close, Create Shipment, Create Invoice, Re-Open.

Behavior as expected.

Test passed.

Issue #3489, opened 2021-06-09
[Bug Report] Window footer shows number of lines in null (ContextInfo)


EdwinBetanc0urt:
#### Steps to reproduce
1. Open any of the windows `Expense Report`, `Invoice Batch`, `Invoice (Vendor)`, `Invoice (Customer)`, `Purchase Order`, `Sales Order`, and create record.
2. Create lines.

#### Screenshot or Gif

![footer-expense-report-n-lines-error](https://user-images.githubusercontent.com/20288327/121292614-41e50d80-c8b8-11eb-9b0e-af22c3afa779.png)
![footer-invoice-batch-error](https://user-images.githubusercontent.com/20288327/121292630-45789480-c8b8-11eb-91f1-f4a0d0461aa7.png)
![footer-invoice-n-lines-error](https://user-images.githubusercontent.com/20288327/121292638-4a3d4880-c8b8-11eb-8d6a-9fc2bb9f7a7d.png)
![footer-order-n-lines-error](https://user-images.githubusercontent.com/20288327/121292644-4dd0cf80-c8b8-11eb-87ec-b8093dee54e9.png)

#### Other relevant information
- Your OS: Linux Mint 19.1 Cinnamon x64.
- Java version: 1.8.0_281.
- ADempiere version: 3.9.3
- Browser: Mozilla Firefox.
- Browser Version: 88.0.1.
- DB Provider: PostgreSQL.
- DB version: 13.5.

PR #3490, opened 2021-06-09
fix: NULL line counter in ContextInfo.

EdwinBetanc0urt:
Before the cast (lines is bigint)
output-invoice-error

After the cast (lines is integer)
output-invoic-fix

Corrections at the footer of the windows with this PR:

footer-expense-report-n-lines-fix

footer-invoice-batch-fix

footer-invoice-n-lines-fix

footer-order-n-lines-fix

fixes https://github.com/adempiere/adempiere/issues/3489


yamelsenih:
Hi @EdwinBetanc0urt I don't see the problem here.

Can you show the old behavior?

yamelsenih:
I don't sure if is better use the cast on context value instead

EdwinBetanc0urt:
Hi @yamelsenih in the issue this is how it currently looks like with Null where the number of lines should be shown https://github.com/adempiere/adempiere/issues/3489

EdwinBetanc0urt:
> I don't sure if is better use the cast on context value instead

You could add a condition that evaluates the long data type, because if it is not evaluated, it arrives to the continue statement omitting the expected column, and showing Null in the place of the value returned by the count.

https://github.com/adempiere/adempiere/blob/develop/base/src/org/spin/model/MADContextInfo.java#L246

```java
	Object value = rs.getObject(column + 1);
	if(!(value instanceof BigDecimal)
			&& !(value instanceof Double)
			&& !(value instanceof Float)
			&& !(value instanceof Integer)
			&& !(value instanceof String)
			&& !(value instanceof Timestamp)) {
		continue;
	}
	//	
	if(value instanceof BigDecimal) {
		value = ((BigDecimal) value).doubleValue();
	} else if(value instanceof Timestamp) {
		value = new Long(((Timestamp) value).getTime());
	} else if(value instanceof Number) {
		value = new Integer(((Number) value).intValue());
	} else {
		value = String.valueOf(value);
	}
	//	Set value
	arguments[column] = value;
```

yamelsenih:
Maybe the problem is here https://github.com/adempiere/adempiere/blob/develop/base/src/org/spin/model/MADContextInfo.java#L246 and not in the query.

EdwinBetanc0urt:
long-value-context-info


sales-order-status-bar

@yamelsenih Corrected in the source code by evaluating the long returned in the query

base/src/org/spin/model/MADContextInfo.java
```diff
@@ -247,6 +250,7 @@ public static MADContextInfo getFromWindowId(Properties ctx, int windowId) {
 							&& !(value instanceof Double)
 							&& !(value instanceof Float)
 							&& !(value instanceof Integer)
+							&& !(value instanceof Long)
```
yamelsenih, line 253: Please validate it with Type of DB

yamelsenih (changes requested):

base/src/org/spin/model/MADContextInfo.java
EdwinBetanc0urt: has already been validated with DB data type

yamelsenih:
Thanks.  Testing now

yamelsenih:
I agree for merge but I am waiting for all queued PRs before

e-Evolution:
Hi  @EdwinBetanc0urt I can see that this class is auxiliary, it seems to me that the approach is wrong since it would have to validate all the types according to the database it is used, in this case I only see support for PostgreSQL,

What about for example Number Oracle Type?

ADempiere must be compatible with at least Oracle and PostgreSQL.

Maybe a better approach is to use DisplayType based on dictionary reference for column to determine cast of value.

https://github.com/adempiere/adempiere/blob/55dcd48f9df2bfecd0e691fb6dccfe1735fff72e/base/src/org/compiere/model/PO.java#L1262

EdwinBetanc0urt:
Hi @e-Evolution, thanks for your recommendations, doing the validation with the DisplayType I don't think is the way, since some columns result from functions, or if you change the name of the column with an alias, maybe they are not saved in the application dictionary as can be Lines and ConvAmt, in the case of the contextInfo for the invoice: 

```sql 
COUNT (*) AS Lines
```

```sql 
currencyBase(o.GrandTotal, o.C_Currency_ID, o.DateAcct, o.AD_Client_ID, o.AD_Org_ID) AS ConvAmt
```

I made a comparison of java.sql.Types with oracle.jdbc.OracleTypes and the difference between them with the data types that are of interest is:

- java.sql.Types:
  - TIME_WITH_TIMEZONE = 2013;
  - TIMESTAMP_WITH_TIMEZONE = 2014;

- oracle.jdbc.OracleTypes
  - TIMESTAMPNS = -100; (deprecated)
  - TIMESTAMPTZ = -101;
  - TIMESTAMPLTZ = -102;

The values present in the current comparison are the same in both cases, I could add the 3 missing values that OracleTypes uses, following the current approach, if you agree 

@yamelsenih @e-Evolution

marcalwestf:
@e-Evolution  , @yamelsenih  any progress here?

EdwinBetanc0urt:
Hi @yamelsenih @marcalwestf @e-Evolution, there here are 3 commits solving the problem I report, each in a different way:

- This one correcting through sql cast https://github.com/adempiere/adempiere/pull/3490/commits/d02951072085e42e769df74a23bbf52622ad3001.

- This one taking into account the logvalues (in my opinion the most viable for this case) https://github.com/adempiere/adempiere/pull/3490/commits/752c3e8816769e2fca35d6b4cff22d1f900521d5

- This validates the data type of the column in the database https://github.com/adempiere/adempiere/pull/3490/commits/e2665571b393e201de216951a6f738f4a1bbf570

I consider that the error should be corrected first and then change the implementation approach, knowing that at least this functionality should be operative.

EdwinBetanc0urt:
Greetings @e-Evolution @marcalwestf @yamelsenih , are there any comments on how to move forward on this issue? I suggest resolving the bug first and then refactoring to the approach that is best considered.

e-Evolution (approved):
Current behavior without this bugfix

image

in the status bar show null Lines(s)

apply the patch 

image

now the status bar show ok

PR #3491, opened 2021-06-10
Fixed error with Crosstab for Smart Browser

yamelsenih:
## Bug
The problem is that for crosstab based on List not exist a cast from
String to value, it is when try compare a string without **'**

## Step by Step
- Go to POS Cash Closing Smart Browser
- Run it

Best Regards

### DIFF
--- a/client/src/org/eevolution/grid/Browser.java
+++ b/client/src/org/eevolution/grid/Browser.java
@@ -1227,7 +1227,13 @@ public abstract class Browser {
 										.getColumnName()).append("=")
 								.append(fieldKey.getAD_View_Column().getColumnSQL())
 								.append(")");
-					} else if(!Util.isEmpty(parentcolumnName.get())){
+					} else if(!Util.isEmpty(parentcolumnName.get())) {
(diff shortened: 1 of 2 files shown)


e-Evolution:
Hi  @yamelsenih Yamel

I have reviewed the code and I think that the validation of the reference should be with the one defined in the browser field instead of using that of the column because in the end the control of the reference and as shown is defined in the browser field and it could be different from the one defined in the column. 


Proposed change:
```
		MColumn column = MColumn.get(field.getCtx(), xColumn.getAD_Column_ID());
		String valueToReplace = "@Axis_Record_ID@";
		if(DisplayType.isText(field.getAD_Reference_ID())
				|| DisplayType.List == field.getAD_Reference_ID()) {
			valueToReplace = "'@Axis_Record_ID@'";
		}
```

kind regards
Victor

yamelsenih:
Hi @e-Evolution the validation is based on field of Smart browser referenced to column, you can see the change and validate it with cash closing and add document action access from role.

yamelsenih:
Let me see your propose also

Issue #3506, opened 2021-07-09
[Bug Report] Display rule does not work correctly between records when browsing in grid mode


e-Evolution:
## Bug report
Display rule does not work correctly between records when browsing in grid mode

#### Steps to reproduce

1.- Order that shows the button to process the order 

<img width="1619" alt="image" src="https://user-images.githubusercontent.com/786968/125171022-95789e80-e177-11eb-90e2-017831c06c96.png">

2.- Open the window of sales orders for multiple records

3.- Select an order that complies with behavior based on the display rule

![image](https://user-images.githubusercontent.com/786968/125170738-028b3480-e176-11eb-9b57-f2c585b53f75.png)

4.- Select a new record where the display rule should take the context of the new record
![image](https://user-images.githubusercontent.com/786968/125170813-7a595f00-e176-11eb-8c9a-3a88cadbeaed.png)

5.- The new registry does not update the behavior based on the display rule for the record.

![image](https://user-images.githubusercontent.com/786968/125170890-dde38c80-e176-11eb-8160-4b4e36515817.png)

Screenshot:

Configuration to replicate the issue:
1.- Create a calculated column into C_Order table for SOCreditStatus using this SQL (SELECT SOCreditStatus FROM C_BPartner WHERE C_BPartner.C_BPartner_ID=C_Order.C_BPartner_ID)

<img width="1543" alt="image" src="https://user-images.githubusercontent.com/786968/125170595-4e89a980-e175-11eb-9604-fc1ba7357eb3.png">

2.- Adding the SOCreditStatus into the Sales Order Window Header

![image](https://user-images.githubusercontent.com/786968/125170636-7842d080-e175-11eb-8f47-57f4590a5e5c.png)

3.- Adding the Display rule @SOCreditStatus@='O' for Process Order:

![image](https://user-images.githubusercontent.com/786968/125170683-a88a6f00-e175-11eb-86ea-9d3b74e6ef9f.png)

#### Expected behavior
If the grid is selected and navigating between records, the display rule should be applied for the new record, as it works in simple record mode and when navigating between records.

#### Other relevant information
- Your OS: MacOS
- Java version: JDK 8 and 11
- ADempiere version: 3.9.3
- Browser safari
- DB Provider: PG
- DB version: !·

PR #3508, opened 2021-07-09
[Bug Fixed] Display rule does not work correctly between records when browsing in grid mode for #3506

e-Evolution:
bugfix to #3506


yamelsenih:
Hi @e-Evolution can you add here a gif with behavior?

Best regards

marcalwestf:
**Test Scenario I**

- According to @e-Evolution this occurs only on **ZK**, so the test will be made on ZK. 
- Define in _Application Dictionary_ for window _Sales Order_, tab _Order_, Field _Process Order_ a _Display Logic_ @GrandTotal@<100. This would display the Field _Process Order_ (i.e. the Process button) only if the _GrandTotal_ is less than 100.
image
- Log in as _GardenWorld Admin_ in _GardenWorld_, open _Sales Order_ window.
- Create an Order that complies with the _Display Logic_ and another that doesn't comply with the _Display Logic_.
- Open the Sales Order that complies with the Display Logic.
- Change to the Grid view and go to the Order that doesn't comply with the Display Logic.
- Open this Order (which doesn't comply) with the rule.
- Result: **the button is not displayed**, as expected.

image
image


Summary: with a simple _Display Logic_ rule (@GrandTotal@<100), this case cannot be reproduced.

Another scenario has to be constructed to reproduce this error.

marcalwestf:
Test Scenario II

- Test on Standard Adempiere.
-  Create in table _C_Order_ a SQL Column with the Element _SOCreditStatus_.
  The SQL Column value is _(SELECT SOCreditStatus FROM C_BPartner WHERE C_BPartner.C_BPartner_ID=C_Order.C_BPartner_ID)_
image

- For window _Sales Order_, tab _Order_, bring Field _Credit Status_.
- For window _Sales Order_, tab _Order_, Field _Process Order_, create a _Display Logic_ = @SOCreditStatus@='O'.
- Log in as _GardenWorld Admin_ in _GardenWorld_, open _Sales Order_ window.
- Configure _Business Partners_ so that one complies and the other doesn't complywith the _Display Logic_ .
- Create an _Order_ that complies with the _Display Logic_ and another that doesn't comply with the Display Logic.
- Open the _Sales Order_ that complies with the Display Logic.
- Change to the _Grid_ view and go to the _Order_ that doesn't comply with the Display Logic.
- Open this Order (which doesn't comply) with the rule.
- Result: the button is not displayed, as expected.
  
image


Doing the same steps on branch e-Evolution:feature/3506 the same result is achieved -> **at least under the scenarios and circumstances described, this error cannot be reproduced neither on Standard Adempiere nor on branch e-Evolution:feature/3506**.

BUT
when looking at the _Context_, the TabInfo shows a **wrong** _C_Order_ID_ for getting it:
image
image

This means

- I cannot reproduce the error under the test scenarios described
- It is something wrong with the context.

At @e-Evolution Victor, is there any other way this case can be reproduced?

marcalwestf:
at last, and thanks to @e-Evolution Victor's help, I was able to reproduce the error and verify the fix.

For this, you must

- Display in  _Single-row_ mode a record with Display Rule = true
- Then, change to _Grid_ mode and **select with the mouse!!!** a record which Display Rule=false
- The "old" Display Rule applies to the new record, which is wrong.

marcalwestf (approved):
- Test on ZK
- Reproduce the use case with standard Adempiere -> Display Rule doesn't apply
- Reproduce the use case with this PR -> Display Rule applies

Result: test passed.

PR #3509, opened 2021-07-09
Add support to group, class and classification names

yamelsenih:
- Add Name for product group
- Add Name for product class
- Add Name for product classification
- 
Screenshot_20210709_180548


SusanneCalderon:
@yamelsenih , @marcalwestf 
I tested with product class name, product group name and product classification name.
New Product Group Product Class and Product Classification were generated correctly.
Already existing group/class/classification were not identified by the name (missing value), but were identified correctly by value.

PR #3511, opened 2021-07-11
Create Index on Migration Data

carlosaparadam:
Add a index on migration data for optimize migration process


marcalwestf:
@carlosaparadam Hi Carlos, a small observation:
imagen
Sequence in File Name and in XML file do not match.
They should be both 7910.

Actually, the sequence should be **7920** or **7940**, because 7910 has been already committed.

marcalwestf (approved):
After loading and executing the XML file:
imagen
imagen

Test passed.

marcalwestf (approved):
According to newest changes

PR #3519, opened 2021-07-12
Invoice with price list that includes taxes

nsarlabos:
- The system determines if it is a list with tax included by means of a variable that is loaded when saving the head of the invoice, for which the price list to select must have previously marked the check of tax included as appropriate, if it is marked after saving the invoice header will have no effect.

- Tax rates must have the "Document Level" check unmarked

### DIFF
--- a/base/src/org/compiere/acct/Doc_Invoice.java
+++ b/base/src/org/compiere/acct/Doc_Invoice.java
@@ -50,6 +50,10 @@ import org.compiere.util.Env;
  *  	<li>BF: 2797257	Landed Cost Detail is not using allocation qty
  *  
  *  @version  $Id: Doc_Invoice.java,v 1.2 2006/07/30 00:53:33 jjanke Exp $
+ *
+ *	@author Nicolas Sarlabos, nicolas.sarlabos@openupsolutions.com, http://www.openupsolutions.com
+ *			<li> FR [ 1459 ] Invoice with price list that includes taxes
+ *			@see https://github.com/adempiere/adempiere/issues/1459
  */
 public class Doc_Invoice extends Doc
 {
@@ -166,6 +170,7 @@ public class Doc_Invoice extends Doc
 			docLine.setQty(cm ? Qty.negate() : Qty, invoice.isSOTrx());
 			//
 			BigDecimal LineNetAmt = line.getLineNetAmt();
+			BigDecimal LineTotalAmt = line.getLineTotalAmt();
 			BigDecimal PriceList = line.getPriceList();
 			int C_Tax_ID = docLine.getC_Tax_ID();
 			//	Correct included Tax
@@ -174,9 +179,9 @@ public class Doc_Invoice extends Doc
 				MTax tax = MTax.get(getCtx(), C_Tax_ID);
 				if (!tax.isZeroTax())
 				{
-					BigDecimal LineNetAmtTax = tax.calculateTax(LineNetAmt, true, getStdPrecision());
+					BigDecimal LineNetAmtTax = tax.calculateTax(LineTotalAmt, true, getStdPrecision());
 					log.fine("LineNetAmt=" + LineNetAmt + " - Tax=" + LineNetAmtTax);
-					LineNetAmt = LineNetAmt.subtract(LineNetAmtTax);
+					LineNetAmt = LineTotalAmt.subtract(LineNetAmtTax);
 					for (int t = 0; t < m_taxes.length; t++)
 					{
 						if (m_taxes[t].getC_Tax_ID() == C_Tax_ID)
--- a/base/src/org/compiere/model/CalloutInvoice.java
+++ b/base/src/org/compiere/model/CalloutInvoice.java
@@ -34,6 +34,9 @@ import org.compiere.util.Env;
  *	
  *  @author Jorg Janke
  *  @version $Id: CalloutInvoice.java,v 1.4 2006/07/30 00:51:03 jjanke Exp $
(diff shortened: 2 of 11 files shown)


marcalwestf:
@nsarlabos  hi Nicolás, could you please write a step-by-step description of how to reproduce the error and the solution?
Thanks

nsarlabos:
Hi @marcalwestf , I attach two GIFs, where in the first one I show the step by step showing the error, and in the other the same but after the fix.

**Error:**

Error

**Solution:**

Solution

nsarlabos:
Hello @SusanneCalderon  I did the tests in "develop" and I understand that the operation is not correct. In the case of invoices, although the tax is well calculated in the tax tab, the line total field is left with a higher amount (although this field is not shown in the documents receivable window, but it is shown in documents for pay), and in the header the subtotal is equal to the total, when the subtotal should be the net WITHOUT taxes.

In this example, the price of the product "Fertilizer # 50" is USD 122, in the price list with taxes included, and it has associated the tax type "Basic" with a rate of 22%.

DxC

DxP

In the case of orders, I have noticed the same details as in the invoices, and furthermore, later the invoices created from order do not have their amounts right since the tax is recalculated.

DxP desde orden

We understand that the correct behavior is that shown in the GIF under the heading "Solution" (attached above).

Regarding your query about the "Document Level" check in the tax rate, really in our case it has not presented us with problems with our clients, but we are open to evaluating the changes that you believe necessary.

SusanneCalderon (changes requested):
We could reproduce the error. But with a correction of the method amt in CalloutInvoice you do not have to change all the other classes.


Line 685, before change
```

			//	Add it up
			mTab.setValue("LineTotalAmt", lineNetAmount.add(TaxAmt));
		//}


```
Line 685, correction

```
//	Add it up
			if (isTaxIncluded(WindowNo))
				mTab.setValue("LineTotalAmt", lineNetAmount);
			else {

				mTab.setValue("LineTotalAmt", lineNetAmount.add(TaxAmt));
			}
```
With this correction the tax, line net amount and total amount was set correctly and furthermor the calculation of the tax in InvoiceTax and in the header was correct, see following screenshots.

**Case 1, Sales Invoice Tax Not included**
Before Save
01_SalesInvoice_TaxNotIncluded_BeforeSave

After Save
01_SalesInvoice_TaxNotIncluded_AfterSave

After Prepare
01_SalesInvoiceInvoiceTax_TaxNotIncluded_AfterPrepare

01_SalesInvoice_TaxNotIncluded_AfterPrepare

**Case 2, Sales Invoice Tax Included**

Before Save
02_SalesInvoice_TaxIncluded_BeforeSave

After Save
02_SalesInvoice_TaxIncluded_afterSave

After Prepare
02_SalesInvoice_TaxIncluded_AfterPrepare

02_SalesInvoiceInvoiceTax_TaxIncluded_AfterPrepare

Case 3, Purchase Invoice, Tax Not Included

After Save
03_PurchaseInvoice_TaxNotIncluded_afterSave

After Prepare
03_PurchaseInvoice_TaxNotIncluded_afterPrepare

03_PurchaseInvoiceINvoiceTax_TaxNotIncluded_afterPrepare

Case 4, Purchase Invoice, Tax Included

Before Save
04_PurchaseInvoice_TaxIncluded_beforesave

After Save
04_PurchaseInvoice_TaxIncluded_aftersave

After Prepare
04_PurchaseInvoiceInvoiceTax_TaxIncluded_afterPrepare

04_PurchaseInvoice_TaxIncluded_afterPrepare

nsarlabos:
In the case of purchase invoices created from an order (with the Create From button), the same problem is observed as in invoices, where the subtotal is equal to the total:

OC

Also, it is observed that if at the moment of creating the invoice header the price list with tax included is not selected (the order price list), then it happens that the tax is added again. This has also been fixed in this pull request, in the InvoiceCreateFrom class, to assign the list that has the order to the invoice.

DxP desde OC

To reproduce this error, the following steps must be followed:

- Create purchase order with list of taxes included
- Create supplier invoice header without choosing the order price list
- Create the invoice lines using the Create From button

nsarlabos:
As a summary, the errors found are the following:

**In invoices and orders created manually:**

- In the lines, the "Net of line" field is left with the tax and therefore the "Total of the line" field is also wrong, since the tax is added again.

linea factura

- In the header, the subtotal and total fields remain the same, both with the tax included.

cabezal factura

orden

**In vendor invoices created from order:**

In this case the same happens, leaving the subtotal equal to the total.
Also, it is observed that if at the moment of creating the invoice header the price list with tax included is not selected (the order price list), then it happens that the tax is added again. This has also been fixed in this pull request, in the InvoiceCreateFrom class, to assign the list that has the order to the invoice.

DxP desde OC

To reproduce this error, the following steps must be followed:

- Create purchase order with list of taxes included
- Create supplier invoice header without choosing the order price list
- Create the invoice lines using the Create From button

Regarding the modified classes in this pull request, at the time we consulted with @e-Evolution Víctor, and he just told us that it would be necessary to touch several classes in order to make these changes.

SusanneCalderon:
@nsarlabos 
We have tested your case. We used a price List including tax:
imagen

The tax is VAT 22%:
imagen

The purchase order generated correctly:
PO_TaxIncluded


We generated the invoice Line with _Create from Order._ Before using this method the Invoice must be saved and the price List is mandatory. Using this process the modified callout is not used.
PO_TaxIncluded_02

nsarlabos:
New issue and pull request will open

Issue #3518, opened 2021-07-12
Swing: Button Posted freezes the current window


SusanneCalderon:
Reference to Issue #3329 
#### Steps to reproduce

1. In Swing use the button Posted and select Repost 
2. The window is freezed and the aplication is not generated

Reason: Class _AEnv_ in _client/src/compiere/apps_   calls DocumentEngine with TrxName null
In DocumentEngine the method postImmediate requires TrxName without capturing an exception
So the window gets freezed

mckayERP:
Noted.

e-Evolution:
Hi @mckayERP Mike , it is strange because the tests have no problem integrating the refectory #3329.

So @SusanneCalderon , @mckayERP We should deprecate the method? 
I suppose any call to this code has the same problem:

DocumentEngine.postImmediate (Env.getCtx (), AD_Client_ID, AD_Table_ID, Record_ID, force, null); 

On the other hand, @mckayERP you should review the design testing to refactory #3329

kind regards
Victor

e-Evolution:
@SusanneCalderon  You  can check if this happen also in ZK UI please?

mckayERP:
Verified that there is an error when trxName is null;

DocumentEngine.postImmediate has a statement

`requireNonNull(trxName)`

Created a test to validate the error.  Both the Swing AEnv class and the ZK AEnv have a postImmediate method.

The Swing version uses `DocumentEngine.postImmediate()`.  This calls doc.postImmediate().
The ZK version calls `Doc.postImmediate()`. This also calls doc.postImmediate().

Created some tests to verify the behaviour:


    class IT_DocumentPosting extends CommonGWSetup {

    ....
    
    @Nested
    @DisplayName("Given an existing document "
            + "that has no errors")
    class GivenAnExistingDocumentWithNoErrors {
        
        private int invoiceId;
        
        @BeforeEach
        void getExistingDocument() {
            
            int windowNo = 0;
            invoiceId = 103;
            MInvoice existingInvoice = new MInvoice(ctx, invoiceId, null);
            Env.setContext(ctx, windowNo, "IsSOTrx", existingInvoice.isSOTrx());
            Env.setContext(ctx, windowNo, "IsApproved", existingInvoice.isApproved());
            Env.setContext(ctx, windowNo,  "C_Currency_ID", existingInvoice.getC_Currency_ID());
            
        }
        
        @Test
        final void whenRepostedFromSwing_postingShouldSucceed() {
            
[...]

PR #3520, opened 2021-07-12
Adempiere #3518 Swing: Button Posted freezes the current window

SusanneCalderon:
Using Swing the method postImmediate freezes the window due to a requirenonnull in DocumentEngine
Solution: 
Using in _client/org/compiere/apps/AEnv.java_ the same code as in _zkwebui\WEB-INF\src\org\adempiere\webui\apps\AEnv.java_
https://github.com/adempiere/adempiere/issues/3518


mckayERP:
Hi Susanne, 

I'd Rather fix the problem with the DocumentEngine class.  It shouldn't throw an exception for null trxName. 

Mike

SusanneCalderon:
@mckayERP 
Accepted.

PR #3524, opened 2021-07-13
Remove test that trxName be not null from DocumentEngine.postImmediate

### LINKED ISSUE #3518: Swing: Button Posted freezes the current window

mckayERP:
Fixes #3518.

A null trxName is a valid value.

SwingDocPosting

ZKDocPosting


SusanneCalderon:
@mckayERP , @e-Evolution , @marcalwestf 
I tested the solution with different doctypes.

PR #3525, opened 2021-07-13
Add chart support for PA Goal

yamelsenih:
Just add support to charts based on **AD_Chart** table

Is nice this change because can join the current measure with Adaxa
implementation for Charts

See the definition:
Screenshot_20210713_114708

Screenshot_20210713_114729
Screenshot_20210713_114753

Screenshot_20210713_114825

Screenshot_20210713_114853
Screenshot_20210713_114922

Screenshot_20210713_114631


marcalwestf (approved):
- Load and apply XML file
imagen
- created a Goal es described
- Goal was displayed as in the explanation

Test passed

PR #3526, opened 2021-07-13
fixed error with create from on current payment

yamelsenih:
When the Smart Browser for create payment from Invoices is nice for
generate payments but when is created over a current payment, all values
are overwrited from parameters

Screenshot_20210713_122333


marcalwestf (approved):
Good catch @yamelsenih .
The remaining amount is now taken from the existing _Payment_ instead of from the Process Parameter _payAmt_.
imagen

PR #3554, opened 2021-07-20
Refactory: EMail Dialog based on new notification queue

yamelsenih:
Add new queue notification

This apply for:
- `org.adempiere.webui.window.WEMailDialog`
- `org.compiere.apps.EMailDialog`

Related to: #3552


rcapecce (changes requested):
Hi @yamelsenih I can't test this funcionality because the from and to emails are worng when the system call this class

PR #3555, opened 2021-07-20
Refactory: Workflow Activity based on new notification queue

yamelsenih:
- Remove old implementation of send mail
- Add new queue notification

Related to: #3552

### DIFF
--- a/base/src/org/compiere/wf/MWFActivity.java
+++ b/base/src/org/compiere/wf/MWFActivity.java
@@ -28,6 +28,7 @@ import java.sql.Savepoint;
 import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.List;
 import java.util.Properties;
@@ -38,7 +39,6 @@ import org.adempiere.exceptions.AdempiereException;
 import org.compiere.model.I_AD_Process;
 import org.compiere.model.MAttachment;
 import org.compiere.model.MBPartner;
-import org.compiere.model.MClient;
 import org.compiere.model.MColumn;
 import org.compiere.model.MConversionRate;
 import org.compiere.model.MMailText;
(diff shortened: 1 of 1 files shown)


base/src/org/compiere/wf/MWFActivity.java
```diff
@@ -1213,30 +1212,23 @@ else if (!(doc.processIt (DocAction.ACTION_Approve)))
...
-							+ ": " + doc.getDocumentNo(), 
-							(doc.getSummary() != null ? doc.getSummary() + Env.NL : "" )
+				//	Get instance for notifier
+				DefaultNotifier notifier = (DefaultNotifier) QueueLoader.getInstance().getQueueManager(DefaultNotifier.QUEUETYPE_DefaultNotifier)
+						.withContext(Env.getCtx())
+						.withTransactionName(trx.getTrxName());
+				//	Send notification to queue
+				notifier
+					.clearMessage()
+					.withApplicationType(DefaultNotifier.DefaultNotificationType_UserDefined)
```
rcapecce, line 1222: Hi @yamelsenih, in this case, when you use the "withApplicationTYpe" with the option "DefaultNotificationType_UserDefined", you don't have the user stablished.

rcapecce (changes requested):
Hi @yamelsenih 
When you run this configuration with the DefaultNotificationType_UserDefined, you don't have the user in this line
https://github.com/erpcya/adempiere/blob/7caaa9692e8e858fce3c6582ae91b6228a31fed3/base/src/org/spin/queue/notification/DefaultNotifier.java#L313
Can you check this?

base/src/org/compiere/wf/MWFActivity.java
SusanneCalderon: Hi @yamelsenih 
When you do not specify an email address (Text in Workflow Node) you get an error.

e-Evolution:
@yamelsenih can please solve the issue report to @rcapecce and @SusanneCalderon , this way I can merge this change

base/src/org/compiere/wf/MWFActivity.java
yamelsenih: Hi @all the user defined just allows define directly from user notification type and a notification can be send by email, telegram and other media type depending of user defined and supported client

yamelsenih:
Hi @e-Evolution this was resolved yet

PR #3556, opened 2021-07-20
Fixed error after press ok on Bank Statement Matcher

yamelsenih:
## What is the bug?
The bank statement movements loaded previously should be imported after
press ok button on **Bank Statement Matcher** form when it is open from
**Bank Statement** window

## Step for Reproduce
- Create a new Bank Statement
- Load a Bank Statement Extract from process **Load Bank Statement**
- Open **Bank Statement Matcher** from **Bank Statement** window
- Search the No Matched imported payments
- Press Ok
- The result is nothing
- Note that is almost one payment is matched then all payments are
imported


marcalwestf (approved):
With these changes not matched import lines are also imported.
I had to use a matcher class in order to test it.

Test passed.

PR #3557, opened 2021-07-20
Enlarge length for Description on Order and Invoice

yamelsenih:
Increase size of columns:

- `C_Order.Description`
- `C_OrderLine.Description`
- `C_Invoice.Description`
- `C_InvoiceLine.Description`

This allows define a wide description for clients like services where is
used the sales order with specification of contracts


marcalwestf (approved):
- Load and Apply XML file
- Check column Description on both AD and DB
- Result: both increased to 16000 characters:
image
- Check on Sales Order:
image


Test passed

PR #3559, opened 2021-07-20
Refactory: Send Invoice Print receipt By Email based on new notification

yamelsenih:
queue
Add new queue notification

This apply for:
- `org.compiere.process.InvoicePrint`

Related to: #3552


rcapecce (approved):
Hi @yamelsenih , this work fine to me in both cases.
I test this with swing and with web debug.
In both cases, the notification reached the notification queue.
In other case, I report this problem refering to the queue itself (https://github.com/adempiere/adempiere/pull/3630) that has a context error, but refering to this issue, its ok

marcalwestf (approved):
Select Invoice to be sent.
Configure Invoice User to be Email recipient and set a valid email address.
Call Process _Print Invoices_:
imagen

Email was received correctly.

Test passed.

PR #3560, opened 2021-07-20
Refactory: Send Invoice Print receipt By Email based on new notification queue

yamelsenih:
Add new queue notification

This apply for:
- `org.compiere.process.DunningPrint`

Related to: #3552

### DIFF
--- a/base/src/org/compiere/process/DunningPrint.java
+++ b/base/src/org/compiere/process/DunningPrint.java
@@ -20,19 +20,18 @@ import java.io.File;
 import java.util.List;
 
 import org.compiere.model.MBPartner;
-import org.compiere.model.MClient;
 import org.compiere.model.MDunningLevel;
 import org.compiere.model.MDunningRun;
 import org.compiere.model.MDunningRunEntry;
 import org.compiere.model.MMailText;
 import org.compiere.model.MQuery;
 import org.compiere.model.MUser;
-import org.compiere.model.MUserMail;
 import org.compiere.model.PrintInfo;
 import org.compiere.print.MPrintFormat;
 import org.compiere.print.ReportEngine;
 import org.compiere.util.AdempiereUserError;
-import org.compiere.util.EMail;
+import org.spin.queue.notification.DefaultNotifier;
+import org.spin.queue.util.QueueLoader;
 
 /**
  *	Dunning Letter Print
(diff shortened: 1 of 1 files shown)


base/src/org/compiere/process/DunningPrint.java
```diff
@@ -155,46 +153,37 @@ else if (to.getEMail() == null || to.getEMail().length() == 0)
...
-				}
-				//
+				//	Get instance for notifier
+				DefaultNotifier notifier = (DefaultNotifier) QueueLoader.getInstance().getQueueManager(DefaultNotifier.QUEUETYPE_DefaultNotifier)
+						.withContext(getCtx())
+						.withTransactionName(get_TrxName());
+				//	Send notification to queue
+				notifier
+					.clearMessage()
+					.withApplicationType(DefaultNotifier.DefaultNotificationType_UserDefined)
```
rcapecce, line 170: Hi! @yamelsenih , with this ApplicationType the withUserID() method should be mandatory.
As I see in this code block
https://github.com/adempiere/adempiere/blob/develop/base/src/org/spin/queue/notification/DefaultNotifier.java#L311

rcapecce:
Hi @yamelsenih!
I tested this functionality and I see an error when the Notification is added to queue.
As I see, the exception is due to this line
https://github.com/adempiere/adempiere/blob/develop/base/src/org/spin/queue/notification/DefaultNotifier.java#L311
Can you check if this comment is correct?

base/src/org/compiere/process/DunningPrint.java
```diff
@@ -155,46 +153,37 @@ else if (to.getEMail() == null || to.getEMail().length() == 0)
...
+					.addRecipient(getAD_User_ID())
```
rcapecce, line 171: This maybe should be the AD_User_ID of the variable "to", not the AD_User_ID of the running process

yamelsenih:
> Hi @yamelsenih!
> I tested this functionality and I see an error when the Notification is added to queue.
> As I see, the exception is due to this line
> https://github.com/adempiere/adempiere/blob/develop/base/src/org/spin/queue/notification/DefaultNotifier.java#L311
> Can you check if this comment is correct?

Hi @rcapecce this was resolved here: #3568

base/src/org/compiere/process/DunningPrint.java
rcapecce: The user id was corrected, this is loading the correct user id, now I need to test this PR (https://github.com/adempiere/adempiere/pull/3568) for the exception launched to

base/src/org/compiere/process/DunningPrint.java
rcapecce: Hello @yamelsenih , yes, with this change in the DefaultNotifier, this doesn't happen now

rcapecce (approved):
Hi! @yamelsenih with the change in the DefaultNotifier.java, this is not required now.

PR #3561, opened 2021-07-20
Refactory: Send Mail based on new notification queue

yamelsenih:
Add new queue notification

This apply for:
- `org.compiere.process.SendMailText`

Related to: #3552

### DIFF
--- a/base/src/org/compiere/process/SendMailText.java
+++ b/base/src/org/compiere/process/SendMailText.java
@@ -27,10 +27,10 @@ import org.compiere.model.MInterestArea;
 import org.compiere.model.MMailText;
 import org.compiere.model.MStore;
 import org.compiere.model.MUser;
-import org.compiere.model.MUserMail;
 import org.compiere.util.DB;
-import org.compiere.util.EMail;
 import org.compiere.util.Msg;
+import org.spin.queue.notification.DefaultNotifier;
+import org.spin.queue.util.QueueLoader;
 
 /**
  *  Send Mail to Interest Area Subscribers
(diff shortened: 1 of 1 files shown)


base/src/org/compiere/process/SendMailText.java
```diff
@@ -254,61 +254,41 @@ else if (ok.booleanValue())
...
-		return new Boolean(OK);
+		//	Get instance for notifier
+		DefaultNotifier notifier = (DefaultNotifier) QueueLoader.getInstance().getQueueManager(DefaultNotifier.QUEUETYPE_DefaultNotifier)
+				.withContext(getCtx())
+				.withTransactionName(get_TrxName());
+		//	Send notification to queue
+		notifier
+			.clearMessage()
+			.withApplicationType(DefaultNotifier.DefaultNotificationType_UserDefined)
+			.withUserId(m_from.getAD_User_ID())
```
rcapecce, line 283: Whats is the difference between withUserId and addRecipient methods?
Because the DefaultNotifier.QUEUETYPE_DefaultNotifier requiere the UserID, but not the recipient ID for email.
In this case, when you call the method without selecting the "From User" parameter, this throw NPE here.
Should not be the user of the recipient instead of the "m_from" property?

rcapecce (changes requested):
Hi @yamelsenih, when I call the process without selecting the "From User" parameter, this throw a NPE.
I think the "From User" parameter should be mandatory or in the notifier you use the recipient user instead of the parameter user

PR #3562, opened 2021-07-20
Refactory: Web Enviroment based on new notification queue

yamelsenih:
Add new queue notification

This apply for:
- `org.compiere.uti.WebEnv`

Related to: #3552


rcapecce (commented):
Hi @yamelsenih , can you add the steps to reproduce and test this from an IDE?

PR #3566, opened 2021-07-20
Refactory: Asset Delivery By Email based on new notification queue

yamelsenih:
Remove old implementation of Payment Servlet Class
Add new queue notification

This apply for:
- `org.compiere.process.AssetDelivery`

Related to: #3552


SusanneCalderon (approved):
Tested. 

- The notification appeared in the notification queue.
- Email was sent to user

PR #3568, opened 2021-07-21
change wrong validation for user on notifier

yamelsenih:
The current validation verify that exists the user for notification if
the notification type is defined as "User Defined".

The problem is that this validation must be by user of recipient instead
user sender


rcapecce:
Hi @yamelsenih , I test this PR and works fine the change, related to #3560

PR #3592, opened 2021-08-16
Improve modification indication UX in Swing editor

dram:
Instead of indicating that data is modified only when focus is out of editor, this commit try to mark modification immediately after text in editor is changed.

With this change, user can type <F4> or click Save icon directly to save data after text editing is finished, without having to move focus out of editor.

BTW: I have only tested VString and VText, not sure where VMemo is used, may someone give me a hint?

Before:

before

After:

after

### DIFF
--- a/client/src/org/compiere/grid/ed/VCellEditor.java
+++ b/client/src/org/compiere/grid/ed/VCellEditor.java
@@ -203,9 +203,10 @@ public final class VCellEditor extends AbstractCellEditor
 	{
 		if (m_table == null)
 			return;
+
 		log.fine(e.getPropertyName() + "=" + e.getNewValue());
-		//
-		((GridTable)m_table.getModel()).setChanged(true);
+
+		m_table.setValueAt(e.getNewValue(), m_table.getEditingRow(), m_table.getEditingColumn());
 	}   //  vetoableChange
 
 	/**
--- a/client/src/org/compiere/grid/ed/VMemo.java
+++ b/client/src/org/compiere/grid/ed/VMemo.java
@@ -20,8 +20,6 @@ import java.awt.Component;
 import java.awt.Container;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
-import java.awt.event.FocusEvent;
-import java.awt.event.FocusListener;
 import java.awt.event.KeyEvent;
 import java.awt.event.KeyListener;
 import java.awt.event.MouseAdapter;
@@ -50,8 +48,7 @@ import org.compiere.util.Msg;
  *  @author 	Jorg Janke
  *  @version 	$Id: VMemo.java,v 1.2 2006/07/30 00:51:27 jjanke Exp $
  */
-public class VMemo extends CTextArea
-	implements VEditor, KeyListener, FocusListener, ActionListener
+public class VMemo extends CTextArea implements VEditor, KeyListener, ActionListener
 {
 	/**
 	 * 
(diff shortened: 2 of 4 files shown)


dram:
Hi @marcalwestf , I have added two gif screencasts, hope this helps.

marcalwestf:
@dram  thanks a lot.
Now it is very clear!
I think it will be soon tested.

marcalwestf:
@dram  in table _AD_Rule_, column _Script_, you find a memo:
image

marcalwestf:
1. Tested as described a grid containing text. 
    Fields: _Search Key_, _Name_, _Description_, _Comment/Help_.
    Work as described: changes in grid as soon as a key is hit.
2. Change field _Script_ on window _Rule_:
20210831_ChangeMemo

marcalwestf:
3. Change memo field on grid: changes are OK, but when you switch back to the single line view, it forgets the change.
    This happens also with the other types. I will post a gif this regarding.
20210831_ChangeMemo2

marcalwestf:
It seems that switching from single row to grid, the changes are kept, but not the other way around.
20210831_ChangeText

marcalwestf:
@dram  and also another favor please: the changes are done with a different encoding, so that when we compare the changes, we must select "Ignore white spaces" to see them well.
Perhaps you are so kind and make sure the encoding is such that a compare yields directly only changes? Thanks a lot!!!
20210831_IgnoreWhiteSpaces

marcalwestf (changes requested):
- tests went well as described above
- missing just taking changes from grid display to single-row display

dram:
@marcalwestf thank you for your comprehensive tests!

Regarding to the two problems you mentioned:

1. "missing just taking changes from grid display to single-row display", I think this problem is not related to this PR, as it already exists in current develop branch.
2. file encoding: I think this problem is caused by line endings of those files changed by this PR, which is ended with CRLF, it may cause problem when checking out in Linux systems, so this problem is also not related to this PR.

marcalwestf:
@dram  
1. "_I think this problem is not related to this PR, as it already exists in current develop branch._".   -> I know. I thought perhaps you'd take it up once you're there  :) 
2.  " _I think this problem is caused by line endings of those files changed by this PR, which is ended with CRLF, it may cause problem when checking out in Linux systems_"  -> Agree 100% with you. 
The solution would be that you keep the files you want to commit in the same way as you find them. In your case, take the files and convert them. As many of us do develop with iOS and Linux, it would be cumbersome if we'd change every time we would need to check changes.

Thanks in advance.

dram:
Hi @marcalwestf,

I did some more investigation, the change loss problem when toggling between single and multi rows is caused by those lines:

https://github.com/adempiere/adempiere/blob/develop/client/src/org/compiere/grid/GridController.java#L765-L770

If those lines are commented out, it works properly, so it is related to issue #421 , and maybe @mckayERP can have a look at that.

Regarding to the line ending problem, there are two approaches to cope with it:

1. Change all touched files to LF endings, and then do modification based on them.
2. Do not change file endings, in this case, as those files in develop branch are CRLF endings, keep it as is.

Currently, I take the second approach, it should not cause problem with proper git client configs. The problem you mentioned may be caused by that your git client converting all CRLF endings to LF automatically when checking out.

BTW, following are some quick statistics:

```
% find . -name *.java | wc -l
5328
% find . -name *.java -exec file {} \; | grep CRLF | wc -l
1993
```

There are 5k Java sources files, and with almost 2k of them are CRLF endings. With this large number of CRLF ending files, personally, I'd prefer second approach to first one.

marcalwestf:
@dram  thanks for your elaborate answer.
It seems that actually, there was an automatic change on one of my tools, because when I download the files directly from Github, and compare them with the head revision, only the differences are displayed. 

As soon as the checks are passed, I will merge this PR.

If you wish, and it is not burdensome to you, I can create an issue regarding the switching between single-row and grid and you might create a pull request? I would test it and eventually accept it.
I cannot do the pull request and accept it myself....
Thanks !

marcalwestf (approved):

dram:
Hi @marcalwestf , feel free to create an issue, I'll have a look at that problem.

PR #3593, opened 2021-08-19
Add new currency for Venezuela VED

yamelsenih:
From 2021-10-01 the currency VES of Venezuela will be changed to VED
(Temp)
Reference:
http://www.bcv.org.ve/comunicado-bolivar-digital

[comunicado_bolivardigital.pdf](https://github.com/adempiere/adempiere/files/7014834/comunicado_bolivardigital.pdf)

I will to add a new currency for keep the old transactions with VES and
new transaction with VED, the VED currency will be change to VES after
conversion and the old VES currency will be change to VEO

The conversion rate will be 1 VED ≃ 1000000 VES


marcalwestf (approved):
- Load XML
image

- Apply XML
image

- Display Currency 
image

Test passed

PR #3595, opened 2021-08-23
Minor change for validation SO

yamelsenih:
Currently is impossible change a warehouse for a sales order line if it
is invoiced

The problem is that the warehouse just change the reserved quantity and
delivery, the invoice is not affected for it.

### DIFF
--- a/base/src/org/compiere/model/MOrderLine.java
+++ b/base/src/org/compiere/model/MOrderLine.java
@@ -578,11 +578,6 @@ public class MOrderLine extends X_C_OrderLine implements IDocumentLine
 			log.saveError("Error", Msg.translate(getCtx(), "QtyDelivered") + "=" + getQtyDelivered());
 			return false;
 		}
-		if (getQtyInvoiced().signum() != 0)
-		{
-			log.saveError("Error", Msg.translate(getCtx(), "QtyInvoiced") + "=" + getQtyInvoiced());
-			return false;
-		}
 		if (getQtyReserved().signum() != 0)
 		{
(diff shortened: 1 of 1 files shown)


marcalwestf:
@yamelsenih  you're right.
I am going to test it.

yamelsenih:
Thanks

marcalwestf:
@yamelsenih 
When completing a Sales Order, there is a reservation on the Sales Order line:
Selection_012

And in the code, you check if there is a quantity reserved:
image

So even if you take the check if there is an invoice, the warehouse will not be changed, because of the quantity reserved.
I wonder if there is a situation when the Sales Order is completed, and there is no reservation...

Which is the use case are you actually testing?

yamelsenih:
Hi Mario
The reserve quantity is ok and is keeping. But the invoice not have effect over stock or reservation.

My test:

- Add line with 5 units for HQ warehouse
- complete order
- check reservation: 5 units
- re-activate the order
- change quantity to 0 units
- prepare order
- check the reservation: 0 units
- now change the warehouse and set quantity to 5 units
- complete order
- check reservation: now exists 5 units for fertilizer

The validation is veri useful when a line was delivered but a invoice never make changes on stock o reservation

Best regards

marcalwestf:
very good catch!!!! Of course, this is a legit scenario.

marcalwestf (approved):
- Did the steps as described by @yamelsenih 
- The warehouse can with the modification be changed and saved under these circumstances
- This is a better solution than avoiding etc.
Test passed.

PR #3596, opened 2021-08-24
Add payment method reference for payment

yamelsenih:
Currently only exist Tender Type for payment, this pull request allows
add a payment method reference to payment

Screenshot_20210823_231216


marcalwestf:
@yamelsenih  

- loaded and applied the XML.
- Created some _Store Payment Methods_
- Opened the _Payment_ window
- Created a new record and select a  _Store Payment Method_
- Save
- -> OK

From the technical point of view, this works well.
Selection_014


I'd like to know which is the semantic reason for this.
Can you describe/depict a use case/scenario where a  _Store Payment Method_ is useful?
Seems to be more informative, isn't it?

It also seems that there should be (later) a validation that  the _Tender Type_ of both _Payment_ and _Store Payment Method_  match.

yamelsenih:
Hi @marcalwestf jus a new field that can be used from web store or POS for define many payment method usinf a single Tender Type, current now only exist tender type for payment and this change allows have a additional attribute for save in the payment

marcalwestf (approved):
Test passed as detailed in earlier posts.

PR #3602, opened 2021-08-29
Add Waiting Time for Queue Type

yamelsenih:
Just add a waiting time for queue type, this allows define a waiting
time and duration unit for process queue items.
Now a queue item can be processed when it was expected a waiting time afgter the last process
Screenshot_20210829_002753

### DIFF
--- a/base/src/org/spin/queue/model/I_AD_QueueType.java
+++ b/base/src/org/spin/queue/model/I_AD_QueueType.java
@@ -166,6 +166,19 @@ public interface I_AD_QueueType
 	  */
 	public String getQueueType();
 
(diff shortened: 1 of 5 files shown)


marcalwestf (approved):
- Load + apply  XML
- Set in window _Queue Type_ a waiting time of a couple of minutes (Role: _System Administrator_).
- Get a record created in table _AD_Queue_ (example: Request without configuring email)
- Call process _Flush System Queue_
- The queue entry is flashed (or not) accordingly to the parameters set.
- The new methods in code are executed correctly and only when the time allows.

Test passed.

PR #3603, opened 2021-08-30
Fixed process selection when summary row is the first row

carlosaparadam:
## Problem
On Smart Browse when the summary row is the first row and selected another record, the selected records don't go to process.

## Steps for Reproduce
- Open an Bank Statement
- Create From Payment
- Sort By Document Number 
- When the summary row be the first row, select a row

Bad_Selection_SB

## After Fixed

Fixed_Selection_SB


marcalwestf (approved):
- do the steps as described
- when selecting the first record after ordering the found results, the first record is not left out.
- tested other record combinations and they are also accepted.
Test passed.

PR #3605, opened 2021-09-02
Fix TimeUtilTest in timezone with positive offset

dram:
`TimeUtil.getDay` will get result in locale time zone, so increase test case of `1970-01-01` by one day, to bypass the problem in timezone with positive offset.

To test this, please run following command before and after change:

```
TZ=Asia/Shanghai sh tools/lib/ant/apache-ant-1.10.10/bin/ant
```


e-Evolution:
@mckayERP, Hi Mike Would you review this change, please?

mckayERP (approved):

PR #3607, opened 2021-09-03
Add Included and Excluded Payments for bank statement

yamelsenih:
Currently not exists a way for know if a payment already exists inside
matched payments when the form for **Match Payments** is used, then each
matcher go to find any payments without know if a payment already exists
as matched with a previous matcher.

## How test it?
- Create three payments with the same amount and currency
- Load a Bank Statement: Enforce it for load four lines, three with the
same amount that the previous payments created and one distinct
- Go to **Bank Statement Match** form
- Simulate a match
- Note that only one payment is matched

## What is the problem?
Each matcher make a search to db with a criteria and the first run get a
payment. After this run, the second matcher find the same payment that
the previous and is ignored.

### DIFF
--- a/base/src/org/compiere/impexp/BankStatementMatcherInterface.java
+++ b/base/src/org/compiere/impexp/BankStatementMatcherInterface.java
@@ -16,6 +16,8 @@
  *****************************************************************************/
 package org.compiere.impexp;
 
+import java.util.List;
+
 import org.compiere.model.MBankStatementLine;
 import org.compiere.model.X_I_BankStatement;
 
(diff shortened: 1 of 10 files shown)


italonc:
Code review was performed. Functional tests will begin.

italonc:
the findmatch method now allows to send 2 payment lists to add or exclude
the necessary payments.

SusanneCalderon (approved):
I tested in a working database with a bank statement of 383 lines

- 3 payments with the same amount
- 2 lines in the bank statement with the same amount

Steps
- Import of the bank statement to table i_Bankstatement
- Simulate the match. I used class _CurrencyAndAmount_ to simulate according to payment amount

Result
All lines are matched correctly. The two lines with the sames amount are matched to two different payments, according to the statementdate. 
imagen

PR #3608, opened 2021-09-03
Retain selected row when sorting grid table

### LINKED ISSUE #421: Embedded tab is not updated

### LINKED ISSUE #3606: Toggling between single and multi row causes content to be lost
#### Steps to reproduce
1. Open any window with text content line Search Value, Name, etc.
2. Display it in grid mode (=multi-row)
3. Edit the text context, without saving
4. Change to single-row display
5. Contents are lost, and also the save button becomes deactivated

#### Screenshot or Gif
20210831_ChangeText

#### Expected behavior
The contents should be also displayed when switching form multi-row to single-row display, similar to what happens when switching from single-row to multi-row.

#### Additional context
Solution explained by @dram  in PR https://github.com/adempiere/adempiere/pull/3592

dram:
Fixes #3606
This commit choose another way to fix #421, i.e. instead of updating embedded tabs after sorting, this commit tries to retain selected row after sorting, so no update is needed.

The implementation is ported from iDempiere, with small tweaks.

It works both in Swing and ZK UI.

Code added by #2088 and #2238 have been reverted, as it will cause problem as described in #3606.

### DIFF
--- a/base/src/org/compiere/model/GridTab.java
+++ b/base/src/org/compiere/model/GridTab.java
@@ -2079,10 +2079,16 @@ public class GridTab implements DataStatusListener, Evaluatee, Serializable
 		log.fine("#" + m_vo.TabNo + " - " + e.toString());
 		int oldCurrentRow = e.getCurrentRow();
 		m_DataStatusEvent = e;          //  save it
-		//  when sorted set current row to 0
+		//  when sorted set current row
 		String msg = m_DataStatusEvent.getAD_Message();
 		if (msg != null && msg.equals("Sorted"))
-			setCurrentRow(0, true);
+		{
+			oldCurrentRow = m_currentRow;
+			if (e.getCurrentRow() >= 0)
+				setCurrentRow(e.getCurrentRow());
+			else
+				setCurrentRow(0, true);
+		}
 		//  set current row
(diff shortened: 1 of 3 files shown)


dram:
After looking at relevant code in iDempiere, I think there is better way to solve this problem, so mark this PR as WIP for now, will update it later.

dram:
Code and description have been updated, now this PR is ready to be reviewed, sorry for the inconvenience.

mckayERP (approved):

marcalwestf:
@dram  I took the insolence of editing your post, just to indicate that this is the way to refer to an issue.
When doing this, the issue is closed when the pull request is merged or closed as well.

Please do excuse me taking this liberty and hope on your understanding.

dram:
@marcalwestf, please feel free to do any editing if needed, I will mention issues in that way next time, thanks!

PR #3585, opened 2021-07-29
[Feature Request] Change Connection pool library c3p0 to HikariCP

e-Evolution:
fixed #3582

PR #3609, opened 2021-09-03
Remove c3p0 connection leak workaround

dram:
As mentioned in comments of #3585, c3p0 connection leak workaround is
not needed anymore, as we have changed to use HikariCP instead.

I have only tested in PostgreSQL, may someone who has MariaDB, MySQL and Oracle environments can help to have a check.


e-Evolution:
Hi @dram Xin Wang, Thank you for the bugfix, when I implemented the change to HikariCP I saw this code and I  hesitated to leave it or remove it, but it seems that in HikariCP it is not necessary to validate the maximum number of connections

PR #3610, opened 2021-09-03
Remove duplicated slf4j-api package

dram:
New version of slf4j-api have been added in #3585, so old version can be
removed.


e-Evolution:
@dram Xin, Thank you to remove the duplicate library

PR #3614, opened 2021-09-07
Fix broken input method in single-row mode in Swing UI

### LINKED ISSUE #3613: [Bug Report] Text editor is broken in single-row mode in Swing UI when using on-the-spot input method
## Bug report

Text editor is broken in single-row mode in Swing UI when using on-the-spot input method.

#### Steps to reproduce

1. Open window Project Management > Project
2. In single-row mode, type text into "Description" field, using on-the-spot input method

#### Screenshot or Gif

current

#### Expected behavior

The expected behavior is shown as another screencast:

expected

- Your OS: Win 10
- Java version: Java 11
- ADempiere version: develop branch

dram:
In single-row mode, text input widgets and models behind them have
kind of two-way data bindings, which are established in GridController.

In order to mark data as modified as soon as possible. we currently
update model data more eagerly, as committed in #3592, which will cause
problem, i.e.:

1. user type characters in text widget
2. model got updated because of vetoable change
3. text widget got updated again because of property change

The third step not only adds an unneeded update operation, but also
cause problem if interfered with on-the-spot input methods, which will
make uncommitted characters to be saved in the text fields.

This commit tries to fix this problem by checking that text widgets
only set values triggered by property change in multi-row mode (i.e.
when those text input widgets are not shown).

Fixes #3613

### DIFF
--- a/client/src/org/compiere/grid/ed/VMemo.java
+++ b/client/src/org/compiere/grid/ed/VMemo.java
@@ -169,7 +169,7 @@ public class VMemo extends CTextArea implements VEditor, KeyListener, ActionList
 	 */
 	public void propertyChange (PropertyChangeEvent evt)
 	{
-		if (evt.getPropertyName().equals(org.compiere.model.GridField.PROPERTY))
+		if (evt.getPropertyName().equals(org.compiere.model.GridField.PROPERTY) && !getText().equals(evt.getNewValue()))
 			setValue(evt.getNewValue());
 	}   //  propertyChange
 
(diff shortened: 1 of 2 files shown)


marcalwestf:
@dram 
Tested by changing the field _Description_ in the  _Project_ window.
Problem is that the contents are saved when the _ESC_ key or the _Undo Changes_ icon are pressed, not only when the _Save_ button is pressed.

Perhaps they are side-effects?

Please, do see gif:
20210910_Change_Description_without_save

dram:
Hi @marcalwestf , thank you for spot this problem.

I have reimplement this PR in another way, hope it will fix this issue more properly. Please test again, thanks!

marcalwestf (approved):
- On Swing UI
- Edit text field in single and multi-mode
- Switching between single and multi-mode, pressing ESC, saving record in different combinations
- The content is always as expected: no data is lost

Test passed

PR #3617, opened 2021-09-08
Disable default keymaps for F6 and F8 in JSplitPane

dram:
# Description

Disable default keymaps for F6 and F8 in JSplitPane inside GridController, which conflict with shortcuts of "Lookup Record" and "Grid Toggle" in menu and toolbar.

# Test

Steps to reproduce the problem:

1. Login Swing client as GardenUser
2. Open Menu > Project Management > Project
3. Type F6 or F8 in keyboard

Before applying this PR, F6 and F8 will have no effect.

This is port from iDempiere: https://idempiere.atlassian.net/browse/IDEMPIERE-4934


marcalwestf (approved):
- Started on Swing
- Do as explained
- F8 => toggle between single and multi-row
- F6 => Lookup Record dialog

Test passed

PR #3619, opened 2021-09-08
Improve sidebar UI for Compiere theme in Swing

dram:
# Description

1. Make Compiere button opaque by default, to be consistent with Adempiere theme

2. Only draw 3D effect if Compiere button is opaque

# Screenshots

Before:

9beb4a1206ff24dea63b7ef9c944d2d

After:

445636ff80c93c743124260445aef26

This PR is port from iDempiere: https://idempiere.atlassian.net/browse/IDEMPIERE-4938


marcalwestf (approved):
- Do as described
- Tests before and after as expected
- Test passed

PR #3620, opened 2021-09-09
fixed error with getLines(String, String) method

yamelsenih:
When is called the method getLines(whereClause, orderBy), exists a error because the custom where clause never is  send to query. Instead is used the internal where clause, see the error line:

```Java
if (!Util.isEmpty(where, true))
			whereClause.append(whereClause);
```


marcalwestf (approved):
- Checked code
- Comments are OK: the _where clause_ as it was produced an error, because it added wrong syntax
- With this PR, the method's _where_ is used (its syntax must be OK)
- test passed

PR #3621, opened 2021-09-12
Reduce heap usage of Compiere theme

dram:
# Description

After merging #3619, heap usage is increased in Compiere theme, which is caused by flat backgrounds drawing using BufferedImage, which use quite a lot of memory.

This commit trying to draw flat background directly without BufferedImage.

# Test

Without this PR, Project window will be failed to open in Compiere theme mode (with default `-Xmx256M` option), i.e.:

1. Login ADempiere as GardenUser using Swing client
2. Open Project Management > Project

With step 2, OutOfMemoryError will be occured.

### DIFF
--- a/base/src/org/compiere/plaf/CompiereUtils.java
+++ b/base/src/org/compiere/plaf/CompiereUtils.java
@@ -77,7 +77,7 @@ public class CompiereUtils
 			cc = CompiereColor.getDefaultBackground();
 
 		//  Paint AdempiereColor
-		if (cc != null)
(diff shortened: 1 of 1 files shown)


marcalwestf (approved):
- **Before this change**:
  - Change to _Compiere_ theme
  - Open any window
  - A Java memory _heap space error_ is produced:
image

- **After this change**:
  - Change to Compiere theme
  - Open a window
  - No Java memory _heap space error_ is produced

**Test passed**.

PR #3630, opened 2021-09-23
Wrong cached context on EMailSender

### LINKED ISSUE #3629: [Bug Report] Wrong cached context on EMailSender
## Bug report

When you (or the Scheduler) run the Flush System Queue process, the EmailSender class get a cached Client with an incorrect context.
This siguation appears in an installed ADempiere instance, in this case, when the server start, load and cache the GardenWorld client, but with a System context (inside has the context variable "#AD_Client_ID" = System), then, later, when you run the "Flush System Queue" in GardenWorld, the EmailSender class get this catched Client (GardenWorld), internally the Email class sets the Client context (System) to itself, but this class needs the GardenWorld Client to get the correct Email Server Configuration (get the System Email Configuration and this is not setted).

This doesn't happend on debug because there is no cache

rcapecce:
https://github.com/adempiere/adempiere/issues/3629
Fixed #3629

### DIFF
--- a/base/src/org/compiere/util/EMail.java
+++ b/base/src/org/compiere/util/EMail.java
@@ -113,6 +113,7 @@ public final class EMail implements Serializable
 	 */
 	public EMail (MClient client, int eMailConfigId, String from, String to, 
 		String subject, String message, boolean html) {
+		this.client = client;
 		ctx = client.getCtx();
 		isManual = false;
 		//	FR [ 402 ]
@@ -250,6 +251,8 @@ public final class EMail implements Serializable
 	private String				token = "";
 	/** Context - may be null		*/
 	private Properties			ctx;
+	/** Client						*/
+	private MClient				client;
 	/**	Is Manual					*/
 	private boolean				isManual = false;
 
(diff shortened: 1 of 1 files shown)


base/src/org/compiere/util/EMail.java
```diff
@@ -625,7 +628,6 @@ private void setEMailParameters(int eMailConfigId) {
 		}
 		//	Validate Config
 		if(eMailConfigId == 0) {
-			MClient client = MClient.get(ctx);
```
rcapecce, line 628: Here get the Client from the wrong context (System).
Now the Email Server Configuration is getted from the correct Client

base/src/org/compiere/util/EMail.java
```diff
@@ -113,6 +113,7 @@ public EMail (MClient client, String from, String to,
 	 */
 	public EMail (MClient client, int eMailConfigId, String from, String to, 
 		String subject, String message, boolean html) {
+		this.client = client;
```
rcapecce, line 116: I save the Client to later, get the Email Server Configuration

PR #3642, opened 2021-10-05
Fixed error with receipt based on major quantity of purchase order

yamelsenih:
Currently exists a problem when a express receipt is created with a quantity major of purchase order, this validation already exist for **Match PO**


e-Evolution:
The logic was removed because exist this validation from MatchPO https://github.com/adempiere/adempiere/blob/develop/base/src/org/compiere/model/MMatchPO.java#L589

PR #3644, opened 2021-10-05
Fixed error with UOM for OutBound Line

yamelsenih:
Currently when a outbound order is generated the uom is not used correctly, also the movement quantity is not the real movement based on available for shipment. This pull request just allows define a correct uom for outboundline based on product.

**Features**
- Add correct UOM for OutBound Order Line
- Add correct movement quantity
- Add support to attribute set instance from Generate Outbound Order form
- Add support to default locator from Generate Outbound Order form


e-Evolution:
Adding in the form the Outbound Location , this way the user can select where the material is pick 

image

Now can select the location , if you not set then the default logic is used.

image

PR #3661, opened 2021-10-20
Error when reversing an assignment already reversed or canceled

### LINKED ISSUE #3660: [Bug Report] Error when reversing an assignment already reversed or canceled
## Bug report

When a payment is reversed, the allocations for that payment are obtained and reversed. But it has been seen that the method that performs this action obtains assignments in a reversed or canceled state, which causes an error.

#### Steps to reproduce
1. Create a payment and assign it to an invoice
2. Cancel or revert the previously generated assignment
3. Reverse the payment

#### Screenshot or Gif

Error

#### Expected behavior
The method "private void deAllocate (boolean isAccrual)" in the MPayment class, should control that the obtained assignments are in a state other than VOID "and" REVERSED, but currently it asks for a state other than VOID "or" REVERSED.

nsarlabos:
Fixes #3660

### DIFF
--- a/base/src/org/compiere/model/MPayment.java
+++ b/base/src/org/compiere/model/MPayment.java
@@ -2080,8 +2080,7 @@ public final class MPayment extends X_C_Payment
 		List<MAllocationHdr> allocations = Arrays.asList(MAllocationHdr.getOfPayment(getCtx(), getC_Payment_ID(), get_TrxName()));
 		log.fine("#" + allocations.size());
 		allocations.stream()
-				.filter(allocationHdr -> !allocationHdr.getDocStatus().equals(DOCSTATUS_Reversed)
-						|| !allocationHdr.getDocStatus().equals(DOCSTATUS_Voided))
+				.filter(allocationHdr -> allocationHdr.getDocStatus().equals(DOCSTATUS_Completed))
 				.forEach(allocationHdr -> {
 					allocationHdr.set_TrxName(get_TrxName());
 					if (isAccrual) {


marcalwestf:
hi @nsarlabos  Nicolás,
I felt free to test your suggestion with SQLs on an operating DB.

First, the present situation:
20211103_Payment_reverse_02


Then, your suggestion:
20211103_Payment_reverse_01

Eventually, the code in _**MAllocationHdr.reverseIt()**_:
20211103_Payment_reverse_03

With this in mind, your suggestion gets rid of documents with the status "VO" and "RE" (which would be caught at  _**MAllocationHdr.reverseIt()**_), but allows still documents with the status "DR" and "IP" to be reversed, which is wrong. Even in _**MAllocationHdr.reverseIt()**_ this would be allowed.

The correct change would be to allow only documents with the status "CO" for a reverse correction. What do you think?

By the way, it is always better and less confusing to define a SQL condition in a positive way than in a negative way, i.e. instead of "not A and not B", better "C and D".

nsarlabos:
Hi @marcalwestf Mario, I totally agree with you, if you think I can make a new pull request to consider only the documents in the "Complete" state.

marcalwestf:
@nsarlabos  Nicolas, or perhaps once we're here, just change the condition of the code and we're done.

marcalwestf (approved):
The use case to reproduce this error on seed 393 is the following (perhaps there are more):
1.- Create and complete a _Vendor Invoice_ 
2- Create an  _AP Payment_
3.- Assign the _Vendor Invoice_ to the _AP Payment_
4.- Complete the _AP Payment_
    (among other things, an Allocation for the Payment-Invoice is created and completed)
5.  **Close** (!!!!) **the Allocation** created and completed at Payment complete.
  (this is the cause of the error; also the Allocation cannot be undone anymore)
6.- Void or Revert the AP Payment

When this is done, an Exception is thrown and a message is displayed (perhaps there are more):
imagen
This is because without the changes proposed Allocations with status=CO, RE, VO, DR, IN are processed, and for example, a closed allocation throws the above exception.
The Payment remains in Complete status **AND** the Invoice is not assignable anymore.

With the changes by @nsarlabos , only allocation headers with status completed are processed, thus the exception is not thrown anymore and the Payment can be voided or reversed. If the Payment has several allocations, only allocations with status==CO will be processed which is OK. If by mistake one allocation was voided, this allocation will not be processed and not cause the whole process to fail.

The invoice remains unassignable because of a closed allocation which includes it, but this is another problem, which stems from the fact that the user closed the Allocation. This is a problem that has to be solved somewhere else.

Issue #3690, opened 2021-11-17
[Feature Request] New functionality to enable the process of a document through workflow 


e-Evolution:
## Feature request
New functionality to enable the process of a document through workflow 

Currently, when a delivery or invoice is generated manually, they are processed using the document engine through the process this method, the drawback of this approach does not execute the workflow that may have additional authorizations, notifications, or steps configured that are not considered.

#### Describe the solution you'd like
The new functionality enables document processing through the workflow allowing to process of the steps configured within the workflow. 

#### Describe alternatives you've considered
Process documents from the document window using the document action button

marcalwestf:
We have encountered the same issues.
A solution to it would be important.

colabware-app:
Hey @e-Evolution, I’m sure you are struggling to keep up with the number of feature requests here. [With Colabware](https://colabware.co/project/1626013066528x395405457477893570), you can harness the power of crowdfunding to drive innovation and sustain developer livelihood. Please do let us know if you have any questions, and I hope we can assist with your project!

PR #3691, opened 2021-11-17
Enable new functionality to processing document using workflow

e-Evolution:
…document through workflow  https://github.com/adempiere/adempiere/issues/3690

- Enable new functionality to processing document using workflow


SusanneCalderon (changes requested):
I tested both Processes, _InoutGenerate_ without an additional node, _InvoiceGenerate_ with an additional node, which updates the field _POReference_ with text _invoice generated_.

imagen


Case One, Complete new Invoice manually

- Generate Sales Order, Standard Order
- Generate Shipment with Special Form, docAction Complete
- Generate Invoice with Process Generate Invoice, docAction Prepare


- Complete new Invoice
**Result: Field POReference was set**
imagen

Case Two, Process _InvoiceGenerate_ with DocAction _Complete_
Results

- The Invoice was generated, but in Status Drafted
imagen

- Complete Manually gives the following message
imagen

**Console**
The workflow stops in the node _Complete_ and does not start the last node _SetPOReference_
as it works when you complete manually.

e-Evolution:
@SusanneCalderon thank a lot , I will adding in my test with a new node with your example, in my case, I use a new node to send a email , Would you testing this too, please 

kind regards

e-Evolution:
@SusanneCalderon can please set the image with your node setup to SetPOReference?

SusanneCalderon:
@e-Evolution 
This kind of node we use in several configurations.
imagen

SusanneCalderon:
@e-Evolution 
Debugging, it is interesting that the workflow does not get until this node. It stops at the end of activity of the node DocComplete. 
When you start the workflow from VDocAction the node is executed correctly.

SusanneCalderon:
@e-Evolution 
I tried a sendEmail and it worked. I then again tried a SetVariable Node, this time generated from client GardenWorld and the same error happened.

e-Evolution:
@SusanneCalderon Thank you now I got all the elements to replicate the behavior

yamelsenih:
Hi @e-Evolution is nice it. I have some comments:

- I just saw a implementation for Shipment generation but I think that if you will be add a functionality for core is nice if it can be applied to core documents. Exists many documents like:
  - Complete POS orders generate invoice and delivery
  - Complete Warehouse order generate Shipment
  - Complete Invoice order generate invoices

I think that this solution is not complete solution.

A way for resolve it can be the follow:
- Solution fast
  - Change all core implementations of DocAction
  - Add support on processIt method to new functionality
 
- Real solution is convert the DocAction interface to abstract class and move all functionality to parent

I think that is better make a discussion about it.

Best regards.

SusanneCalderon:
@yamelsenih 
I agree. It is dangerous not to synchronize similar cases. For example methods like _GenerateInvoiceFromLoan_ do not use the workflow. As user I would assume the same behavior.

e-Evolution:
Close this pull request to try other approach

e-Evolution:
Close this pull request to try other approach

Issue #3692, opened 2021-11-19
[Bug Report] Remove JVM parameter no compatible with JDK 17


e-Evolution:
## Bug report
The develop branch soon 3.9.4 was enable to execute using JDK 17 , but the exist parameters that was removed or no are use more in JDK 17

#### Steps to reproduce
1.- Try RUN_setup using JDK17

#### Expected behavior
The RUN_setup.sh should work without problems

#### Other relevant information
- Your OS: MacOS 12.0.1
- Java version: 17
- ADempiere version: 3.9.4
- DB Provider: PostgreSQL
- DB version:  14.1

PR #3693, opened 2021-11-19
#3692 [Bug Report] Remove JVM parameter no compatible with JDK 17 htt…

e-Evolution:
…ps://github.com/adempiere/adempiere/issues/3692


marcalwestf (approved):
- build 
- install with java 17
- Run Adempiere
- Result: no errors

Test passed.

Issue #3684, opened 2021-11-10
[Bug Report]


AlbertoAbudinen:
## Bug report
If there is more than one person working at the same time on the same server, even if they are different users, a message begins to appear that does not allow the work to continue.

See image below

#### Steps to reproduce

1.- two  users must connect to the system at the same time
2.- they must begin to carry out transactions, example create orders
3.- after a few minutes the error message will appear

#### Screenshot or Gif
![image](https://user-images.githubusercontent.com/31511336/141103339-27fc5f83-da47-493c-b565-41022e68c740.png)

#### Other relevant information
- Your OS: windows 10
- Java version: 11 
- ADempiere version: 394
- Browser [e.g. chrome, safari] chrome
- Browser Version [e.g. 22]
- DB Provider: postgresql
- DB version: 12

piracio:
Hi Alberto, is possible to clarify if these two users are using the same sessions or they are using different sessions?  I will explain what session means
Opening two TABS in the same browser will use the same session.
Opening one TAB on normal and another one using PRIVATE Browsing will open a new session.
Using two different browsers will open two sessions. 
Two users are using the same account? or they are two different users?

AlbertoAbudinen:
Hello, the test was carried out on 2 different computers, each one
connected with different users, with different IPs and opening different
windows of the system at the same time.

All windows show the error when there is more than 1 user, we test it with
2 and 3 different users.

colabware-app:
Hey @AlbertoAbudinen, it’s hard finding open-source contributors to build your features for free. By [submitting your request through Colabware](https://colabware.co/project/1626013066528x395405457477893570), harness the power of crowdfunding to encourage developers, whilst enabling others to support your vision. Please do let us know if you have any questions, and I hope we can assist with your project!

AlbertoAbudinen:
Hello, in tests carried out today, the problem of the message persists, and also as you can see in the photos, the sessions are changed, it says that Felipe Hube is connected when the user who connected is Ramiro Vergara.

![image](https://user-images.githubusercontent.com/31511336/144297968-5d820071-d93a-4eb1-9800-fe4b48212555.png)


![image](https://user-images.githubusercontent.com/31511336/144297987-4d1af3ce-f6c8-4d04-b7af-0de3b8e74513.png)

PR #3694, opened 2021-11-22
#3684 [Bug Report] Incorrect management of sessions with same user in…

e-Evolution:
… multiples instances and different web browser https://github.com/adempiere/adempiere/issues/3684

- Fixed the context based on AEnv.getDesktop that was resister in context


marcalwestf:
@e-Evolution  starting the revision

marcalwestf (approved):
- build this branch
- install binary from build
- use different browsers (Firefox, Chrome)
- run Adempiere on different users and different clients
- run Adempiere on same user and same client
- no proplems

Test passed

PR #3698, opened 2021-12-10
Update C_REGION.xml

audetpascale:
Set name to abbreviation, keep en / fr names.
https://www.iso.org/obp/ui/#iso:code:3166:CA


marcalwestf (approved):
- Checked link delivered for Canada's province names.
- Checked differences on IDE: only full names instead of abbreviations are used.
- No build needed

Test passed.

Issue #3704, opened 2022-01-24
[Bug Report] Possible leak resource into the process Import Order


e-Evolution:
## Bug report
Possible leak resource transaction into the process Import Order

#### Steps to reproduce
1.- The close result set and prepare statement should be into the finally clause, the reason is that if any thing fail then the resource never will be close 

#### Expected behavior
The close result set and prepare statement should be into the finally clause

#### Other relevant information
- Your OS: MacOS
- Java version: jdk11
- ADempiere version: 394 to Release
- Browse safari
- DB Provider: PG
- DB version: 14

PR #3705, opened 2022-01-24
[Bug Report] Possible leak resource into the process Import Order

e-Evolution:
…der https://github.com/adempiere/adempiere/issues/3704


marcalwestf:
@e-Evolution  now starting to test it

marcalwestf:
@e-Evolution  it seems that with the changes, only **one** Purchase Order / Sales Order record is created. The following records are rejected with the error message: "_ERROR: duplicate key value violates unique constraint "c_order_documentno"  Detail: Key (documentno, c_doctype_id, c_bpartner_id)=(PO-20222-44433, 0, 1000019) already exists._"

See image where you can see that only one record has been created:
image

Issue #3703, opened 2022-01-24
[Bug Report] posible leak transaction when Account Processor can not get some document


e-Evolution:
## Bug report
[Bug Report] posible leak transaction when can not get some document
- The current logic create a new transaction , but if document can not get then only return false but the new transaction never is close 

#### Expected behavior
This problem cause leak transaction 

#### Other relevant information
- Your OS: MacOS
- Java version: jdk11
- ADempiere version: 3.9.4 to release
- Browser safari
- DB Provider: PG 
- DB version: 14

PR #3706, opened 2022-01-24
#3703 [Bug Report] posible leak transaction when Account Processor ca…

e-Evolution:
…n not get some document https://github.com/adempiere/adempiere/issues/3703


SusanneCalderon (approved):
@e-Evolution ,
I tested with process Client_AccountingProcessor. 
Simulating the error doc == null the trx was closed correctly and the process continued with a new transaction.
I tested in Swing and webUI.

Issue #3710, opened 2022-01-28
[Bug Report] Implement clause finally  in all blocks where a result set is used to avoid resource leaks


e-Evolution:
## Bug report
Implement clause finally  in all blocks where a result set is used to avoid resource leaks

#### Steps to reproduce
1- Search any block with ResultSet class and check that the clause finally is implemented

Incorrect : 

```
		try {
			PreparedStatement pstmt = DB.prepareStatement(sql, null);
			ResultSet rs = pstmt.executeQuery();
			rs.close();
			pstmt.close();
		} catch (SQLException e) { / /log.log(Level.SEVERE, "checkLogin", e);
		}

```

Correct : 

```
		PreparedStatement pstmt = null;
		ResultSet rs = null;
		try {
			PreparedStatement pstmt = DB.prepareStatement(sql, null);
			ResultSet rs = pstmt.executeQuery();
		} catch (SQLException e) {
			//	log.log(Level.SEVERE, "checkLogin", e);
		} finally {
			DB.close(rs, pstmt);
			rs = null;
			pstmt = null;
		}

```

#### Expected behavior
a database resource never should finally open

#### Other relevant information
- Your OS: MacOS
- Java version: jdk 11
- ADempiere version: 3.9.4 to release
- Browser [e.g. chrome, safari
- DB Provider: PG
- DB version: 14

PR #3711, opened 2022-01-28
 #3710 [Bug Report] Implement clause finally in all blocks where a re…

e-Evolution:
…sult set is used to avoid resource leaks https://github.com/adempiere/adempiere/issues/3710


e-Evolution:
implement the best approach into #3712

PR #3721, opened 2022-02-03
Inject flag for validate cached entities

yamelsenih:
This change just inject a flag for validate object managed as cache

The events are:
- For cache:
  - `put(key, value)`
  - `get(key)`
- For PO:
  - `get_TrxName()`
  - `set_TrxName()`


e-Evolution:
we implement other approach

PR #3726, opened 2022-02-04
Fixed compilation errors

yamelsenih:
Just a minor change for resolve compilation problems from eclipse


marcalwestf:
@yamelsenih  looks like a directory sneaked in:
Selection_067


This causes a compiler error.
After deleting the erpya directory, the compiler error disappears.

marcalwestf:
you're right @yamelsenih : running the command
_git clean -fd_ 
I got read of old sources.

Thanks!

marcalwestf (approved):
Compiler error solved.
Accepted.

PR #3731, opened 2022-02-09
First testing by cache transaction

yamelsenih:
This change just inject a flag for validate object managed as cache and return null when a object is cached

The events are:
- For cache:
  - `put(key, value)`
  - `get(key)`
- For PO:
  - `get_TrxName()`

### DIFF
--- a/base/src/org/compiere/model/PO.java
+++ b/base/src/org/compiere/model/PO.java
@@ -125,6 +125,24 @@ public abstract class PO
 	private static final String USE_TIMEOUT_FOR_UPDATE = "org.adempiere.po.useTimeoutForUpdate";
 
 	private static final int QUERY_TIME_OUT = 10;
+	private boolean isCachedEntity = false;
+	
+	
+	/**
+	 * Validate if this PO is managed from cache
+	 * @return
+	 */
+	public final boolean isCachedEntity() {
+		return isCachedEntity;
+	}
+
+	/**
+	 * Set value for cache flag
+	 * @param isCachedEntity
+	 */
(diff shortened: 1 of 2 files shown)


e-Evolution:
we try other approach

marcalwestf (approved):
Steps taken:
- build + install
- run Jetty
- On several windows: navigate on same window several times, de-referencing the same objects.
- Call several processes with same parameters.

marcalwestf:
@e-Evolution  hi Víctor, I looked at the code and tested it and it seemed OK.
I supposed it was the first step to more developments.

Which are the reasons for dismissing it and which is the other approach?
Thanks!!

e-Evolution:
@marcalwestf Hi Mario!

When the tests were done it showed strange behaviors, the reason is that any cached entity should be read only, unfortunately there is quite a lot of code that uses the get_TrxName method to use a transaction, in most cases the transaction with which the cache was generated and it was closed or commitment, going this way meant making a refactor too big since it would have to remove the code that uses a get_TrxName of an entity that was saved in the cache.

Currently the Cache is not a good implementation since it is not intended to be distributed, which does not allow mounting a distributed cluster.

The distributed cache is one of the most complicated problems in the software industry, so the best option must be evaluated and implemented.

Now I am analyzing:

https://hazelcast.com/ and https://ignite.apache.org/use-cases/in-memory-cache.html

On the other hand, the correct handling of the cache must be considered to avoid these errors, since before placing an application cache, the database caches or the connection pool must be exhausted.

here exist more detail :

https://www.scylladb.com/2017/07/31/database-caches-not-good/

PR #3735, opened 2022-02-13
Feature/migrate legacy code result set and query for MTree

### LINKED ISSUE #3723: [Feature Request] New way to run ResultSet on MTree
## Feature request

With the improvement of PR #3722, the use of the ResultSet in this class is changed due to the fact that the close is not being executed in a finally block, this can cause an error that the close of persistent objects is not executed correctly

To make sure that persistent objects are closed correctly, the function DB.runResultSet(...) is used, which closes them transparently.

#### Additional context

As it is now, if any errors happen, these lingers will not be closed because they are not in a finally block

#### How to validate

This section si lauched when you open the Menu Window as System Administrator.

#### Aditional Information

Hi @e-Evolution, forcing an error to be thrown, I could see how the debug went to the finally block to close the objects, and the NPE that I forced, followed as an exception backwards in the stack.

e-Evolution:
fixed #3723 [Feature Request] New way to run ResultSet on MTree https://github.com/adempiere/adempiere/issues/3723
- The implementation functional completely


marcalwestf (commented):
Tested on forms:
- BOM & Formula Info
- Tree Maintenance
- Workflow Editor

Tested on windows which implement trees:
- Accounting Element, tab Element Value
- Accounting Dimensions (all tabs)
- Organization

Tested on Servers
- Tomcat
- Wildfly
- Jetty  

All tests passed **except on Jetty**.
I run Jetty on different built+installed branches, commits and it worked well.
Only with this branch (feature/MigrateLegacyCodeResultSetAndQuery), Jetty did not respond.

Has this been tested on Jetty?

e-Evolution:
@marcalwestf can please send the log with Jetty , what is the problem ?

Please can set here  the Adempiere/utils/nohup.out and  Adempiere/utils/nohup.err

marcalwestf:
@e-Evolution  can you please test with Jetty?
I tried it several times, but I can't manage to get the Jetty server runninng.

Thanks!!

marcalwestf (approved):
Started anew on a clean server, with port free.
This time the Jetty Server was up and running.

I suppose there must have been something with my configuration.

**Test passed**

Issue #3712, opened 2022-02-01
[Bug Report] Incorrect handling of DB transactions in the order import process


e-Evolution:
## Bug report

Currently, the order import process is carried out in a very long transaction, causing the database to lock records for a long time, which is not operationally correct for proper concurrency operation.

A maxim for the correct handling of database transactions is that they have to be short-lived transactions so that the records have a minimum lock.

Any long-running Adempiere process should not under any circumstances use the process transaction to perform any updates or inserts in the database.

#### Steps to reproduce
1. Create 3 companies Acme01 , Acme02, Acme03
2. Create an import of 10000 orders with 100 lines each order
3. Ejecuta los procesos de importación de Ordenes en las tres compañias a lavez 

#### Expected behavior
All imported orders should be processed for all companies, without blocking the entire system

#### Other relevant information
- Your OS: macOS
- Java version: JDK 17
- ADempiere version: 394 with Hikari CP
- Browser: Safari
- DB Provider: PG
- DB version: 14

#### Additional context
Currently, the business logic for importing orders is not designed for a high concurrency environment.

PR #3743, opened 2022-02-25
#3742 [Feature Request] Add support the Import Order in Parallel and …

### LINKED ISSUE #3742: [Feature Request] Add support the Import Order in Parallel and Change the approach to import
## Feature request
Currently, the import process of orders is done in a very long transaction, since if there are many orders in the import, many system tables remain locked, on the other hand, the import is done sequentially and parallel processing is not used.

The new approach:

- Now the transactions shorter and this cause  releasing database locks

-  Import logic was changed,  using a parallel import, based on the document number and other fields that group the records to be imported.

- Now are searches all import order records sorted by Organization Id, Business Partner, Business Partner Location, Bill To, Document Type, and Document No.

- Mark all the records as in processing, so that if the import is executed by another user, it does not cause strange behavior when trying to import the same records by a user simultaneously.

- All records are grouping by fields (Organization Id, Business Partner, Business Partner Location, Bill To, Document Type, and Document No), a Order is create and import the order lines in a execution in to parallel with short transactions.

- Finally, at the end of creating the orders, apply the document action to the orders that were created without any errors in their lines.

- In the event that a line to be imported has an error, it is not imported but the exception is registered to document the problem, and this order is marked so that no document action is executed, leaving the order in a draft state.

#### Describe alternatives you've considered
Using the current approach causes long-term database locks, sequential processing makes processing slow, and does not take advantage of multi-processor servers.

#### Additional context

Tests of multiple companies were executed importing a large volume of orders with hundreds of lines simultaneously.

e-Evolution:
The new approach:

- Now the transactions shorter and this cause releasing database locks

- Import logic was changed, using a parallel import, based on the document number and other fields that group the records to be imported.

- Now are searches all import order records sorted by Organization Id, Business Partner, Business Partner Location, Bill To, Document Type, and Document No.

- Mark all the records as in processing, so that if the import is executed by another user, it does not cause strange behavior when trying to import the same records by a user simultaneously.

- All records are grouping by fields (Organization Id, Business Partner, Business Partner Location, Bill To, Document Type, and Document No), a Order is create and import the order lines in a execution in to parallel with short transactions.

- Finally, at the end of creating the orders, apply the document action to the orders that were created without any errors in their lines.

- In the event that a line to be imported has an error, it is not imported but the exception is registered to document the problem, and this order is marked so that no document action is executed, leaving the order in a draft state.

Fixed #3742
Relation with #3712


marcalwestf:
@e-Evolution : (for Eclipse developers: add library _vavr-1.0.0-alpha-4.jar_ to .classpath)

marcalwestf:
I tried to import both on Swing and ZK.
First, I will describe my findings on Swing.
I imported several times. Once, it didn't create Sales Orders, because of "No database connection".
image

The following images refer to another import, where the Sales Order headers were created, but not the lines. 

image

Some Orders where not imported.
On ZK more or less the same behavior.

I didn't try to debug, so I cannot pinpoint to the error.

Then I switched to the develop branch and I could import the Orders immediately.

marcalwestf (approved):
- build + install
- run on Jetty
- import and complete 50 on credit sales orders on one client
- import and complete 1000 on credit sales orders on 3 different clients at the same time
- check results

In all cases all sales orders were correctly created and completed.

**Test passed**

Issue #3769, opened 2022-03-18
[Bug Report] The Distribution Line flag isInvertAccountSign not work


e-Evolution:
## Bug report
[Bug Report] The Distribution Line flag isInvertAccountSign not work

#### Steps to reproduce
1.- If you create a distribution line with isInvertAccountSign in on or off the effect in distribution is the same

#### Expected behavior
The flag should invert the account  effect 

#### Other relevant information
- Your OS: macOS
- Java version: 11
- ADempiere version: 3.9.3
- Browser safari
- DB Provider: PG
- DB version: 14.2

PR #3770, opened 2022-03-19
#3769 [Bug Report] The Distribution Line flag isInvertAccountSign not…

e-Evolution:
… work https://github.com/adempiere/adempiere/issues/3769

### DIFF
--- a/base/src/org/compiere/acct/Fact.java
+++ b/base/src/org/compiere/acct/Fact.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 
+import org.compiere.model.I_C_BankStatement;
 import org.compiere.model.MAccount;
 import org.compiere.model.MAcctSchema;
 import org.compiere.model.MAcctSchemaElement;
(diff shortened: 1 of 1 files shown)


SusanneCalderon (commented):
@e-Evolution , I tested for "isallownegativeposting" and "isallownegativeposting" = "N". 
I found only a reversed accounting for the case IsAllowNegativePosting and notInverAccountsign.
So I never got the expected effect of reverse accounting.

Posting without distribution
imagen
imagen



- !IsAllowNegativePosting 
imagen
Case 1: Not Invert
imagen
Case 2: Invert AccountSign

Allow Negative posting

imagen
Case 1: Not Invert
imagen
Case 2: Invert

e-Evolution:
@SusanneCalderon This functionality is used when need create new accounting facts from original fact accounting:

Here you can see that when a accounting fact to 5.1.2.1.2.2 account is detected 

image


Then new facts accounting are created but with application type as budget

image


Here the last result of this GL Distribution 
image

This functionality was intended for government accounting, but can also be used to reclassify debit and credit to different ledger statistical accounts. 

kind regards
Victor

PR #3781, opened 2022-04-08
Remove Swing dependency of base source folder

yamelsenih:
The big problem of ADempiere is the bad use of dependency.
The **base** source folder should have only the **core** without any **UI** related like **Swing** or **ZK**.

## What is the problem?
The base source folder have many references to Swing UI like JPanel, JButton and others that make a problem if you want make a cleaned **core**

## What is the initial solution?
- Just move the swing dependence from base and change from base to client.
- Creeate interfaces or contracts to manage mandatory dependences like is the **View** inside **ReportEngine**
- Add Util class for provide generic methods used on **Swing** or **ZK**

Note: org.compiere.Adempiere was moved to org.adempiere.Adempiere and now is running from client source folder


yamelsenih:
New PR here #3784 3784

PR #3784, opened 2022-04-10
Remove Swing dependency of base source folder

yamelsenih:
The big problem of ADempiere is the bad use of dependencies.
The **base** source folder should have only the **core** without any **UI** related like **Swing** or **ZK**.

## What is the problem?
The base source folder has many references to Swing UI like _JPanel_, _JButton_, and others that cause a problem if we want to get a clean **core**.

## What is the initial solution?
- Move the swing dependence from _base_ and change from base to _client_.
- Create interfaces or contracts to manage mandatory dependencies like in the **View** inside **ReportEngine**
- Add _Util_ class to provide generic methods used on **Swing** or **ZK**

Note: _org.compiere.Adempiere_ was moved to _org.adempiere.Adempiere_ and now is running from _client_ source folder


marcalwestf:
@yamelsenih  Yamel, I think this is a good approach, which can help us in building and deploying.

Just one suggestion: The URL inside the HTML files has the domain _org_. Perhaps in the meantime, we should use _net_, what do you think about it?

marcalwestf (commented):
The new class to start from the IDE (e.g. Eclipse) is 
**_org.adempiere.Adempiere_**

imagen

marcalwestf (changes requested):
- Compile: OK
- Build: OK
- Install: OK
- Run _RUN_Adempiere.sh_: error

Cause: 
Selection_145

The class should be _org.adempiere.Adempiere_ instead of _org.compiere.Adempiere_

yamelsenih:
Hi Mario. Note that I add a note in the pill request:

> Note: org.compiere.Adempiere was moved to org.adempiere.Adempiere and now is running from client source folder

yamelsenih:
Is not a error, the ADempiere from controller has many references and just keep it.

The launcher instead I change it from package because the launcher have references to client and swing ui.

Just you should change package on launch and is ok

marcalwestf (approved):
- Install it on Tomcat, Jetty, and Firefly
- do some tests: Requisitions, Purchases, Material Receipts, Invoices, Material Movements, Sales, Accounting, Smart Browser processes, etc.
- Tests executed without errors

yamelsenih:
Hi @marcalwestf just now I will change it

yamelsenih:
Hi @marcalwestf this change was done: https://github.com/adempiere/adempiere/pull/3784/commits/a01aaf6626513453f3391460edfc854c2fe82361

PR #3792, opened 2022-04-13
fix: The Boolean constructor is deprecated.

EdwinBetanc0urt:
Currently there are multiple (102) warning messages concerning the deprecated boolean constructor since java version 9:

`The constructor Boolean(boolean) is deprecated since version 9`
`The constructor Boolean(String) is deprecated since version 9`

boolean-constructor

That may well work by replacing the constructor `new Boolean(boolean|String)` by `Boolean.valueOf(boolean|String)`.


marcalwestf (approved):
- syntax check for  every change
- check no compile errors introduced
- check reduced compiler warnings (now 4403)
- build
- run built image

Tests passed

PR #3800, opened 2022-04-15
fix: The Integer constructor is deprecated.

EdwinBetanc0urt:
Currently there are multiple (988) warning messages concerning the deprecated integer constructor since java version 9:

`The constructor Integer(int) is deprecated since version 9`
`The constructor Integer(String) is deprecated since version 9`

integer-constructor

That may well work by replacing the constructor `new Integer(int|String)` by `Integer.valueOf(int|String)`.


marcalwestf:
@EdwinBetanc0urt  Edwin, please do solve the conflicts so we start the accepting process
Thanks!

EdwinBetanc0urt:
Greetings @marcalwestf , I solved the conflict.

marcalwestf (approved):
-  syntax check for **every single change**
- check no compile errors introduced
- check reduced compiler warnings
- build
- run built image
- test on Smart Browsers, Special Forms, BOM Tree, Callouts, Packout, POS, etc.

Tests passed

PR #3801, opened 2022-04-15
fix: BigDecimal Round deprecated. 

EdwinBetanc0urt:
Currently there are multiple (625) warning messages concerning the deprecated bigdecimal round, set divide and set scale since java version 9:

`The method setScale(int, int) from the type BigDecimal is deprecated since version 9`
`The method divide(BigDecimal, int, int) from the type BigDecimal is deprecated since version 9`
`The field BigDecimal.ROUND_UP`
`The field BigDecimal.ROUND_EVEN`
`The field BigDecimal.ROUND_HALF_UP`
`The field BigDecimal.ROUND_HALF_DOWN`

big-decimal-deprecated

That can work by implementing the `RoundingMode` class

### DIFF
--- a/base/src/org/adempiere/engine/AbstractCostingMethod.java
+++ b/base/src/org/adempiere/engine/AbstractCostingMethod.java
@@ -4,6 +4,7 @@
 package org.adempiere.engine;
 
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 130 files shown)


marcalwestf:
@EdwinBetanc0urt  as these changes concern the costing calculation, thorough tests must be done.

EdwinBetanc0urt:
> @EdwinBetanc0urt as these changes concern the costing calculation, thorough tests must be done.

Greetings, add methods where the parameters for BigDecimal rounding calculations receive RoundingMode type instead of int, and leave depreciated methods with int, calling the new method using RoundingMode.valueOf(int)

marcalwestf (commented):
Looked at **every single change** and checked if there are inconsistencies.
Result: all changes are syntactically correct and consistent.
No compiler errors. 
Build correct.

The next step would be to test by calculating costs either by buying or by running lower lever costs on BOMs.

marcalwestf:
@e-Evolution  Víctor, could you please run some costing tests?

Issue #3806, opened 2022-04-25
[Feature Request]  Adding new fields Latitude, Longitude and Altitude  to Location to save the Geo Localization


e-Evolution:
## Feature request
 Adding new fields Latitude,  Longitude and Altitude to Location to save the Geo Localization

#### Describe the solution you'd like
 Adding new fields Latitude and Longitude to Location in Location Dialog

![image](https://user-images.githubusercontent.com/786968/165188762-dc719850-f23b-40cc-a64d-928d5bcacfe6.png)

#### Describe alternatives you've considered
Not exist some solution now

e-Evolution:
New Location Dialog :

![image](https://user-images.githubusercontent.com/786968/165216497-80c73539-351a-4faf-a73e-30e43da45067.png)

e-Evolution:
Update Location Window with new field

![image](https://user-images.githubusercontent.com/786968/169181899-67032284-962d-4460-b3ab-c6333cde73ed.png)

PR #3809, opened 2022-04-26
#3806 [Feature Request] Adding new fields Latitude, Longitude and Alt…

e-Evolution:
…itude to Location to save the Geo Localization https://github.com/adempiere/adempiere/issues/3806


marcalwestf (approved):
- Upload and apply migration file
- Create a new location
- Location + coordinates can be saved.

Test passed.

marcalwestf:
@e-Evolution  Victor, the next step would be to call Google maps with these coordinates immediately, wouldn't be?

PR #3814, opened 2022-04-28
Fix bugs generated in closers

### LINKED ISSUE #3812: Error generated when approaching unposted documents
https://github.com/adempiere/adempiere/blob/edfc3956cebd15c3edd7ab07374836619d477ec9/base/src/org/adempiere/model/GenericZoomProvider.java#L132

This method when querying the RV_UnPosted view, generates an error

image

dixonmartinez:
Fix bugs generated in closers

Fixes #3812

### DIFF
--- a/base/src/org/adempiere/model/GenericZoomProvider.java
+++ b/base/src/org/adempiere/model/GenericZoomProvider.java
@@ -140,7 +140,6 @@ public class GenericZoomProvider implements IZoomProvider {
 			query.addRestriction(targetTableName + "." + po.get_TableName() + "_ID=" + po.get_ID());
 		}
 
-		query.addRestriction(targetTableName + "." + po.get_TableName() + "_ID=" + po.get_ID());
 		if (tab.getWhereClause() != null && tab.getWhereClause().length() > 0)
 			query.addRestriction("(" + tab.getWhereClause() + ")");
 		query.setZoomTableName(targetTableName);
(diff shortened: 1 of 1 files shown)


marcalwestf:
hi @dixonmartinez tDixon thanks for your contribution.

If it is possible, could you please describe:

1. What do these changes specifically avoid/fix, i.e. in which context? 
2. How to test this pull request: which windows, reports, and processes must be run.

This helps a lot when testing.
Thanks a lot.

dixonmartinez:
Hi [@marcalwe](https://github.com/marcalwestf) 
I am emulating the case on a clean ADempiere database.

I have found the following
In the database in which the error is detected, the column C_Invoice_ID exists in the RV_UnPosted view, when clicking on the closer in the documents payable window, it throws an error indicating that the column does not exist in the RV_UnPosted view.

image

So in order to emulate the case you must do the following.
1.- Add the C_Invoice_ID column to the RV_UnPosted view
2.- Login as GardenAdmin or SuperUser
3.- Locate the Documents payable window
4.- Click on the record approacher
5.- If you have the source without the modification that I made, it should generate an error
6.- When evaluating with the modifications that I made, this error should not happen, since I put it to evaluate if there is a Record_ID and AD_Table_ID in the table that it tries to consult, if so, it will apply these filters.

dixonmartinez:
The other error if it could be emulated in a seed ADempiere database

For this case you must do the following:
1.- Go to the product window
2.- Click the zoom button
3.- Check the console or error log on the server and you will be able to observe the following
image

This is because when trying to evaluate the tab's where clause it doesn't parse it with the context.
ErrorProduct

yamelsenih:
I taste it and is ok

PR #3826, opened 2022-04-29
This pull request just add support to gradle builder to org.eevolution.warehouse source folder

yamelsenih:
The gradle structure is the follow:
- adempiereTrunk:
  - org.eevolution.warehouse:
    - build.gradle

Now you can run the follow command to base source folder:
```Shell
gradle build
```
The dependence from tools is:
```Java
    api 'io.github.adempiere:base:3.9.4-develop-1.1'
    api 'io.github.adempiere:manufacturing:3.9.4-develop-1.0'
```

See the result:
https://repo1.maven.org/maven2/io/github/adempiere/warehouse/

### DIFF
--- a/org.eevolution.warehouse/build.gradle
+++ b/org.eevolution.warehouse/build.gradle
@@ -1,0 +1,99 @@
+apply plugin: 'java-library'
+apply plugin: 'maven-publish'
+apply plugin: 'signing'
+
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+}
+
+if (JavaVersion.current().isJava8Compatible()) {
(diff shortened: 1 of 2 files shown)


yamelsenih:
The new printing method just implement the same approach that it https://github.com/adempiere/adempiere/blob/develop/org.eevolution.warehouse/src/main/java/base/org/spin/process/GenerateInvoiceInOutBound.java#L166

e-Evolution:
@yamelsenih Hello Yamel, I tried the functionality but it does not print the distribution order, so it is not working even though it asks if the document is going to be printed and it is confirmed that if the document is required to be printed, it does not generate the PDF of the distribution order to make the collection even if it is generated.

Before your proposed change it worked with this change it doesn't work

ReleaseOutboundOrder

PR #3830, opened 2022-05-03
Close Order does not update Total Lines Amount and total amount

### LINKED ISSUE #3810: Close Order does not update Total Lines Amount and total amount
## Bug report

When a Sales Order is Closed and it has a line with Delivered Quantity less than the Ordered Quantity, the line is updated, correcting the Ordered Quantity but it doesn't update the header of the order with the NEW  Total Lines Amount.

<br class="Apple-interchange-newline">

#### Screenshot or Gif

https://user-images.githubusercontent.com/31936815/165549155-89acb05e-3a68-4ea8-93ae-1621f4ab3aa7.mp4

#### Expected behavior
It should update the Total Lines Amount, the Grand Total and the Tax Order

nsarlabos:
Fixes #3810

### DIFF
--- a/base/src/org/compiere/model/MOrder.java
+++ b/base/src/org/compiere/model/MOrder.java
@@ -2184,7 +2184,13 @@ public class MOrder extends X_C_Order implements DocAction
 		//	Clear Reservations
 		if (!reserveStock(lines))
 		{
-			m_processMsg = "@Error@ @Undo@ @QtyReserved@ @From@ (@closed@)";
+			m_processMsg = "@FailedToUpdateReservations@";
+			return false;
+		}
+
+		if (!calculateTaxTotal())
+		{
+			m_processMsg = "@ErrorCalculatingTax@";
 			return false;
 		}
 
@@ -2240,8 +2246,14 @@ public class MOrder extends X_C_Order implements DocAction
 		//	Clear Reservations
 		if (!reserveStock(lines))
 		{
-			m_processMsg ="@Error@ @Undo@ @QtyReserved@ @From@ (@closed@)";
-			return "Failed to update reservations";
+			m_processMsg = "@FailedToUpdateReservations@";
+			return m_processMsg;
+		}
+
+		if (!calculateTaxTotal())
+		{
+			m_processMsg = "@ErrorCalculatingTax@";
+			return m_processMsg;
 		}
 
 		//Calculate Sizes (Weight & Volume)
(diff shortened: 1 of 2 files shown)


base/src/org/compiere/model/MOrder.java
```diff
@@ -2187,6 +2187,12 @@ public boolean closeIt()
 			return false;
 		}
 
+		if (!calculateTaxTotal())
+		{
+			m_processMsg = "Error calculating tax";
```
e-Evolution, line 2192: @nsarlabos Could you create messages in the dictionary so that the response to the user is multi-language?

`m_processMsg = "@ErrorCalculatingTax@";`

base/src/org/compiere/model/MOrder.java
```diff
@@ -2243,6 +2249,12 @@ public String reopenIt() {
 			return "Failed to update reservations";
```
e-Evolution, line 2249: @nsarlabos Could you create messages in the dictionary so that the response to the user is multi-language?

`m_processMsg = "@FailedToUpdateReservations@";`

base/src/org/compiere/model/MOrder.java
```diff
@@ -2243,6 +2249,12 @@ public String reopenIt() {
...
 		}
 
+		if (!calculateTaxTotal())
+		{
+			m_processMsg = "Error calculating tax";
```
e-Evolution, line 2254: @nsarlabos Could you create messages in the dictionary so that the response to the user is multi-language?

m_processMsg = "@ErrorCalculatingTax@";

base/src/org/compiere/model/MOrder.java
```diff
@@ -2243,6 +2249,12 @@ public String reopenIt() {
...
+			return "Error calculating tax";
```
e-Evolution, line 2255: here you should return m_processMsg;  vs same text

e-Evolution (changes requested):
@nsarlabos Could you create error messages in the dictionary so that the response to the user is multi-language?

391 / 5.000
Resultados de traducción
The change seems correct to me, it is only better to do it multi-language.

1.- Enables the generation of Dictionary Maintenance and Migration Log Script
2.- Create the error messages by placing the word between @@ as a key, type of message Error and the message text of the error message
3.- Generates the XML for the changes in the dictionary.
4.- Send the changes in this pull request

nsarlabos:
Hi @e-Evolution I have made the requested changes

e-Evolution:
@nsarlabos Thank you to accept the suggestion

PR #3835, opened 2022-05-06
Add P_ADEMPIERE_APPS_TYPE support to Ini loader

yamelsenih:
Currently exists a enviroment variable added to ADempiere connection pool, the problem is that it jus is setted as OS variable instead of Ini loader like is all ini enviroment of adempiere. This pull request add support to Ini loader and keep the OS enviroment as second option.

The follow error is throwed if not exist the enviroment variable:
```Shell
org.adempiere.exceptions.AdempiereException: The ADEMPIERE_APPS_TYPE environment variable is not set, so it is not possible to initialize the Hikari Connection Pool
	at org.compiere.db.DB_PostgreSQL.lambda$getDataSource$1(DB_PostgreSQL.java:639)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at org.compiere.db.DB_PostgreSQL.getDataSource(DB_PostgreSQL.java:639)
	at org.compiere.db.DB_PostgreSQL.getFromConnectionPool(DB_PostgreSQL.java:532)
	at org.compiere.db.CConnection.getConnection(CConnection.java:1363)
	at org.compiere.util.DB.createConnection(DB.java:463)
	at org.compiere.util.DB.getConnectionRW(DB.java:393)
	at org.compiere.util.DB.isConnected(DB.java:361)
	at org.compiere.util.CLogErrorBuffer.publish(CLogErrorBuffer.java:211)
	at java.logging/java.util.logging.Logger.log(Logger.java:980)
	at java.logging/java.util.logging.Logger.doLog(Logger.java:1007)
	at java.logging/java.util.logging.Logger.log(Logger.java:1030)
	at java.logging/java.util.logging.Logger.severe(Logger.java:1777)
	at org.compiere.db.CConnection.getConnection(CConnection.java:1367)
	at org.compiere.util.DB.createConnection(DB.java:463)
	at org.compiere.util.DB.getConnectionRO(DB.java:412)
	at org.compiere.model.POInfo.getPOInfoColumnList(POInfo.java:355)
	at org.compiere.model.POInfo.loadInfo(POInfo.java:173)
	at org.compiere.model.POInfo.<init>(POInfo.java:129)
	at org.compiere.model.POInfo.getPOInfo(POInfo.java:84)
	at org.compiere.model.X_AD_Client.initPO(X_AD_Client.java:79)
	at org.compiere.model.PO.<init>(PO.java:223)
	at org.compiere.model.PO.<init>(PO.java:183)
	at org.compiere.model.X_AD_Client.<init>(X_AD_Client.java:38)
	at org.compiere.model.MClient.<init>(MClient.java:122)
	at org.compiere.model.MClient.<init>(MClient.java:153)
	at org.compiere.model.MClient.get(MClient.java:77)
	at org.compiere.model.MClient.get(MClient.java:106)
	at org.compiere.model.MCountry.loadAllCountries(MCountry.java:128)
	at org.compiere.model.MCountry.getCountries(MCountry.java:114)
	at org.spin.grpc.service.AccessServiceImplementation.<init>(AccessServiceImplementation.java:87)
	at org.spin.server.AllInOneServices.start(AllInOneServices.java:64)
	at org.spin.server.AllInOneServices.main(AllInOneServices.java:179)
===========> CConnection.getConnection: Cannot invoke "javax.sql.DataSource.getConnection()" because "this.datasourceLongRunning" is null [1]
===========> CConnection.getConnection: Cannot invoke "javax.sql.DataSource.getConnection()" because "this.datasourceLongRunning" is null [1]
===========> CConnection.getConnection: Cannot invoke "javax.sql.DataSource.getConnection()" because "this.datasourceLongRunning" is null [1]
```

### DIFF
--- a/base/src/org/compiere/db/DB_MariaDB.java
+++ b/base/src/org/compiere/db/DB_MariaDB.java
@@ -524,7 +524,7 @@ public class DB_MariaDB implements AdempiereDatabase {
 				datasourceLongRunning = new HikariDataSource(config);;
 				log.warning("Starting Client Hikari Connection Pool");
 			} else {
-				Optional<String> maybeApplicationType = Optional.ofNullable(System.getenv("ADEMPIERE_APPS_TYPE"));
+				Optional<String> maybeApplicationType = Optional.ofNullable(Optional.ofNullable(Ini.getApplicationType()).orElse(System.getenv("ADEMPIERE_APPS_TYPE")));
 				datasourceLongRunning = maybeApplicationType
 						.map(applicationType -> {
 							if ("wildfly".equals(applicationType)) {
@@ -603,7 +603,7 @@ public class DB_MariaDB implements AdempiereDatabase {
 				datasourceShortRunning = cpds;
 				log.warning("Starting Client Hikari Connection Pool");
 			} else {
-				Optional<String> maybeApplicationType = Optional.ofNullable(System.getenv("ADEMPIERE_APPS_TYPE"));
+				Optional<String> maybeApplicationType = Optional.ofNullable(Optional.ofNullable(Ini.getApplicationType()).orElse(System.getenv("ADEMPIERE_APPS_TYPE")));
 				datasourceShortRunning = maybeApplicationType
 						.map(applicationType -> {
 							if ("wildfly".equals(applicationType)) {
--- a/base/src/org/compiere/db/DB_MariaDB.java
+++ b/base/src/org/compiere/db/DB_MariaDB.java
@@ -524,7 +524,7 @@ public class DB_MariaDB implements AdempiereDatabase {
 				datasourceLongRunning = new HikariDataSource(config);;
 				log.warning("Starting Client Hikari Connection Pool");
 			} else {
-				Optional<String> maybeApplicationType = Optional.ofNullable(Optional.ofNullable(Ini.getApplicationType()).orElse(System.getenv("ADEMPIERE_APPS_TYPE")));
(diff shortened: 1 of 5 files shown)


e-Evolution:
@yamelsenih this change is not necessary ADEMPIERE_APPS_TYPE is an initialization variable that is inside AdempiereEnv.properties defined when the RUN_silentsetup is performed, so it is defined by startup not by operating system configuration.

Putting this at the operating system level can cause problems on deployments of the same host with different application servers.

If you have this problem check that ADEMPIERE_APPS_TYPE is inside your AdempiereEnv.properties

e-Evolution:
@yamelsenih I review your code and is right if only the ADEMPIERE_APPS_TYPE is defined from Optional.ofNullable(Ini.getApplicationType()) but not from System.getenv("ADEMPIERE_APPS_TYPE"))

It seems that I define it like this in the first instance and then I don't change it but it is correct that I use Ini class

yamelsenih:
Hi @e-Evolution it was changed

e-Evolution:
@yamelsenih  I revert this changes, it is causing problem with Wildfly https://github.com/adempiere/adempiere/issues/3862

The problem is that Ini.java takes the values from Adempiere.properties when these settings are in AdempiereEnv.properties so the change is not working. so since the AdempiereEnv.properties variables are set by myEnvironment.sh taking the value directly from the system is correct, each deployment takes the current configuration

PR #3838, opened 2022-05-07
Validate Class Name for document tables

yamelsenih:
This pull request just add a validation to generate model class with
correct name:

The previous problem is that if a table like C_Invoice was generated, the model class generated is MCInvoice instead MInvoice (and validating it).

Also allows validate core packages to generate references


marcalwestf (approved):
- Created different models for different tables
- Files are generated correctly
- Test passed
imagen

PR #3845, opened 2022-05-09
Just resolve the problem for print a document

yamelsenih:
Currently adempiere can send document from a process to print. All is ok but if you try send a document and avoid the ask for printing, it always ask if you want print document instead print directly. A example for it can be it:

Note that the follow gif show always the dialog but it the method was called with "askPrint" parameter as false
```Java
getPrintFormatId("Manufacturing_Order_Workflow_Header ** TEMPLATE **", "PP_Order_Workflow_Header_v"), false);
```

Print Release Order

### DIFF
--- a/org.eevolution.manufacturing/src/main/java/ui/swing/org/eevolution/form/VPrintDocument.java
+++ b/org.eevolution.manufacturing/src/main/java/ui/swing/org/eevolution/form/VPrintDocument.java
@@ -45,26 +45,39 @@ import org.eevolution.service.dsl.ProcessBuilder;
 public class VPrintDocument implements IPrintDocument {
     @Override
     public void print(PO document, int printFormatId, int windowNo, boolean askPrint) {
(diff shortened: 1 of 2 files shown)


marcalwestf (commented):
- look at code
- in all cases, the parameter _askPrint_ is considered, which was not considered before
- now, when _askPrint_==false, it is printed immediately, and when _askPrint_==true, the "ask" Dialog is displayed.
- build is OK
- Still to do is a test on both ZK and Swing

marcalwestf (approved):
Steps to Test
- Create a Sales Order
- Create a WMInOutBound based on Sales Order
- Create a Distribution Order
- Create Shipment from WMInOutBound
- Debug VPrintDocument
- Test process with askPrint ==_true_ and _false_

Result: askPrint controls whether is direct printed or there is a question before.
WPrintDocument changes are analogous to VPrintDocument.

Selection_165
Selection_166
Selection_167

PR #3846, opened 2022-05-09
Add a new method to set table selection Process Builder

yamelsenih:
Just add a minimal change to add table selection support on process builder

Now you can use two methods to set selection

```Java
public ProcessBuilder withSelectedRecordsIds(int tableSelectionId, List<Integer> selectedRecordsIds);
public ProcessBuilder withSelectedRecordsIds(int tableSelectionId, String aliasTableSelection, List<Integer> selectedRecordsIds);
```


marcalwestf (approved):
- Code check
- Add a new signature to method _withSelectedRecordsIds_
- Enhancement: _aliasTableSelection_ can be added

Changes accepted

PR #3847, opened 2022-05-09
Add support to default PDF to Process Info

yamelsenih:
A minor change that allows define the PDF file reference to `pdfReportFile` of Process Info when the `reportAsFile` object have a file with .pdf extension


marcalwestf (approved):
Code Review:
- class variable _pdfReportFile_ is set when file ends with ".pdf"
- Additional call of setPDFReport(File pdfFile) unnecessary
- Functionality improved

Changes accepted

PR #1854, opened 2018-08-02
#1853, Add feature for open result from process, reference to issue:

yamelsenih:
https://github.com/adempiere/adempiere/issues/1853

Hi Everybody, this change allows open a result of process directly after run process, see step by step:

I want know what is business partner that was updated after run **Validate Business Partner** process, for make it just i add some lines to process:

Add information or ID os business partner processed
<pre>
addLog(bp.getC_BPartner_ID(), null, bp.getSO_CreditUsed(), Msg.getElement(getCtx(), "SO_CreditUsed"));
</pre>
screenshot_20180801_213446
Show result after process
<pre>
openResult(I_C_BPartner.Table_Name);
</pre>
screenshot_20180801_213430

**Running process**

 - Open Validate **Business Partner** process
screenshot_20180801_213528
 - After process view result
screenshot_20180801_213544

The window linked to table C_BPartner is opened after runin process.

Nice.

Best regard

PR #3848, opened 2022-05-09
Prevent NPE when openResult is called

yamelsenih:
This pull request just prevent a NPE when the `openResult` method is called without records.

The openResult feature was added by me here #1854


marcalwestf (approved):
- Checked Syntax
- No compiler errors
- Build OK

Test passed

PR #3849, opened 2022-05-09
Add support to process documents after process line

yamelsenih:
Add a flag that allows save a record after process a document instead of save it by each line processed, is very useful when you have a export format with embedded export formats, a example can be:

- Export Format: **C_Order**
  - Export Format Line: **C_Invoice**

When the order is created, the record is created as draft but the invoice is created just after create order and throw a error because the order is drafted.

A basic example:
 - Create Order
 - Add a line
 - Create invoice
 - Add line of invoice
 - Process Invoice: >>> Error, the sales order is drafted

This change allows that the order is processed and after it the invoice is created and processed.

After change the example is:
 - Create Order
 - Add a line
 - Process Ooder
 - Process Invoice
 - Create invoice
 - Add line of invoice

### DIFF
--- a/base/src/org/adempiere/util/rpl/EntityWrapper.java
+++ b/base/src/org/adempiere/util/rpl/EntityWrapper.java
@@ -240,10 +240,15 @@ public class EntityWrapper {
      * @param key
      * @return
      */
-    public EntityWrapper getReference(String key) {
+    @SuppressWarnings("unchecked")
+	public EntityWrapper getReference(String key) {
     	Object value = attributes.get(key);
-    	if(value instanceof EntityWrapper) {
+    	if(EntityWrapper.class.isAssignableFrom(value.getClass())) {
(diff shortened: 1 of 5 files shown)


marcalwestf:
@yamelsenih , @e-Evolution  perhaps you can give me a hint on how to test this.
I want to come through the methods of both classes _EntityWrapper_ and _EntityWrapperFactory_, but I haven't managed to do it.

Here is what I have hitherto done:
- Create a Replication Strategy
- Create Export Formats
- Then I tried testing the export model, creating both order and invoice and adding lines to both, and repeating the export model.

Another way to test them would be by means of a main() method for both classes.
Is there a step-by-step description of Replication?
Thanks

See images:
1- Create a Replication Strategy
imagen
imagen

2- Create Export Formats
imagen
imagen

yamelsenih:
This pr will be rewrited

PR #3850, opened 2022-05-09
Just change rounding up to decimal 128

yamelsenih:
Currently the `rate` method of CalloutEngine have a rounding mode to half up, this pull request just add a decimal 128 of digits because a rounding can be little explicit if you use it on conversion rates and others functionalities with many difference.

This error affect to countries with many difference between two currency, example:
USD **1** ~ VES **4,500,000.239**
With a conversion from USD to VES

### DIFF
--- a/base/src/org/compiere/model/CalloutEngine.java
+++ b/base/src/org/compiere/model/CalloutEngine.java
@@ -18,7 +18,7 @@ package org.compiere.model;
 
 import java.lang.reflect.Method;
 import java.math.BigDecimal;
-import java.math.RoundingMode;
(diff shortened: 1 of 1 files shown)


marcalwestf (approved):
- Open the _Product_ window, _UOM Conversion_ tab
- Create a new _Conversion_ entry
- Set a **divider** of 3 to get a very big mantissa
- Debug _CalloutEngine.rate()_
- Observe results on both _CalloutEngine.rate()_ and  _Product_ window, _UOM Conversion_ tab

Result: values correct.
Test passed

imagen
imagen

Issue #356, opened 2016-04-29
Decrease verbosity of SQL statement closing lines.


trifonnt:
Many places in the source code we can find lines like below:
```java
		String sql = "SELECT * FROM AD_Table WHERE UPPER(TableName)=?";
		PreparedStatement pstmt = null;
		try
		{
			pstmt = DB.prepareStatement (sql, null);
			pstmt.setString(1, tableName.toUpperCase());
			ResultSet rs = pstmt.executeQuery ();
			if (rs.next ())
				retValue = new MTable (ctx, rs, null);
			rs.close ();
			pstmt.close ();
			pstmt = null;
		}
		catch (Exception e)
		{
			s_log.log(Level.SEVERE, sql, e);
		}
		try
		{
			if (pstmt != null)
				pstmt.close ();
			pstmt = null;
		}
		catch (Exception e)
		{
			pstmt = null;
		}
```

Better approach is code to be re-factored to:

```java
		String sql = "SELECT * FROM AD_Table WHERE UPPER(TableName)=?";
		PreparedStatement pstmt = null;
		ResultSet rs = null;
		try {
			pstmt = DB.prepareStatement (sql, null);
			pstmt.setString(1, tableName.toUpperCase());
			rs = pstmt.executeQuery ();
			if (rs.next ())
				retValue = new MTable (ctx, rs, null);
		} catch (Exception e) {
			s_log.log(Level.SEVERE, sql, e);
		} finally {
			DB.close(rs, pstmt);
		}
```

yamelsenih:
Hi @trifonnt I think that is better change it by MTable.get (Properties ctx, String tableName), and change the method in MTable for a that use Query class.

Best regards.

Issue #223, opened 2016-02-24
Bug fix when a format for Order , Shipment and Invoice is defined as Jasper report the report not is show


e-Evolution:
Case:

Define print formant and set an jasper report , Adempiere create Report Engine using the Sub type Order document , but when try execute of jasper report for shipment , the system send of record id for C_Order_ID but the report engine create an Query based on shipment table (M_InOut) and primary key [M_InOut_ID] , So when try execute of query the record not found because record id is based on C_Order_ID.

Solution if report engine have an Query , then of record id is take from query object and record id that this case is C_Order_ID is replace for M_InOut_ID.

PR #3851, opened 2022-05-09
Add conversion amount of payment to commission

yamelsenih:
Currently the commission defined as completely paid don't have support to converted amount and result a wrong calculation of documents when the payments are in many currencies. This pull request resolve it.

## Step by Step
- Apply this [XML](https://github.com/adempiere/adempiere/pull/3851/files#diff-7f5122d7260cd6b20ff2a6197082f5f49c75232316c2f5d90187ca1b9f25d8da)
- Go to **Commission Definition** window
- Note that exists a new commission definition named **Example of 10 %**
- The commission definition **Example of 10 %** is configured:
  - Calculation Basis: Receipt
  - Frequency Type: Yearly
  - Paid Totally
  - The sales representative tab has a Sales Rep. **GardenUser BP**
Screenshot from 2022-05-10 19-06-51
- Go to Currency window
- Find the **VED** currency
- Go to **Conversion** tab
- See that exists a conversion from **USD** to **VED** and viceversa
  - **USD** to **VED**: multiply rate **1,500.00** >> **1 USD** ~ **1,500.00 VED**
  - **VED** to **USD**: divide rate **1,500.00** >> **1 VED** ~ **0.000666667 VED**
- Go to Invoice (Customer):
  - See the invoice number **1000**
    - Currency **USD**
    - Amount **USD 1,000.00**
- Go to Payment:
  - See the payment number **400002**
    - Reference to invoice **1000**
    - Currency **VED**
    - Amount **VED 1,000.00**
    - Converted Amount: **USD 0.666666667**
- Go to Commission Run
- See a commission run with document no **100**
- This was processed usin the old source code
- The amount is calculated from 1000 that is the totally paid on **USD** (Here the error)
See the open item:
Screenshot from 2022-05-10 19-26-43

Note: If you use the pull request version and process the commission it don't generate nothing because the payment amount converted to invoice amount is less of invoice.


marcalwestf:
@yamelsenih thanks for the detailed description.
I will follow it and refer to it when reviewing.

marcalwestf (commented):
**Code check**
One question @yamelsenih : in the original code, all lines of an invoice are ignored when the commission is defined for invoices completely paid, and when the invoice hasn't been completely paid.

In the code suggested, this filter has been eliminated:
imagen

Later in the code, the commission is considered for these cases with _InvoiceopenToDate()_:
imagen

My question is that in the lines there would be (in the worst case many) invoices listed which will have no values because they belong to invoices that haven't been completely paid. This can cause confusion.

I'd like to know first if my assumption is correct and if it is, the reason for the modification.

Thanks

yamelsenih:
> **Code check** One question @yamelsenih : in the original code, all lines of an invoice are ignored when the commission is defined for invoices completely paid, and when the invoice hasn't been completely paid.
> 
> In the code suggested, this filter has been eliminated: imagen
> 
> Later in the code, the commission is considered for these cases with _InvoiceopenToDate()_: imagen
> 
> My question is that in the lines there would be (in the worst case many) invoices listed which will have no values because they belong to invoices that haven't been completely paid. This can cause confusion.
> 
> I'd like to know first if my assumption is correct and if it is, the reason for the modification.
> 
> Thanks

Hi @marcalwestf this was removed because a invoice can be over paid and should list for run commission.

Invoice | Amount | Payment | Payment Amount | Over Payment | IsPaid
-- | -- | -- | -- | -- | --
#356 | 100 | #223 | 150 | 50 | false

marcalwestf:
@yamelsenih thanks for the clarification.

Could you please merge your branch with the latest development branch?
The _MCommissionRun_ class has been updated lately on the latest development branch, and as I don't exactly know how the checks here work (i.e. if the action just executes on the pull requested branch, or if there is a migration before execution of the action), I fear there might be conflicts when merging.

I re-started the checks because of this and they were OK.

marcalwestf:
@yamelsenih small change: the sequence and the file name should be the same. In this case, the sequence should be "9480".
Also, the migration name should contain the word "completely" instead of "completelly". 
imagen

marcalwestf (changes requested):
When executing the XML, the following error appears:
Selection_191

The column _C_Currency_ID_To_  is not defined in table _C_Currency_:
imagen



The column _C_Currency_ID_To_  has only been defined in table _C_Conversion_Rate_:
imagen

Thus, the definition of column _C_Currency_ID_To_  in table _C_Currency_ is missing.

yamelsenih:
> When executing the XML, the following error appears: Selection_191
> 
> The column _C_Currency_ID_To_ is not defined in table _C_Currency_: imagen
> 
> The column _C_Currency_ID_To_ has only been defined in table _C_Conversion_Rate_: imagen
> 
> Thus, the definition of column _C_Currency_ID_To_ in table _C_Currency_ is missing.

Hi Mario, maybe you can update the seed? I see that the currency 50003 alrready exists after [this pull request](https://github.com/adempiere/adempiere/pull/3593) but in your database is missing.

See the id here: https://github.com/adempiere/adempiere/pull/3593/files#diff-02dd42a913ebfa2c0177ec1880f26f0036ab95a435a74798508392e21461af70R5

marcalwestf (approved):
- Loaded and Executed XMLs _09230_Add_new_Currency_for_Venezuela_Bolivar_Digital.xml_ and _09480_Example_of_Commission_calculation_completelly_paid.xml_
- Check commission definition _Example of 10 %_ -> OK
- Check Currency _VED_ and Conversion -> OK
- Check _Invoice_ -> OK
- Check _Payment_ -> OK
- Check _Commission Run_  -> OK
- Execute action _Prepare_ on Commission Run -> OK
- Check that commission line was deleted -> OK

Test passed

PR #3855, opened 2022-05-10
Add support to update sequence on Pay Print/Export option

### LINKED ISSUE #3854: [Bug Report]  Payment Print/Export don't update sequence at export option
## Bug report
Currently at generate a payment in Payment Print/Export only update sequence if the selected option is print
#### Steps to reproduce

1. Create an account bank document with a payment rule distinct to check and set a current sequence

2. Create a payment selection with the bank account and payment rules previously setting

3. In the Payment Print/Export window selected the Payment Selection previously generated, Debit Direct Payment Rule and Export option.

5.  After Export the Payment Selection is generated correctly, but the Bank Account Document is no update the sequence.

6. Now In the Payment Print/Export window selected the Payment Selection previously generated, select the Check Payment Rule and Export option

7.  After Export the Payment Selection is generated correctly, and the Bank Account Document is update the sequence successful.

#### Expected behavior
The document number must be update in ever of the cases, because is very used for bank reconciliation and if this don't increase ever , the payments ever will set the same document number

- Your OS: Debian
- Java version: 11
- ADempiere version: 3.9.3
- DB Provider: Postgres
- DB version: 11

carlosaparadam:
Fixes #3854

### DIFF
--- a/base/src/org/compiere/model/MBankAccount.java
+++ b/base/src/org/compiere/model/MBankAccount.java
@@ -18,6 +18,7 @@ package org.compiere.model;
 
 import java.sql.ResultSet;
 import java.util.ArrayList;
+import java.util.Optional;
 import java.util.Properties;
 
 import org.adempiere.exceptions.AdempiereException;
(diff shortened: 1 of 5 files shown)


SusanneCalderon (approved):
Following steps to test:

- Check value of current next in Bank Account Document Tab (1000)
- Generate a Pay Selection with 3 invoices of two providers, bank account 1234 of money bank
- Complete the Pay Selection
- Call the pay Selection in the form Print/Export and select the pay selection
- Field DocumentNo is set with value 1000
- Print the 2 checks
- Debug VPayPrint, PayPrint, MBankAccount and MBankAccountDoc
- Check the generated payments (check no is set to 1000 and 1001)
- Check tab BankAccountDoc, value of currentNext is set to 1002 as expected.

Test passed

marcalwestf (approved):
- Do the same steps as the test executed by @SusanneCalderon 
- Instead of printing do select export
- Results also increment sequence

Test passed

PR #3857, opened 2022-05-11
Add support to get document base type with transaction name

carlosaparadam:
### This pull request just add two changes:

- Get document base type record with transaction name
- Prevents NPE at validate Document Type if Defined Document Base Type don't exists.


marcalwestf (approved):
- check code -> OK
- methods with named transaction are added -> OK
- Open window _Document Type_
- Create and manipulate a _Document Type_ such as the method _validDocBaseType()_ is called
- Save
- Debug/Check whether new methods are called and work correctly
- Records are correctly saved

Tes passed

PR #3875, opened 2022-05-19
Add Support to Login with OpenID Connect

carlosaparadam:
## Login with [Open ID Connect  ](https://openid.net/connect/)

This functionality allow login in Adempiere ZK OpenID Connect, specifically with this services:

- Github
- Google

### Setting for Github
---

- With your Github account, go to **Settings / Developer Setting / OAuth Apps**  and create an application

- Set the **Authorization callback URL** the url of Adempiere ZK

- Create an Application Registration with this characteristics

- Set the parameters with the information generated on Github (Client Identifier and Secret), in the Redirect URL set the url of Adempiere ZK

 
### Setting for Google
---

- In your google console, go to **Credentials / Create Credentialas / Client ID Oauth** 

- Add a Redirect URI with the url of Adempiere ZK
- Create an Application Registration with this characteristics

- Set the parameters with the information generated on Google (Client Identifier and Secret), in the Redirect URL set the url of Adempiere ZK

### Test functionality
---

- Logout and see the Adempiere Login 

- Select Login With Google and select an account or login with your google account

- See the error message if not exists an login user registered with this email

- Create an user with a valid email

- Add a role for the user

- Logout and **Login With Google** and selected a valid email for Adempiere 
- After login this is redirect for Role selection panel

- Logout and **Login With Github** and selected a valid account for Adempiere 

- After login this is redirect for Role selection panel


e-Evolution:
Hi Carlos @carlosaparadam I try testing this functionality 

I create a user with my email and enable permission to github.com but when I clicked Login with Gitihub

image

this redirect to github.com page 

I set my github password  but always got 

image

I think something is missing

carlosaparadam:
hi Victor @e-Evolution, i think so that query for find the user is for value and in the value is `victor.perez`

e-Evolution:
@carlosaparadam I configured the google Application client

image

but when try login with google 

image

Here I think that you need to fix the label to set "Login with Google" vs only "Login Google"

When I clicked Login Google, I get un redirect but get this error:

image

I think something is missing, Have you some idea of the problem.

carlosaparadam:
Hi Victor @e-Evolution, the tag is application registration name you can set the tag do you like, please would you show the error on `Detalles de la Solicitud`

Regards

e-Evolution:
> hi Victor @e-Evolution, i think so that query for find the user is for value and in the value is `victor.perez`

I got the success login using GitHub account, In this case an import note is that the search key for User should be the same that the GitHub account.

Thank you.

e-Evolution:
> > hi Victor @e-Evolution, i think so that query for find the user is for value and in the value is `victor.perez`
> 
> I got the success login using GitHub account, In this case an import note is that the search key for User should be the same that the GitHub account.
> 
> Thank you.

I forgot to confirm that the label that appears is the same that you put in the name of the App Register

image

e-Evolution:
> Hi Victor @e-Evolution, the tag is application registration name you can set the tag do you like, please would you show the error on `Detalles de la Solicitud`
> 
> Regards

My problem was setting the incorrect redirect, I  change, but I had that restart the server to get the new configuration I got success the login with a Google account, Thank you.

e-Evolution:
@carlosaparadam Hi Carlos , @yamelsenih Yamel.

@carlosaparadam Hi Carlos, @yamelsenih Yamel.

Here are my observations to accept this great functionality.

Since the new tables that were added are part of the dictionary when selecting the AD_ prefix, they should have the Dictionary entity inside the applications dictionary, and it should be part of the core, so instead of calling the MADUserAuthentication model, it should call MUserAuthentication, for this it should be in the org.adempiere.model package

on the other hand, to be consistent in the package names of the business logic domain classes, we have adopted the package name as service, for example

currently defined as:

org.spin.util.support.authentication;
org.spin.authentication.support.provider;

and it should be:

org.spin.authentication.service;
org.spin.authentication.service.provider;

thus, if the functionality is not a process, it must be added to the service package as a business logic that belongs to a domain service.

This improvement works very well, excellent work. Carlos, thank you very much for your contribution to the project.

carlosaparadam:
Hi Victor @e-Evolution, the requested changes has made

Issue #3902, opened 2022-06-03
[Bug Report] ADempiere WebStore login.jsp [${initParam.#AD_Client_ID}] contains invalid expression(s): 


e-Evolution:
## Bug report
ADempiere WebStore login.jsp [${initParam.#AD_Client_ID}] contains invalid expression(s)

#### Steps to reproduce
1.- RUN_setup.sh and deployment with Jetty
2.- Check the file   utils/nohup.err  using command #more  nohup.err

#### Screenshot or Gif
2022-06-03 16:56:58.057:WARN :oejshC.wstore_war:main: unavailable
org.apache.jasper.JasperException: /login.jsp (line: [66], column: [60]) [${initParam.#AD_Client_ID}] contains invalid expression(s): [javax.el.ELException: Failed to parse the expression [${initParam.#AD_Client_ID}]]
        at org.apache.jasper.compiler.DefaultErrorHandler.jspError(DefaultErrorHandler.java:41)
        at org.apache.jasper.compiler.ErrorDispatcher.dispatch(ErrorDispatcher.java:292)
        at org.apache.jasper.compiler.ErrorDispatcher.jspError(ErrorDispatcher.java:98)
        at org.apache.jasper.compiler.Validator$ValidateVisitor.getJspAttribute(Validator.java:1443)
        at org.apache.jasper.compiler.Validator$ValidateVisitor.checkXmlAttributes(Validator.java:1252)

#### Expected behavior
When starting the server with Jetty it should not show any error

#### Other relevant information
- Your OS: macOS
- Java version: 11 and 17
- ADempiere version: develop to release 3.9.4
- Browser safari
- DB Provider: PG
- DB version: 14

PR #3903, opened 2022-06-03
#3902 [Bug Report] ADempiere WebStore login.jsp [${initParam.#AD_Clie…

e-Evolution:
…nt_ID}] contains invalid expression(s): https://github.com/adempiere/adempiere/issues/3902

### DIFF
--- a/webStore/src/web/assets.jsp
+++ b/webStore/src/web/assets.jsp
@@ -33,7 +33,7 @@
 	  <c:if test='${not webUser.EMailVerified}'>
         <form action="emailServlet" method="post" enctype="application/x-www-form-urlencoded"
 	  	  name="EMailVerification" target="_top">
-          	<input name="AD_Client_ID" type="hidden" value='<c:out value="${initParam.#AD_Client_ID}" default="0"/>'/>
+          	<input name="AD_Client_ID" type="hidden" value='<c:out value="${initParam['#AD_Client_ID']}" default="0"/>'/>
           	<input name="Source" type="hidden" value=""/>
           	<input name="Info" type="hidden" value=""/>
             <input name="ForwardTo" type="hidden" value="assets.jsp">
--- a/webStore/src/web/emailVerify.jsp
+++ b/webStore/src/web/emailVerify.jsp
@@ -31,7 +31,7 @@
 	  <c:if test='${not webUser.EMailVerified}'>
         <form action="emailServlet" method="post" enctype="application/x-www-form-urlencoded" 
 	  	  name="EMailVerification" target="_top">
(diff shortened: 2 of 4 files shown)


marcalwestf (commented):
- build an executable
- install with Oracle's JDK 17 and Jetty 10.07
- Run Application Server
- Log in as _SuperUser_
- Watch file _$ADEMPIERE_HOME/utils/nohup.err_

Result: the _jsp_ error reported disappeared

BUT:
1- a different error appeared in _$ADEMPIERE_HOME/utils/nohup.err_ :
 ` java.lang.IllegalArgumentException: Table Name Not Found - AD_QueueType`

Selection_196

By the way, this very error is also displayed when **installing with Tomcat**:
`===========> RequestProcessor.lambda$sendEmail$5: Table Name Not Found - AD_QueueType [22]`

2-  Moreover, when executing the process _Role Access Update_, the following error pops  up:
`org.postgresql.util.PSQLException: ERROR: current transaction is aborted, commands ignored until end of transaction block, SQL=SELECT IsActive,Updated,Created,IsDependentEntities,IsExclude,IsReadOnly,AD_Client_ID,AD_Org_ID,UpdatedBy,AD_Table_ID,Record_ID,AD_Role_ID,CreatedBy,UUID FROM AD_Record_Access WHERE (AD_Role_ID=?) AND IsActive=? ORDER BY AD_Table_ID
`imagen

I ignore if the first error causes this error, or if it is independent of it.

marcalwestf:
@e-Evolution Víctor on the one hand, the error you reported disappeared, so this pull request could be accepted.

On the other hand, a new error is generated at the same place, so if the error can be solved relatively easy, it could be done on top of this pull request.

Also, the second error could be managed the same way.

The other possibility would be to accept this PR and create two more issues.

You tell me what would be more suitable.
Thanls!!

e-Evolution:
@marcalwestf Hi Mario

Hello Mario!

Try this pull request from a new seed , from what I can see the errors it shows are because the database you are working on already has the refactoring migration XMLs applied, but the code does not match that seed.

If you generate the seed of the database with the XML Script of the pull request it should work

marcalwestf:
Ok, thanks @e-Evolution Víctor

marcalwestf (commented):
Took the 393 seed and tested again.
The same error is generated.

e-Evolution:
@marcalwestf Hi Mario , In my case I have no error, I put here the output of nohup.err and nohup.out

image

image

Apparently it has to do with how you do your deployment

@rcapecce Could you help me with this test please?

marcalwestf (approved):
Tested again, this time with a database with latest XML loaded and applied.
Login OK, no error in file _nohup.err_

Thanks, @e-Evolution  Victor for your support.

PR #3904, opened 2022-06-06
feat: Add selection columns in Conversion Rate.

EdwinBetanc0urt:
They are added as selection columns:
* Currency.
* Currency To.
* Conversion Type.
* Valid From.
* Valid To.
* Divide Rate.
* Multiply Rate.

conversion-rate-lookup

In order to be able to perform a simple search for records, since the option currently available is the advanced search and although it may be more accurate, it is not the most practical.

Before these changes:

https://user-images.githubusercontent.com/20288327/172081103-75e77171-b185-4e62-a70c-c7c4362fa79b.mp4

After these changes:

https://user-images.githubusercontent.com/20288327/172081100-a52a158b-1561-4542-8488-5a725f7c400a.mp4

Additionally, the `Currency To` field is marked as the same line, to have the same visual consistency of the `Conversion Rate` tab in the `Currency` window (left) as in the `Conversion Rate` window (right).

Before this changes:

conversion-rate-without-same-line

After this changes:

conversion-rate-same-line

### DIFF
--- a/migration/393lts-394lts/09700_ConversionRate_Selection_Columns.xml
+++ b/migration/393lts-394lts/09700_ConversionRate_Selection_Columns.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<Migrations>
+  <Migration EntityType="D" Name="Conversion Rate Selection Columns" ReleaseNo="3.9.4" SeqNo="9700">
+    <Comments>Add selection columns to easy search, and same line currency to field in currency window</Comments>
+    <Step SeqNo="10" StepType="AD">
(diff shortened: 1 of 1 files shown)


marcalwestf:
@EdwinBetanc0urt  thanks for your contribution; it seems a help, and not invasive.
When I am available, I will test it.

marcalwestf (approved):
- load and execute XML file _09700_ConversionRate_Selection_Columns.xml_
- check changes in table _C_Conversion_Rate_
- Login as _SuperUser_
- Check selection fields on window _Currency Rate_

Result: selection fields as expected.
Test passed.
imagen

PR #3905, opened 2022-06-06
feat: Add selection columns in smart browse.

EdwinBetanc0urt:
#### Description
In order to be able to perform a simple search for records, since the option currently available is the advanced search and although it may be more accurate, it is not the most practical.

#### Before this changes

without-lookup-record

#### After this changes

They are added as selection columns in the `Smart Browse` window:
* `Smart Browse` tab:
  * Search Key.
  * Name.
  * View.
  * Process.
  * Window.
  * Table.
 
smart-browser-with-lookup-record png

* `Smart Browse Field` tab:
  * Name.
  * Description.
  * View Column,
  * System Element.

smart-browser-field-with-lookup-record

#### Steps to reproduce

1. Login witth `System` role.
2. Open `Smart Browse` window.
3. Lookup record.

https://user-images.githubusercontent.com/20288327/172094957-238dd389-86e7-475b-9a6f-3db65639d306.mp4

### DIFF
--- a/migration/393lts-394lts/09710_Smart_Browse_Selection_Columns.xml
+++ b/migration/393lts-394lts/09710_Smart_Browse_Selection_Columns.xml
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<Migrations>
+  <Migration EntityType="D" Name="Smart Browse Selection Columns" ReleaseNo="3.9.4" SeqNo="9710">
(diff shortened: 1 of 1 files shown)


marcalwestf (approved):
- load and execute XML file _09710_Smart_Browse_Selection_Columns.xml_
- check changes on table _AD_Browse_
- check selection fields on window _Smart Browse_

Result: selection fields as expected.
Test passed

PR #3906, opened 2022-06-07
Disable NotificationRecipient on error

### LINKED ISSUE #3899: [Bug Report] A Request with a closed status does not stop sending email notifications
## Bug report
[Bug Report] A Request with a closed status does not stop sending email notifications

#### Steps to reproduce
1.- Create a new request as Due Type Schedule 
2.- When this request is due notice is triggered
3.- Change Status to 3_Closed
4.- The request not should   continue with trigger notices if was closed

#### Screenshot or Gif

image

The Notice does not  stop, continue to send email even when the request was closed

image

#### Expected behavior
If a request is closed no more notifications should arrive, this is the way it worked before the new notification handling was introduced

- Your OS: Macos
- Java version: 11
- ADempiere version: 3.4
- Browse safari
- DB Provider: PG
- DB version: 13

#### Additional context
Enabled default notification usage using new notification functionality

rcapecce:
new improvement described in **https://github.com/adempiere/adempiere/issues/3899#issuecomment-1148822450**
Fixes #3899

### DIFF
--- a/base/src/org/spin/queue/notification/support/EMailSender.java
+++ b/base/src/org/spin/queue/notification/support/EMailSender.java
@@ -16,7 +16,11 @@
  *****************************************************************************/
 package org.spin.queue.notification.support;
 
-import org.compiere.model.*;
+import org.compiere.model.MAttachment;
+import org.compiere.model.MClient;
+import org.compiere.model.MEMailConfig;
+import org.compiere.model.MUser;
+import org.compiere.model.X_AD_UserMail;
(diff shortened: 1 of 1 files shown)


yamelsenih:
Hi @rcapecce I can see that you remove the throw and is good because the intent to send was done.

For me is ok. Maybe Victor can add a comment based on your issue

e-Evolution:
The testing work, now only send the recipient that is active, but I found another problem that would be annoying #3916

PR #3909, opened 2022-06-09
Feature/#3776 send email notification with org email

### LINKED ISSUE #3776: [Feature Request] Send Email Notification with Organization Email
## Feature request

It is required to have the possibility of sending notifications by mail not only from a single company email, but also from different organizations.

### Describe the solution you'd like

For that, it's proposed to add to the Organization, the configuration of Email and password, just as the configuration is made in the Client.

In the code, it's proposed to determine if it corresponds to send the organization email or the company email, verifying if the record of the AD_Queue table has an organization established, and, in addition, the organization has the mail configuration parameterized.

#### Side: Generate Notification

The notification queue already has a method at the moment of defining the notification to establish by code with which organization to set the AD_Queue (most likely the organization of the PO in question)

#### Implementation Side of Send Mail.

- When the AD_NotificationQueue record is defined, it is already configured to take the AD_Client_ID and AD_Org_ID from the AD_Queue record.

##### EMailSender.sendNotification(MADNotificationQueue notification) Method

- Email Definition
- - The email of the Organization must be obtained if it is defined
- - Otherwise, obtain the email of the company as it has been done until now.
- Parameterization of the mail server
- - Should be changed to use Email class constructor overload, the current constructor uses this overload by not defining the mail server configuration, and then gets the company configuration. It has to be changed to get the one of the organization if it corresponds.

rcapecce:
Resolve #3776


base/src/org/compiere/process/EMailTest.java
```diff
@@ -18,8 +18,7 @@
 
 import java.io.File;
 
-import org.compiere.model.MClient;
-import org.compiere.model.MStore;
+import org.compiere.model.*;
```
e-Evolution, line 21: @rcapecce in ADempiere the best practice indicates that you define exactly the classes that are being imported and the use of .* to import the classes is not recommended, please change it

base/src/org/compiere/process/EMailTest.java
```diff
@@ -52,9 +51,19 @@ protected String doIt () throws Exception
...
+		int tableId = getTable_ID();
+		String clientTest = "";
+		if (tableId == I_AD_Client.Table_ID) {
+			clientTest = client.testEMail();
+		} else if (tableId == I_AD_OrgInfo.Table_ID) {
+			System.out.println("getRecord_ID() = " + getRecord_ID());
+
+//			MOrgInfo = new Query(getCtx(), I_AD_OrgInfo.Table_Name, I_AD_OrgInfo.COLUMNNAME_AD_Org_ID)
+//			MOrgInfo mOrgInfo = new MOrgInfo(getCtx(), getRecord_ID(), get_TrxName());
+//			String ret = EMail.validateMailDelivery(getCtx(), getRequestEMail(), getAD_EMailConfig_ID(), getName());
```
e-Evolution, line 65: @rcapecce If a code is not used, it is better not to add code even if they are comments, on the other hand, using System.out.println to report any message is not recommended, for this, it is better to use a log INFO type.

If your changing to this class doesn't add any real functionality, does it make sense to modify it?

base/src/org/compiere/process/OrgEMailTest.java
```diff
@@ -0,0 +1,44 @@
...
+	protected void prepare()
+	{
+		super.prepare();
+	}
+
+	@Override
+	protected String doIt() throws Exception
+	{
+		MOrgInfo orgInfo = MOrgInfo.get(getCtx(), getOrgId(), get_TrxName());
+		log.info(orgInfo.toString());
```
e-Evolution, line 40: @rcapecce There is no validation for orgInfo , what if it is null?

base/src/org/compiere/util/EMail.java
```diff
@@ -1483,4 +1483,37 @@ public static void main (String[] args)
...
+			return "No Request EMail for " + name;
+		//
+		EMail email = new EMail(MClient.get(ctx), eMailConfigId, requestEMail, requestEMail,
+				"Adempiere EMail Test",
+				"Adempiere EMail Test", false);
+		email.createAuthenticator (requestEMail, requestPassword);
+
+		try
+		{
+			String msg = email.send();
```
e-Evolution, line 1497: @rcapecce Please use natural language to name the variables that makes it easier to read and analyze the code, change msg to message

base/src/org/compiere/util/EMail.java
```diff
@@ -1483,4 +1483,37 @@ public static void main (String[] args)
...
+			if (EMail.SENT_OK.equals (msg))
+			{
+				log.info("Sent Test EMail to " + requestEMail);
+				return "OK";
+			}
+			else
+			{
+				MEMailConfig meMailConfig = new MEMailConfig(ctx, eMailConfigId, null);
```
e-Evolution, line 1505: @rcapecce Please use natural language to name the variables that makes it easier to read and analyze the code, change  meMailConfig to mailConfig

base/src/org/spin/queue/notification/support/EMailSender.java
```diff
@@ -63,18 +65,31 @@ public void sendNotification(MADNotificationQueue notification) {
 		notification.getRecipients().forEach(recipient -> {
 			MClient client = MClient.get(notification.getCtx(), notification.getAD_Client_ID());
 			String eMailFrom = client.getRequestEMail();
+			String eMailPsw = client.getRequestUserPW();
```
e-Evolution, line 68: @rcapecce Please use natural language to name the variables that makes it easier to read and analyze the code, change msg to message eMailPsw to requestUserPassword

Note the readability of the code is more important than saving a few characters, on the other hand IDEs like Intellij help with this job in a very easy way

base/src/org/spin/queue/notification/support/EMailSender.java
```diff
@@ -63,18 +65,31 @@ public void sendNotification(MADNotificationQueue notification) {
...
-					client,
+			EMail email = new EMail(client,
+					eMailConfigId,
 					eMailFrom,
 					recipient.getAccountName(),
 					notification.getDescription(),
-					notification.getText()
-			);
+					notification.getText(),
+					false);
```
e-Evolution, line 92: @rcapecce If the content is HTML or not it should not be forced I understand that this should come from the notification and the notification should be based on the Mail Text Template

e-Evolution (changes requested):

base/src/org/spin/queue/notification/support/EMailSender.java
rcapecce: In this line, we tried to respect what was already accepted before, so we tried to respect the same operation by incorporating the emailConfig method.
That is the reason why that parameter is burned, if you look at the signature of the method used previously, it is already burned from there.
https://github.com/adempiere/adempiere/blob/develop/base/src/org/compiere/util/EMail.java#L101
Currently the notifications do not see that they have support to determine if they are in html format or not.
In case you want to give support to this issue, how do you propose that the correction could be?

rcapecce:
https://github.com/adempiere/adempiere/blob/develop/base/src/org/spin/queue/notification/support/EMailSender.java#L81
@e-Evolution, in this line is it necessary to query if the company's email configuration is set if you want to go to the sending side by the user?
In the event that you want to send notification from the organization, from where should you obtain it?

I understand how the PR is currently regarding this, *complies with the stated requirement.*

In this line you get the company https://github.com/adempiere/adempiere/pull/3909/files#diff-cfb8f848b3d18533a4ec6b1ea5ba3b8d289e9b0c66a9f6ef4ce3160170c268bbR67

Here if there is mail configuration in the organization it overrides it, that would prevail.
https://github.com/adempiere/adempiere/pull/3909/files#diff-cfb8f848b3d18533a4ec6b1ea5ba3b8d289e9b0c66a9f6ef4ce3160170c268bbR71

And finally, if the mail is configured in the user, this will prevail
https://github.com/adempiere/adempiere/pull/3909/files#diff-cfb8f848b3d18533a4ec6b1ea5ba3b8d289e9b0c66a9f6ef4ce3160170c268bbR101

rcapecce:
As I see, the organization is not set in the notifications unless you call the QueueManager.withOrganizationId method or from the QueueManager.withEntry method explicitly, and from what I saw, the PO with the withEntry has never been passed.
That's why I added it here https://github.com/adempiere/adempiere/pull/3909/files#diff-99bfe6ba6bca51a15739c83a2ebf5accf741d61ebeb388652ca4e4550c76e2deR1168
@yamelsenih can you confirm that the organization in the queue is not set.
That's why @e-Evolution added it there.
I see two issues here.
1) The only thing I did in the EMailSender class was to add the organization thing where it is in the PR.
It can be seen as invalid that first it took the email of the company and then that of the User if it had, and I added the one of the organization to the middle, so _I take the one of the company, but if it has the one of the organization, but if it has the of the User_.
If they consider it necessary, it can be refactored by asking the other way around, first for the user, or for the organization and finally for the company.
2) The organization in the notifications is not set, @yamelsenih , how do you think you can fix it, because withEntry should not always be called to pass a PO, so you have to find an implicit way to always set it, do you think it is correct to obtain it from the context here? https://github.com/adempiere/adempiere/blob/develop/base/src/org/spin/queue/util/QueueManager.java#L209

base/src/org/spin/queue/notification/support/EMailSender.java
```diff
@@ -63,18 +65,31 @@ public void sendNotification(MADNotificationQueue notification) {
 		notification.getRecipients().forEach(recipient -> {
 			MClient client = MClient.get(notification.getCtx(), notification.getAD_Client_ID());
 			String eMailFrom = client.getRequestEMail();
+			String eMailPsw = client.getRequestUserPW();
+			int eMailConfigId = client.getAD_EMailConfig_ID();
+
+			if (notification.getAD_Org_ID() > 0) {
+				MOrg mOrg = MOrg.get(notification.getCtx(), notification.getAD_Org_ID());
```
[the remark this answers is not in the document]
e-Evolution: @rcapecce all notification always had organización to context , so this logic always is true

yamelsenih:
Hi, I think that the organization is good when is getted from entity (PO) or from `withOrganizationId` method.

After it just need make the follow:
- Validate the `AD_EMailConfig_ID` by user
- If the previous validation not have nothing then verify the `AD_EMailConfig_ID` by organization of notification
- Else just get it from `AD_EMailConfig_ID` of Client

base/src/org/compiere/process/EMailTest.java
rcapecce: ok @e-Evolution , I change this imports

base/src/org/compiere/process/EMailTest.java
rcapecce: @e-Evolution I delete this section of commented code

base/src/org/compiere/process/OrgEMailTest.java
rcapecce: @e-Evolution I will ask if this reference is null to execute this code block.

base/src/org/compiere/util/EMail.java
rcapecce: @e-Evolution I change this variable name

base/src/org/compiere/util/EMail.java
rcapecce: @e-Evolution ok

base/src/org/spin/queue/notification/support/EMailSender.java
rcapecce: @e-Evolution ok, I will use variables more readeable

base/src/org/spin/queue/notification/support/EMailSender.java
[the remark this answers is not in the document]
rcapecce: @e-Evolution ok the descriptible variable name

base/src/org/spin/queue/notification/support/EMailSender.java
```diff
@@ -63,18 +65,31 @@ public void sendNotification(MADNotificationQueue notification) {
...
+				MOrgInfo mOrgInfo = mOrg.getInfo();
```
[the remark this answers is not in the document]
rcapecce: @e-Evolution ok

rcapecce:
Hi @yamelsenih !

I spoke with @e-Evolution about this request, I'll tell you what we talked about.

Here I separate two parts:

First part
When the notification is created, we saw with Victor that it is necessary to call the "withEntity" method to pass the context of the object to be created.
We also reviewed the withTableId and withRecordId methods and we saw that it is not the best way for them to be separated, since one needs the other to identify a record, for that we define an overload of withEntity so that it calls the other two together.
Those fixes would already be in the branch.

Second part
On the side of sending the notification, in the EMailSender class, we define the following behavior, tell us what you think.
Respecting the logic of going from less to more, we define the following steps (class EMailSender.sendNotification):
* If the user has defined mail, password and mail configuration, I interpret that you want to send it there.
** I want to send with the user data, if it sends ok, if it does not send it will mark an error

* If the user does not have all these data, I continue with the organization, if they have email, password and email configuration defined
** I want to send with the data of the organization, if it sends ok, if it does not send it will mark an error
* If the organization does not have all these data, I continue with the company, if it has an email, password and email configuration
** I want to send with the company data, if it sends ok, if it does not send it marks an error

This part is pending development

base/src/org/spin/queue/util/QueueManager.java
```diff
@@ -179,7 +182,15 @@ public QueueManager withEntity(PO entity) {
 		}
 		return this;
 	}
-	
+
+	public QueueManager withEntity(int tableId, int recordId) {
+		if (getContext() == null || getTransactionName() == null) {
+			throw new AdempiereException("@not.found@");
```
e-Evolution, line 188: please change by AdempiereException("@NotFound@");

rcapecce: ok, I will change this value

base/src/org/spin/queue/util/QueueManager.java
```diff
@@ -136,7 +139,7 @@ public final QueueManager withContext(Properties context) {
 	/**
 	 * @param organizationId the tableId to set
 	 */
-	public final QueueManager withOrganizationId(int organizationId) {
+	private final QueueManager withOrganizationId(int organizationId) {
 		this.organizationId = organizationId;
```
rcapecce, line 143: Validate, only set the organizationId field if there is no established RecordID.
If a RecordID exists, the organization ID of the PO takes precedence over that of the method.

base/src/org/spin/queue/util/QueueManager.java
```diff
@@ -112,6 +114,7 @@ public final int getRecordId() {
 	/**
 	 * @param recordId the recordId to set
 	 */
+	@Deprecated
```
rcapecce, line 117: Refactor all references to delete this method

base/src/org/spin/queue/util/QueueManager.java
```diff
@@ -97,6 +98,7 @@ public final int getTableId() {
 	/**
 	 * @param tableId the tableId to set
 	 */
+	@Deprecated
```
rcapecce, line 101: Refactor all references to delete this method

rcapecce:
Talk to @yamelsenih and @e-Evolution and proceed to apply spoken suggestions in the code

e-Evolution:
@rcapecce can please check this class , it not is posible compile 

image

rcapecce:
> @rcapecce can please check this class , it not is posible compile
> 
> image

Hi @e-Evolution , I'm refactoring the EmailSender class with the changes that we commented, I'll check what you tell me tomorrow

Issue #3916, opened 2022-06-15
[Bug Report] Now it is not possible to deactivate the sending of notifications of a Scheduler deactivate the recipient and restart the Server


e-Evolution:
## Bug report
Now it is not possible to deactivate the sending of notifications of a Scheduler deactivate the recipient and restart the Server

#### Steps to reproduce

1.- Enable Flush System Queue
2.- Send an Email 
3.- Flush System Queue is executed each 5 minutes 
4.- For 5 minutes receive a Notice from  Flush System Queue Scheduler5. 
5.- Deactivate the Scheduler Recipient 
6.- And continue to send the notices of execution

#### Screenshot or Gif

![image](https://user-images.githubusercontent.com/786968/173939798-b82091fe-dc8d-4119-ad3d-45fbd2bb66ca.png)

#### Expected behavior
If you deactivate the Scheduler Recipient, then you not should send a notice

#### Other relevant information
- Your OS: MacOS
- Java version: 11
- ADempiere version:  394 soon to release 
- Browser safari
- DB Provider: PG
- DB version: 14

e-Evolution:
> @e-Evolution I dont understand the Bug that you define.
> 
> I think that the problem might be that for some reason, there is a Notification Queue created by each Process of Flush System Queue that is generated, and this notification try to send an emaill with the LOG of the Process
> 
> ![image](https://user-images.githubusercontent.com/31936815/174116035-5550c122-4b3b-4f46-8459-c551febd08b5.png)
> 
> ![image](https://user-images.githubusercontent.com/31936815/174118044-5f085e6c-5278-4452-9609-e29c8af37411.png)
> 
> The thing is that in our case, there is no account name defined in the AD_NotificationRecipient
> 
> ![image](https://user-images.githubusercontent.com/31936815/174117151-e0eeedec-e8c0-4931-8e9a-fd497b621019.png)
> 
> And in the scheduler Recipient we have it empty
> 
> ![image](https://user-images.githubusercontent.com/31936815/174117290-0d8dd7a0-f5c9-4199-85cf-68281fbd58e3.png)
> 
>  Notification.Queue.created.by.process.log.mp4 
> In order of this, i understand that the bug is that the process for some reason is creating a Notification System and a Notification Queue after every process and it shouldnt be like this. Unless you define a User in the "Scheduler Recipient" AD_SchedulerRecipient WHERE AD_User_ID?
> 
> What happens if you delete the record that you show in the AD_SchedulerRecipient ?

[...]

PR #3917, opened 2022-06-15
#3916 [Bug Report] Now it is not possible to deactivate the sending o…

e-Evolution:
…f notifications of a Scheduler deactivate the recipient and restart the Server https://github.com/adempiere/adempiere/issues/3916


niclopup (approved):
I understand that the Problem that this PR solve is only to "Stop Sending the notification to the recipient of the process when the record is Active=NO"

On the other hand, i thint this didn't solve the other problem of the notification sended to the SuperUser for every time the process run. For this I will create an other issue

PR #3923, opened 2022-06-20
GL Distribution doesn't work for any account

### LINKED ISSUE #3920: GL Distribution doesn't work for any account
## Feature request

GL Distributions Example 
image

If you define a different GL Distribution but only with other Dimension  in N and with the check of "Any Account" in Y, it doesn't work.

image

The reason of this is that right now it only look for the Account, and not for the other Dimensions in the Account Combination.

The code must consider a GL Distribution defined to "ANY ACCOUNT" and with other dimensions defined.

nsarlabos:
Fixes #3920

### DIFF
--- a/base/src/org/compiere/acct/Fact.java
+++ b/base/src/org/compiere/acct/Fact.java
@@ -686,6 +686,11 @@ public final class Fact
 			FactLine factLineSource = (FactLine)m_lines.get(i);
 			List<MDistribution> distributions = MDistribution.get (factLineSource.getAccount(),
 				m_postingType, m_doc.getC_DocType_ID(),factLineSource.getDateAcct());
+
+			if (distributions == null || distributions.size() == 0){
+				distributions = MDistribution.getDistributions(factLineSource.getCtx(), factLineSource.getC_AcctSchema_ID());
+			}
+
 			//	No Distribution for this line
 			//AZ Goodwill
 			//The above "get" only work in GL Journal because it's using ValidCombination Account
--- a/base/src/org/compiere/model/MDistribution.java
+++ b/base/src/org/compiere/model/MDistribution.java
@@ -92,7 +92,7 @@ public class MDistribution extends X_GL_Distribution
 			int orgTrxId, int salesRegionId, int locToId,
 			int locFromId, int user1Id, int user2Id, int user3Id, int user4Id,
 			Timestamp accountDate) {
-		List<MDistribution> distributions = getDistributions(ctx, acctSchemaId);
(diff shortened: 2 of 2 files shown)


nsarlabos:
Hi @e-Evolution, can you tell me about the change you talked about with Nicolás López?

e-Evolution:
@niclopup Nico , @SusanneCalderon Susanne 

please check this code line 

image

Can please check here the code , because should fix here not in the other class.

https://github.com/adempiere/adempiere/blob/758ed8bd791ed4c0d22c976beb1b2f2e49a76826/base/src/org/compiere/acct/Fact.java

e-Evolution:
@niclopup the two important testing :

1.- Create a distribution with account.
2.- Create a distribution with same account and activity to testing multiples dimensions
3.- Both distributions  should work.

nsarlabos:
Hi @e-Evolution 

I have tried to make the change in the "Fact.java", but I have the problem that it makes the distribution also for the accounts of the head (business partner of the head of the invoice), and the asset/liability accounts should not be distributed.

Please, I would be grateful if you could tell me exactly how the change to be made in the "Fact.java" would be, or if we see that perhaps an extra control should be added to avoid taking the accounts from the head of the document.

SusanneCalderon (approved):
@nsarlabos 
I tested to distribute Purchase Invoices to distribute to different cost centers (user1_ID) without problems.

nsarlabos:
> @nsarlabos I tested to distribute Purchase Invoices to distribute to different cost centers (user1_ID) without problems.

@SusanneCalderon With this change we had some errors when posting invoices without accounting distribution

marcalwestf:
@nsarlabos  can you tell us the status of this PR?

It is not completely clear if it works now, or if you found an error.
Thanks.

yamelsenih:
Hello @nsarlabos  can you update this pull request? we want update all pull request and is possible that this pull request will be closed if not exists a update

yamelsenih:
Hello @nsarlabos here a complete refactory for this: https://github.com/adempiere/adempiere/pull/3963

nsarlabos:
> Hello @nsarlabos here a complete refactory for this: #3963

Hi @yamelsenih, so #3963 would be the correct fix? If so we can test it.

In that case I understand that this pull request should be closed.

yamelsenih:
> > Hello @nsarlabos here a complete refactory for this: #3963
> 
> Hi @yamelsenih, so #3963 would be the correct fix? If so we can test it.
> 
> In that case I understand that this pull request should be closed.

Hi @nsarlabos off course, please test the PR and let us if is ok.

I will to close it

Issue #3932, opened 2022-06-28
Database is restored twice running Github Action "ADempiere Build" (file: "main.yaml")


marcalwestf:
## Bug report
**Seed Database** is restored **twice** running the Github Action _ADempiere Build_ (file: _main.yaml_).
This causes the action to last much longer than it should.

#### Steps to reproduce
1. Create a pull request
2. Wait a while until the action  `ADempiere Build` for this pull request is finished.
3. Go to  `Actions` and click on an executed `ADempiere Build` workflow
4. Click on the `create-seed` job
5. Expand the `Build ADempiere ....` section
6. According to action log file, around line **6873**, _Run Import Adempiere_ is executed. A couple of lines further, it sates `and import /home/runner/work/adempiere/adempiere/adempiere/Adempiere/data/Adempiere_pg.dmp`
The import ends around line **14400**:  `done`.
7.  According to action log file, around line **14400**, _Run DB Restore_ is executed `Run DB Restore Adempiere`. A couple of lines further down it says, `Re-Create Adempiere User and import /home/runner/work/adempiere/adempiere/adempiere/Adempiere/data/Adempiere_pg.dmp`.
The restore ends around line 21920:  `*** 2022-06-27 06:36:39.582 Adempiere Log (CLogConsole) ***`

#### Screenshot or Gif
Run Import Adempiere begin:
![imagen](https://user-images.githubusercontent.com/1789408/176066020-3893d110-657c-42df-aced-b986ce520c83.png)
Run Import Adempiere end and Run DB Restore begin:
![imagen](https://user-images.githubusercontent.com/1789408/176066185-45de1170-7310-4d66-b2d7-c7098bd84d02.png)
Run DB Restore end:
![imagen](https://user-images.githubusercontent.com/1789408/176067518-31fac26d-5e0e-495b-b791-79971a68295b.png)

#### Expected behavior
The database should be restored only once.

#### Other relevant information
- Your OS: Github-hosted runner
- Java version: 11
- ADempiere version: 394

#### Where the duplicate seed creation is defined
In _adempiere/build.xml_ -which is executed by ANT from main.yaml- both `Run Import Adempiere` and `Run DB Restore `are executed one after another:
![imagen](https://user-images.githubusercontent.com/1789408/176068138-1dedba46-d6f8-42f4-a306-7fa341d82df8.png)

#### How to avoid running twice the seed creation
Suggestion: by deleting either  `Run Import Adempiere` or `Run DB Restore ` .

[...]

PR #3935, opened 2022-06-30
Skip executing RUN_DBRestore to avoid importing seed twice

marcalwestf:
Fixes https://github.com/adempiere/adempiere/issues/3932

The test would consist of checking out the Action log of _ADempiere Build_ started by this pull request and verifying there that the database is imported only once.

Besides, all other steps (except the installation, which generates a **BUILD FAILED**) should run correctly and the artifacts are generated.


marcalwestf:
this is the result of the action:
imagen

e-Evolution (approved):

PR #3938, opened 2022-07-04
If there are no recipients, not send a notification to the createdBy user of the Scheduelr

### LINKED ISSUE #3924: Notification created of each process can't be deactivate
There is a Notification Queue created by each Process of Flush System Queue that is generated, and this notification try to send an emaill with the LOG of the Process.

image

image

In our case, there is no account name defined in the AD_NotificationRecipient

image

And in the scheduler Recipient we have it empty

image

https://user-images.githubusercontent.com/31936815/174117791-d200ef28-319c-439d-8a1b-bb90034be77c.mp4

In order of this, i understand that the bug is that the process for some reason is creating a Notification System and a Notification Queue after every process and it shouldnt be like this. 

There must be some way to define if you want to create this notification or not. If you want, i understand that it would be enough to define the Scheduler Recipient to recieve it, so i dont see the necesity to create an other notification every time the process run without the chance of define the User (it is created to de SuperUser and I don't know where it is defined).
 

Unless you define a User in the "Scheduler Recipient" AD_SchedulerRecipient WHERE AD_User_ID?

rcapecce:
Fix #3924

### DIFF
--- a/base/src/org/compiere/model/MScheduler.java
+++ b/base/src/org/compiere/model/MScheduler.java
@@ -37,6 +37,9 @@ import org.compiere.util.Msg;
  *  
  *  Contributors:
  *    Carlos Ruiz - globalqss - FR [3135351] - Enable Scheduler for buttons
+ *  @author Raul Capecce, raul.capecce@openupsolutions.com, OpenupSolutions https://openupsolutions.com
+ *			<a href="https://github.com/adempiere/adempiere/issues/3924">
+ * 			@see FR [ 3924 ] Notification created of each process can't be deactivate</a>
  */
 public class MScheduler extends X_AD_Scheduler
 	implements AdempiereProcessor, AdempiereProcessor2
@@ -223,11 +226,7 @@ public class MScheduler extends X_AD_Scheduler
 				}
 			}
 		}
-		//	Add Updater
-		if (list.size() == 0)
-		{
-			list.add(getCreatedBy());
-		}
+
 		//
 		return list.toArray(new Integer[list.size()]);
 	}	//	getRecipientAD_User_IDs


rcapecce:
Hi @yamelsenih and @e-Evolution , can you check this PR?

base/src/org/compiere/model/MScheduler.java
```diff
@@ -223,11 +226,7 @@ public Integer[] getRecipientByUserIds()
 				}
 			}
 		}
-		//	Add Updater
-		if (list.size() == 0)
-		{
-			list.add(getCreatedBy());
-		}
+
```
e-Evolution, line 229: @rcapecce Hi Raul!

I think the this code should be replace by :

`
		if (list.size() == 0)
		{
			list.add(getSupervisor_ID());
		}
`

Is very important that the Supervisor is noticed when a Scheduler is executed successfully and receipt so that one needs to know the process and not have some problems.

Even when this is a report the result of the report is sent to the supervisor as well.

We agreed that it was not necessary to notify the creator if the supervisor was notified and I think we all agreed

yamelsenih:
Hello @rcapecce can you update this pull request? we want update all pull request and is possible that this pull request will be closed if not exists a update

yamelsenih:
Hello @nsarlabos  can you update this pull request? we want update all pull request and is possible that this pull request will be closed if not exists a update

yamelsenih:

niclopup:
As i said in the following comment i think the issue is complete and the thing that is asked to change is not includded in the scope of this issue

https://github.com/adempiere/adempiere/pull/3938#discussion_r1380102759

base/src/org/compiere/model/MScheduler.java
yamelsenih: Hello @rcapecce @niclopup I think that you shoul add this code instead:
```Java
if (list.size() == 0) {
  list.add(getSupervisor_ID());
}
```

yamelsenih:
@rcapecce can you change this?

Just we need that you change this pull request according to:
```Java
if (list.size() == 0) {
  list.add(getSupervisor_ID());
}
```

We will review this pull request the next **2023-11-21** If not exists some changes then we will close this.

rcapecce:
I open a new PR #4198 for this change in other repo because this repo is archived

Issue #3939, opened 2022-07-04
Github Build Action: Restore Database Only When XML Files Involved


marcalwestf:
## Feature request
Right now, the Github Action _ADempiere Build_ defined in file _main.yml_, starts the build process:
![imagen](https://user-images.githubusercontent.com/1789408/177221847-cb9e0ccf-619d-432f-8c33-6c89f479d957.png)

The _build.xml_ file located directly at the root of the adempiere repository has a target called _build,_ which is executed by the `ant build` call.

The following is then executed:
- Building ADempiere
- Deploy Adempiere (Preparation for setup)
- SilentSetup
- Run Import Adempiere
- Run DB Restore Adempiere (this ought to be skipped by accepting PR https://github.com/adempiere/adempiere/pull/3935)
- Run Migrate XML
- Run Export Database

This means that whatever happens **the database is always restored**, including executing all XMLs starting from seed.
This solution has two disadvantages: it takes a long time for the action to finish and also it consumes time and resources Github allocates to our project.

#### Describe the solution you'd like
The first step would be to separate at least the building and installation of Adempiere from the database import, restore, and export using the present tools (ANT + build.xml files).

#### Describe alternatives you've considered
The best solution would be to implement everything in _Gradle_ but it would take too long. Modifying the present solution may be less complex/complicated.
It would also be less invasive and easier to test.
The solution must be also backward-compatible, i.e. it must work as it does today.

#### Additional context
Here are some implementation thoughts, in steps.
The implementation may differ due to syntax and other reasons, but the main idea should be conveyed:

1.- Divide file _adempiere/build.xml_ into several targets
```
<target name="build" depends="xxxx, yyyy, zzzz", ...> 
        <property name="dbrestore" value="true"/>

```

The additional targets (in the example xxxx, yyyy, zzzz) would be 
- Building ADempiere
- Deploy Adempiere
- SilentSetup
- Run Import Adempiere
- Run DB Restore Adempiere
- Run Migrate XML
- Run Export Database

2.- Still in _adempiere/build.xml_, make all database targets dependent on a property, for example _dbrestore_:
```
<target name="run-import-adempiere" description="Run Import Adempiere"
    if="${dbrestore.set}">
    :
    :
</target>

```

3.- Modifications in Github Action file _main.yaml_
The following calls must be possible:
- ant build
- ant -Ddbrestore.set=true build
- ant -Ddbrestore.set=false build
    
[...]

PR #3942, opened 2022-07-06
Github Build Action: Restore Database Only When XML Files Involved

marcalwestf:
Fixes: https://github.com/adempiere/adempiere/issues/3939

With this pull request, the actions involved in a push or pull request were improved: 
- there is always a build of the ADempiere repository and a corresponding artifact created
- there is always an installation, which is done with Tomcat, version 9 (eliminating a build error message)
- there is a database seed creation only if XML files within the `migration` folder have been changed
- a (new) database seed artifact is only created when XML files were modified
- the `build.xml` file was divided into more manageable sections which are sequentially called
- the actions are more verbose so they can be followed and debugged better
- one internal action under `.github/actions/adempiere-build` has been added, avoiding code cluttering in the `main.yml` file
- I have left legacy code as a comment in case somebody wants to know what are the new changes. They can be deleted after it was tested several times.

The tests should be quite easy: 
- either copy the commit into a branch of the private repository and do some commits and pushes, or
- accept this PR and watch in THIS repository what happens when a PR is created and/or merged: it should run as expected.

The action log should look as follows **when no XML file was modified**, and also **no database artifact should be created**:
imagen

Whereas the action log should look as follows **when at least one XML file was modified**, and also **a database artifact should be created**:
imagen

I've done the corresponding test on my own repository and every scenario has been tested; nevertheless, there can always be an unexpected situation.

### DIFF
--- a/.github/actions/adempiere-build/action.yml
+++ b/.github/actions/adempiere-build/action.yml
@@ -1,0 +1,37 @@
+name: 'ADempiere Build'
+description: 'ADempiere is built and installed. Database seed is generated only if a nigration xml file was modified'
+author: 'Mario Calderon, marcalwestf'
+inputs:
+  nodbrestore:  # id of input
+    description: 'true-> no restore of DB; false-> DB will be restored'
+    required: true
+    default: "true"
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        echo $ACTION_BUILD_DIR
+        echo $GITHUB_REF
+        echo $GITHUB_HEAD_REF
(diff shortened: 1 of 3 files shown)


marcalwestf:
The result of THIS pull request test:
imagen

It seems to work without XML files changed.
The time was reduced considerably.

BTW: please do accept first https://github.com/adempiere/adempiere/pull/3935 and https://github.com/adempiere/adempiere/pull/3936 before accepting this one.

e-Evolution (approved):

PR #3943, opened 2022-07-08
Rename Asset Package

yamelsenih:
This pull request basically change the package name of Asset to new structure:

Old structure:
```Java
  - org.compiere.model
  - org.compiere.process
```

New structure:
```Java
  - org.compiere.FA.model
  - org.compiere.FA.model.validator
  - org.compiere.FA.process
```

Here some checks:
## Processes
```SQL
SELECT p.Value, p.Name
FROM AD_Process p
WHERE p.ClassName IN('org.compiere.process.AssetDelivery');
```
 
 Value|Name
-- | --
 Asset_Delivery_Selection | Deliver Assets

## Model Validators
```SQL
SELECT mv.Name, mv.Description
FROM AD_ModelValidator mv
WHERE MV.ModelValidationClass IN('org.compiere.FA.model.ModelValidator');
```

Name|Description
--|--
 Model Validator to Fixed Assets | Fixed Assets

## Forms
```SQL
SELECT f.Name, f.Description
FROM AD_Form f
WHERE f.ClassName IN('');
```

Name|Description
--|--

## Final comments

The support legacy api is from here: https://github.com/adempiere/adempiere-asset-support-legacy


marcalwestf:
This PR is based on the refactoring of fleet package, which is based on the refactoring of WMS.
Thus, both refactoring PRs have to be merged before merging this one; otherwise, they would become obsolete when merging this PR.

marcalwestf:
@yamelsenih one question: in some classes, I noticed that the import is doing as
`import org.compiere.model.*;`
like in class **org.compiere.FA.model.I_A_Asset**:

imagen

This package import can be avoided by explicitly importing each class:
imagen

If I remember well, we use to import each class separately.
Consulting the forums (Stack Overflow et. al.) the recommendation is for different reasons to import every class, instead of *.

The classes involved are among others  I_A_Asset, I_A_Asset_Acct, I_A_Asset_Addition, I_A_Asset_Change, I_A_Asset_Class, I_A_Asset_Delivery, I_A_Asset_Disposed, I_A_Asset_Group etc. (actually, all I_* classes, including I_A_Asset* and I_A_Depreciation* and some X classes).

An explicit import has been done for example in X_A_Asset, so it would be symmetrical to do it also in the "I" and "X" classes.

It seems that importing explicitly involves many files. One file is done rapidly, but it adds up in the end.
My question is if it would be possible to do the imports explicitly.

marcalwestf:
... also the text in the header 

 ```
* For the text or an alternative of this public license, you may reach us    *
 * or via info@adempiere.net                                                  *
 * or https://github.com/adempiere/adempiere/blob/develop/license.html        *
 *****************************************************************************/
```
imagen

Should eliminate the first "or" in the second line:

 ```
* For the text or an alternative of this public license, you may reach us    *
 * via info@adempiere.net                                                     *
 * or https://github.com/adempiere/adempiere/blob/develop/license.html        *
 *****************************************************************************/
```

marcalwestf (changes requested):
I get an exception when saving an Asset:
imagen

In `DB.executeUpdate()` 
imagen

In  `MAsset.afterSave()`:
imagen

The error can be reproduced in the database:
imagen

The only difference is in `MAssetAcct()`:
imagen

marcalwestf:
When executing without the changes of this pull request, Assets can be created on the same database, prior to deactivating `Model Validator to Fixed Assets`.

marcalwestf:
@yamelsenih please disregard the question of changing the headers. Only the error remains.

I noticed that the headers and imports are generated automatically: 
imagen

It is then useless to change the files manually because they would be changed with the next generation.
The change should be done in `ModelInterfaceGenerator`:
imagen

I will create an issue this regarding so new generated class files have a correct header.

yamelsenih:
> I get an exception when saving an Asset: imagen
> 
> In `DB.executeUpdate()` imagen
> 
> In `MAsset.afterSave()`: imagen
> 
> The error can be reproduced in the database: imagen
> 
> The only difference is in `MAssetAcct()`: imagen

Thanks @marcalwestf I will fix it now

yamelsenih:
Hi I create other pull request here #3952

Issue #3945, opened 2022-07-10
[Bug Report] The cities corresponding to a Country/Region are not listed correctly


EdwinBetanc0urt:
#### Steps to reproduce
1. Open the `Country Region and City` window.
2. Select any country other than `United States`.
3. Check that it handles regions.
4. Create 2 regions.
5. Create cities for both regions.
6. Create a city without a region.

##### A. `Country Region and City` window:
Currently it does not have a filter, it lists all cities in the country even if it manages regions and a region is selected.

If a region is selected, you must list the cities that belong to that region and those cities registered without a region, in case the country does not manage regions you must list all cities.

https://user-images.githubusercontent.com/20288327/178133004-010070d6-0141-4fe8-a24e-ea0468f4aa59.mp4

##### B. `Location Address` window: 

When listing the cities it should show only the cities that belong to that country, however it shows all cities in all countries. It should list all the cities of that country, in case of selecting a region it should list the cities belonging to that region or the cities that do not have a region and are from the same country.

https://user-images.githubusercontent.com/20288327/178133010-10a1625b-0211-4f6b-8cc8-e99eca4da1f8.mp4

##### C. `Initial Client Setup Process`.

When listing cities without selecting a region it shows only cities in that country that do not have a region, and should list all cities regardless of region. However when selecting a region it should list the cities belonging to that region and those without a region in the same country, but omits those without a region.

https://user-images.githubusercontent.com/20288327/178133037-8de541a7-05ef-4893-af9c-8d94520eb6ac.mp4

#### Other relevant information
- Your OS: Kubuntu 20.04
- Java version: OpenJDK Temurin-11.0.15+10
- ADempiere version: Develop branch.
- Browser [e.g. chrome, safari]: Mozilla Firefox 
- Browser Version [e.g. 22]: 101.0
- DB Provider: PostgreSQL
- DB version: 14.2

PR #3946, opened 2022-07-10
fix: Add filters to list cities.

EdwinBetanc0urt:
##### A. `Country Region and City` window:

https://user-images.githubusercontent.com/20288327/178134350-876d6ec6-f2f2-4f7d-8272-10e40ef27cae.mp4

##### B. `Location Address` window: 

https://user-images.githubusercontent.com/20288327/178134348-931b4281-9e7f-4226-8461-bd2e94c0c12f.mp4

##### C. `Initial Client Setup Process`.

https://user-images.githubusercontent.com/20288327/178134379-0aabc73a-73cc-47a5-a093-94f054961803.mp4

#### Additional context:

The changes added in the GridField class are due to the fact that when a field value corresponding to an ID is changed to empty, it is removed from the context, and when doing SQL parsing it generates the following error since it does not get a single value (C_Region_ID) it returns an empty string omitting the entire sql to be parsed. Now when setting the value as empty it is set to 0 in the context.

```log
===========> VLookup.getWhereClause: C_City_ID - Cannot Parse=C_City.C_Country_ID=@C_Country_ID@ AND (@C_Region_ID@=0 OR (@C_Region_ID@>0 AND C_City.C_Region_ID=@C_Region_ID@ OR C_City.C_Region_ID IS NULL)) [17]
===========> VLookup.getWhereClause: C_City_ID - Cannot Parse=C_City.C_Country_ID=@C_Country_ID@ AND (@C_Region_ID@=0 OR (@C_Region_ID@>0 AND C_City.C_Region_ID=@C_Region_ID@ OR C_City.C_Region_ID IS NULL)) [17]
===========> InfoGeneral.<init>: Cannot parse context= C_City.C_Country_ID=@C_Country_ID@ AND (@C_Region_ID@=0 OR (@C_Region_ID@>0 AND C_City.C_Region_ID=@C_Region_ID@ OR C_City.C_Region_ID IS NULL)) [17]
===========> InfoGeneral.testCount: SELECT COUNT(*) FROM C_City WHERE IsActive='Y' AND  AND C_City.AD_Client_ID=0 AND C_City.AD_Org_ID=0 AND (C_City.C_City_ID IS NULL OR C_City.C_City_ID NOT IN ( SELECT Record_ID FROM AD_Private_Access WHERE AD_Table_ID = 186 AND AD_User_ID <> 100 AND IsActive = 'Y' )) [17]
org.postgresql.util.PSQLException: ERROR: syntax error at or near "AND"
```

fixes: https://github.com/adempiere/adempiere/issues/3945


marcalwestf:
@EdwinBetanc0urt  one question Edwin, I see you changed the grid at `GridField.updateContext()`:
imagen

Being an important part of the system, I'd like to verify that other parts of the program go through the code snippet and work well, not only the locations.

As I understand from the code, it is executed when there is NOT AN UPDATE from Memo, TextLong, Binary, RowID, Boolean, and timestamp. This would leave only strings and integers, right? 

My question is: what do I have to do in order to check these situations, besides the I have to test concerning the locations?
Thank you!

marcalwestf (changes requested):
### Apply changes
Compiler test, Apply XML, build -> OK
### Test I: positive test
Create a city for a region -> OK.
Only issue (not a big thing though)> the region must be extra selected -> OK.
### Test II: do some manipulation
When creating a city for a region, the region field is not adopted in the city record. The city can be saved either without region selected, or with a region selected:
imagen

If the column `C_Region_ID` of the tab City gets the default value `@C_Region_ID@`:
imagen

Then, the region of the parent tab is inherited and the user must not find and select it, and the city can be created easily:
imagen

BUT, if I try to create a city **without** region  (i.e. leaving the region empty), I get an `StackOverflowError` as soon as the region is left empty:
Selection_262

If I try nevertheless to save the record, I get the following error. Weird is the fact that without the default value set to  `@C_Region_ID@`, it is possible to create a city without region :
imagen


Changing the default value is a legit operation in the Application Dictionary and should not lead to an exception.

marcalwestf (commented):
### Test of locations
The locations can be created without errors.
In particular, it works by first selecting the country, then the region, and at last the city, with the correct filters:
imagen

marcalwestf (commented):
### Test of Initial Client Setup Process
Here the selection of first country, then region and at last city works well.
imagen

EdwinBetanc0urt:
Hi @marcalwestf, I added the default logic of the region field in the city tab, I also changed the display logic of the region tab to a read-only logic, because when selecting a country with region and then another country without region, it did not update correctly the list of cities.


https://user-images.githubusercontent.com/20288327/204270882-10590caf-9e6d-4239-af1e-5e22c2389b94.mp4

marcalwestf:
@EdwinBetanc0urt  hi Edwin, though it didn't make it to 3.9.4, this is an important feature which is worth integrating.
Could you please merge your pull request with the newest develop branch so it can be tested anew?


Thanks!

EdwinBetanc0urt:
Hi @marcalwestf Mario, thanks for your interest in this change, I already mixed the latest that is in the develop branch to this branch

I look forward to any comments or suggestions!

marcalwestf (changes requested):
Testing anew...
1.- Load + Apply Migration File  -> OK

2.- Select country _Venezuela_ and activate field _Country has Region_  -> OK
imagen

3.- Go to _Region_ tab and try to create a new record   -> **ERROR**
The displayed **Country** is _United States_ instead of _Venezuela_.
imagen

4.- In the console, the following error appears:
`-----------> GridTable.createSelectSql: Failed to parse where clause. whereClause=(('@HasRegion@'='Y' AND C_City.C_Region_ID=@C_Region_ID@ OR C_City.C_Region_ID IS NULL) OR ('@HasRegion@'='N')) AND C_City.C_Country_ID=173 [19]
`

5.- If nevertheless a region is created, it is saved as a region from the United Sates:
imagen

EdwinBetanc0urt:
The current problem is that when changing to a country that does not handle regions, the context of the `Region` tab is cleared, and going to the `City` tab generates the error trying to get the default value.

```log
-----------> GridTable.createSelectSql: Failed to parse where clause. whereClause=(('@HasRegion@'='Y' AND C_City.C_Region_ID=@C_Region_ID@ OR C_City.C_Region_ID IS NULL) OR ('@HasRegion@'='N')) AND C_City.C_Country_ID=116 [21]

```


Additionally, when saving a record, it is not seen in the grid and in the total records it says -1 as if they did not exist, this also causes an error to be generated when trying to activate the search engine in the `City` tab.

```log
===========> DB_PostgreSQL.convertStatement: DB_PostgreSQL.convertStatement - Not Converted () - SQL_Statement is null or has zero length [21]
```

However, if you change to a country that does not handle regions, and while in the `Region` tab you hit the button again and then undo (setting the default values in the context including C_Region_ID=0), the tab of `City` operates correctly.


Put this change in draft while I figure out how to solve it, anyway any help to solve it is welcome.

yamelsenih:
Hello @EdwinBetanc0urt  I will to close this pull request. Please if you have any change then a new pull request is welcome.

Issue #3957, opened 2022-07-26
[Bug Report] Update build ant files with last library javax.servlet-api-4.0.1.jar


e-Evolution:
## Bug report
[Bug Report] Update build ant files with last library javax.servlet-api-4.0.1.jar

#### Steps to reproduce
Currently, use the obsolete library 

#### Other relevant information
- Your OS: MacOS
- Java version: 17 
- ADempiere version: build nightly 3.9.4
- DB Provider: PG
- DB version: 14

PR #3958, opened 2022-07-26
#3957 [Bug Report] Update build ant files with last library javax.ser…

e-Evolution:
…vlet-api-4.0.1.jar  https://github.com/adempiere/adempiere/issues/3957


marcalwestf (approved):
- checked changes: seems that library `javax.servlet-api-4.0.1.jar` has been there for a while, but it was never used
- no compiler errors
- build run: successful
- no tests were executed where the library classes were instantiated on an installation that uses a servlet or web container.

Test passed

PR #3961, opened 2022-07-27
Fix #3956

### LINKED ISSUE #3956: [Bug Report] Set Organization Email to send notifications with this email
## Bug report

When you send a notification from the organization email, the notification can't asign the organization id, and then, the notification is sent with the Client email.

#### Steps to reproduce

* Configure the organization email en OrgInfo (email, password, email configuration).
* Create a new Request in the window (All Request) with a user who does not have his email configured for sending and with an organization that does have it.
* Run the Flush System Queue process to flush the pending notifications to send.
* The notification will not be sent from the organization email, but will be sent from the Client email.

#### Expected behavior

Take the email from the Organization to send this notification when the user that create the notification has no configuration to send email and the organization does.

rcapecce:
Fixing set organization order
Fix #3956

### DIFF
--- a/base/src/org/spin/queue/util/QueueManager.java
+++ b/base/src/org/spin/queue/util/QueueManager.java
@@ -174,9 +174,9 @@ public abstract class QueueManager {
 		if(entity != null) {
 			withContext(entity.getCtx())
 				.withTransactionName(entity.get_TrxName())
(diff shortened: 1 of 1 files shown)


e-Evolution (approved):
@rcapecce ok , the change is valid

PR #3963, opened 2022-08-01
#3920 GL Distribution doesn't work for any account  https://github.co…

e-Evolution:
…m/adempiere/adempiere/issues/3920

- The changes allow to generate a distribution for any dimension and solve the problem with the dimensions user 3 and user 4

### DIFF
--- a/base/src/org/compiere/acct/Fact.java
+++ b/base/src/org/compiere/acct/Fact.java
@@ -684,19 +684,9 @@ public final class Fact
 		for (int i = 0; i < m_lines.size(); i++)
 		{
 			FactLine factLineSource = (FactLine)m_lines.get(i);
-			List<MDistribution> distributions = MDistribution.get (factLineSource.getAccount(),
-				m_postingType, m_doc.getC_DocType_ID(),factLineSource.getDateAcct());
-			//	No Distribution for this line
-			//AZ Goodwill
-			//The above "get" only work in GL Journal because it's using ValidCombination Account
-			//Old:
-			//if (distributions == null || distributions.length == 0)
-			//	continue;
-			//For other document, we try the followings (from FactLine):
-			//New:	
-			if (distributions == null || distributions.size() == 0)
-			{
-				distributions = MDistribution.get (factLineSource.getCtx(), factLineSource.getC_AcctSchema_ID(),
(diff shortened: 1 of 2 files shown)


niclopup (changes requested):
I found an error with the following definition:
Distribución con PR de Victor
Error Actividad en Cabezal

But when the Activity is defined in the lines of a AP the distribution is made correctly

niclopup:
The lineas of the Distribution i am testing.

Distribución Actividad 2

linea 2 diist actividad
linea diist actividad


This distribution, if the Activity is defined in the lines, it make the distribution OK
DxP Posted OK Actiivdad en lineas
Linea de DxP Posted OK Actiivdad en lineas

Asiento de DxP Posted OK Actiivdad en lineas

marcalwestf:
one question: does this pull request replace https://github.com/adempiere/adempiere/pull/3923 ?

SusanneCalderon (changes requested):
@e-Evolution , @niclopup 
I could reproduce the error @niclopup mentioned. The error does not occure when the distribution does not generate a reversal.

yamelsenih:
Hello @e-Evolution  can you update this pull request? we want update all pull request and is possible that this pull request will be closed if not exists a update

yamelsenih:

e-Evolution:
@yamelsenih Hello, I think that the patch does apply since you should not have two methods to bring the distribution rules, these should apply to all dimensions, so bringing an accounting distribution only for the account can cause inconsistencies, so the change is Valid, I understand that Nicolas @niclopup found a problem, but I think it is not related to this patch, since what he reports is a different problem based on where the activity takes, whether from the header or the line, this problem happening now with or without this patch so unrelated

It would be good if you validated it and gave your peer review, I consider that the patch does apply, and the error that Nicolas found should be reported and fixed but it has nothing to do with this change.

yamelsenih:
I will review this with @carlosaparadam

yamelsenih:
Hello, I test it and is ok. 

The problem reported by @marcalwestf was a wrong setup for distribution. This behavior is when a distribution line have a percent in zero

yamelsenih:
@carlosaparadam please check it

yamelsenih (approved):
This pull request is ok for me

PR #3969, opened 2022-08-16
Fix #3968 - POInfo cached sql sintax error

### LINKED ISSUE #3968: [Bug Report] POInfo cached sql sintax error
## Bug report

When you going to save an Element, you save and it throw an error in the POInfo, this is because it's reusing a cached sql that call a select alias that not exist in the FROM section built before.
This happens when you select other language in login section like Spanish

#### Steps to reproduce

1. In a develop branch launch a Swing ADempiere
2. Login with SuperUser, System rol and English language
3. Try to save an Element
4. Row saved success
5. Stop the application
6. Login with SuperUser, System rol and Spanish language
7. Try to save an Element
8. You will get a message like _Invalid PO Info_

#### Screenshot or Gif

When you login with English language
Peek 16-08-2022 10-56

When you login with Spanish language
Peek 16-08-2022 10-58

- Your OS: Xubutu 20.04
- Java version: OpenJDK 11
- ADempiere version: 393
- DB Provider: PostgreSQL
- DB version: 12

rcapecce:
In the same execution, in some cases it takes the base language and in others it doesn't, and since the sql is cached, it throws an sql syntax error because there is a SELECT reference to the cached translation table that in the FROM doesn't this.

Since the getPOInfoColumnList method is called only from one place and since it does not use cache from there, for consistency this line is left uncached.

Fix #3968

### DIFF
--- a/base/src/org/compiere/model/POInfo.java
+++ b/base/src/org/compiere/model/POInfo.java
@@ -17,6 +17,7 @@
 package org.compiere.model;
 
 import io.vavr.collection.List;
+import org.adempiere.exceptions.AdempiereException;
 import org.compiere.util.CCache;
 import org.compiere.util.CLogger;
 import org.compiere.util.DB;
@@ -53,6 +54,8 @@ import java.util.logging.Level;
  *  @author Michael McKay, mckayERP@gmail.com
  *  		<li><A href="https://github.com/adempiere/adempiere/issues/2428">#2428 Allow changes to AD_Column and AD_Table</a>
  *  		Uses explicit references to column names rather than '*'
+ *  @author Raul Capecce, Openup Solutions http://openupsolutions.com/
+ *  		<li><a href="https://github.com/adempiere/adempiere/issues/3968">#3968 [Bug Report] POInfo cached sql sintax error</a></li>
  */
 public class POInfo implements Serializable
 {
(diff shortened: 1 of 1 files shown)


base/src/org/compiere/model/POInfo.java
```diff
@@ -245,6 +248,8 @@ private void loadInfo (boolean baseLanguage, String trxName)
 								trxName);
 					}).toList();
 					list.addAll(infoColumns.asJava());
+				}).onFailure(throwable -> {
+					throw new AdempiereException(throwable);
```
rcapecce, line 252: Hi @e-Evolution, I think in this section we need an error handler to show if there is any error like the syntax error that was caught here but then didn't log the error or throw it back

marcalwestf (commented):
### Error Reproducible
- Use repository **without** this PR's modifications
- Create an Element with login language English -> OK
- Create an Element with login language Spanish -> Error

### Solution Reproducible
- Use repository **with** this PR's modifications
- Create an Element with login language English -> OK
- Create an Element with login language Spanish -> OK

Result: modifications avoid the Exception.

I don't accept yet this PR, because I did some debugging (=setting a breakpoint and observing) and call hierarchy analysis of the method `POInfo.getPOInfoColumnList()` and I noticed that this method is called very often, for example when saving a record.
imagen

I wonder if removing the cache will lead to a noticeable speed degradation because the method `POInfo.getPOInfoColumnList()` is heavily used, and without cache, the system would always go to the database and get the list. 

If this should be the case, the solution would be to change the method instead of getting rid of the cache.

IMHO, if the speed is degraded too much, I would recommend for cases like the present to rather work with language English when at the AD than to make the system slower.

Perhaps @mckayERP  , @e-Evolution  and @yamelsenih can post their opinions here.

e-Evolution (approved):

PR #3966, opened 2022-08-07
Rename all packages

yamelsenih:
Branch: feature/#Src-Change-Packages

Rename base Package

This pull request basically change the package name of all `I_` and `X_` classes to new structure:

Old structure:
```Java
  - it.cnr.imaa.essi.lablib.gui.checkboxtree
  - org.compiere.acct
  - org.eevolution.engine.forecast
  - org.eevolution.exceptions
  - org.eevolution.model
  - org.eevolution.model.impl
  - org.eevolution.model.reasoner
  - org.eevolution.model.wrapper
  - org.eevolution.msg
  - org.eevolution.process
  - org.eevolution.report
  - org.eevolution.tools
  - org.eevolution.tools.worker
```

New structure:
```Java
  - it.cnr.imaa.essi.lablib.gui.checkboxtree
  - org.eevolution.manufacturing.acct
  - org.eevolution.manufacturing.controller
  - org.eevolution.manufacturing.engine.forecast
  - org.eevolution.manufacturing.exceptions
  - org.eevolution.manufacturing.model
  - org.eevolution.manufacturing.model.impl
  - org.eevolution.manufacturing.model.reasoner
  - org.eevolution.manufacturing.model.validator
  - org.eevolution.manufacturing.model.wrapper
  - org.eevolution.manufacturing.msg
  - org.eevolution.manufacturing.process
  - org.eevolution.manufacturing.report
  - org.eevolution.manufacturing.tools
  - org.eevolution.manufacturing.tools.worker
  - org.eevolution.manufacturing.utils
  - org.eevolution.model
```

Here some checks:
## Processes
```SQL
SELECT p.Value, p.Name
FROM AD_Process p
[...]

PR #3972, opened 2022-08-16
Restore Database Only When XML Files Involved (take # 2)

marcalwestf:
Fixes: https://github.com/adempiere/adempiere/issues/3939
This PR is the very same as https://github.com/adempiere/adempiere/pull/3942, which was reverted because it was accepted much later than created, and when merged the PR https://github.com/adempiere/adempiere/pull/3966 would have had to change too many files that it was better reverted.

The following is valid while it hasn't been done a complete migration to Gradle.

### Original PR Text
With this pull request, the actions involved in a push or pull request were improved: 
- there is always a build of the ADempiere repository and a corresponding artifact created
- there is always an installation, which is done with Tomcat, version 9 (eliminating a build error message)
- there is a database seed creation only if XML files within the `migration` folder have been changed
- a (new) database seed artifact is only created when XML files were modified
- the `build.xml` file was divided into more manageable sections which are sequentially called
- the actions are more verbose so they can be followed and debugged better
- one internal action under `.github/actions/adempiere-build` has been added, avoiding code cluttering in the `main.yml` file
- I have left legacy code as a comment in case somebody wants to know what are the new changes. They can be deleted after it was tested several times.

The tests should be quite easy: 
- either copy the commit into a branch of the private repository and do some commits and pushes, or
- accept this PR and watch in THIS repository what happens when a PR is created and/or merged: it should run as expected.

The action log should look as follows **when no XML file was modified**, and also **no database artifact should be created**:
imagen

Whereas the action log should look as follows **when at least one XML file was modified**, and also **a database artifact should be created**:
imagen

I've done the corresponding test on my own repository and every scenario has been tested; nevertheless, there can always be an unexpected situation.


marcalwestf:
The build was completed without errors.
Due to the fact that the action found no migration/*/*.xml files changed, it didn't create a seed (step _Upload Seed_ skipped) and the check lasted 7 minutes instead of 23:
imagen

EdwinBetanc0urt:
Greetings @marcalwestf this is a good change to improve the compile time of the actions, I think that maybe from the migration xml files, you should also include in the path if there is any change in the actions `.yml` files that are involved in the creation/alteration of the database, also in the ant `build.xml` or the `.sh`/`.bat` files for migrations or database restore.

It may not make a difference in an xml migration, but it may make a difference in how the database is generated from those files.

Maybe this can serve you as an example, we use it to generate the documentation only if there are changes related to the documentation path or in the action in charge of generating the documentation.

https://github.com/solop-develop/frontend-core/blob/a29677188be27fe93091e3d405ac2c6bd0ff0b4f/.github/workflows/docs.yml#L16-L19

marcalwestf:
@EdwinBetanc0urt  thanks for your suggestion, I will consider your hints and will work on them.
When (if) this pull request is accepted, I will do changes as you suggest and open a new pull request.

Best regards.

e-Evolution (approved):

PR #3977, opened 2022-08-31
#3976 [Bug Report] Backward compatibility was broken for queryData we…

### LINKED ISSUE #3976: [Bug Report] Backward compatibility was broken for queryData web services
## Bug report
Backward compatibility was broken for the queryData web services, in the original implementation the results are retrieved as an element and now the results will be retrieved as attributes.

Bug introduce with this commit a3b7c279e70df6ae856e3c02eaff68147937fdc8

This makes any client that connects to the web services have to change their APIs to process the results that are returned.

On the other hand, we have a problem that is possible that after the Backward compatibility was broken, somebody implemented based on retrieved attributes, so, this also affects any APIs that use this approach.

The solution is to make the results configurable,  based on web server parameter RetriveResultAs that are defined as elements by default settings.

If any service API requires that the results be returned as attributes, they need to add a configuration parameter.
RetriveResultAs = Attributes or RetriveResultAs = Elements

#### Steps to reproduce
- You can setup any web service for  readData or  queryData and check how they retrieve results

#### Screenshot or Gif

image

#### Expected behavior
Contracts can be improved but never backward compatibility broken

- Your OS: macOS
- Java version: JDK 11
- ADempiere version: 3.9.3
- Browser Safari
- DB Provider: PostgreSQL
- DB version: 14

#### Additional context
It is very important for the current implementation base to maintain backward compatibility at all costs since many developments could have been done with the original specification, breaking implies that all that installed base is affected.

e-Evolution:
Fixed #3976

Bug report

Backward compatibility was broken for the queryData web services, in the original implementation the results are retrieved as an element and now the results will be retrieved as attributes.

Bug introduce with this commit https://github.com/adempiere/adempiere/commit/a3b7c279e70df6ae856e3c02eaff68147937fdc8

This makes any client that connects to the web services have to change their APIs to process the results that are returned.

On the other hand, we have a problem that is possible that after the Backward compatibility was broken, somebody implemented based on retrieved attributes, so, this also affects any APIs that use this approach.

The solution is to make the results configurable, based on web server parameter RetriveResultAs that are defined as elements by default settings.

If any service API requires that the results be returned as attributes, they need to add a configuration parameter.
RetriveResultAs = Attribute or RetriveResultAs = Element

Expected behavior

Contracts can be improved but never backward compatibility broken

Other relevant information

Your OS: macOS
Java version: JDK 11
ADempiere version: 3.9.3
Browser Safari
DB Provider: PostgreSQL
DB version: 14
Additional context

It is very important for the current implementation base to maintain backward compatibility at all costs since many developments could have been done with the original specification, breaking implies that all that installed base is affected.


e-Evolution:
Configuration for queryData as default behavior:

image

image

image

image

image

Result When the call of Service : 

image

Adding the Web Service Parameter : 

image

Result When the call of Service with RetriveResultAs parameter As Attribute:

image
 
Result When the call of Service with RetriveResultAs parameter As Element:

image

e-Evolution:
Apply the same Configuration for readData as default behavior:

image

image

image

image

image

Result When the call of Service default  behavior :

image

Adding the Web Service Parameter :

image

Result When the call of Service with RetriveResultAs parameter As Attribute:

image

Result When the call of Service with RetriveResultAs parameter As Element:

image

carlosaparadam:
Hi Victor @e-Evolution, i test it and it's fine

PR #3981, opened 2022-08-31
#3980 [Bug Report] Connection parameters setKeepaliveTime not is vali…

### LINKED ISSUE #3980: [Bug Report]  Connection parameters setKeepaliveTime not is valid for Swing Client
## Bug report
Connection parameters setKeepaliveTime not is valid for Swing Client 

#### Steps to reproduce
1.- Setup ADempiere and run using idk 172. 
2.- execute RUN_Adempiere.sh
3.- Error  Connection setKeepaliveTime parameters

#### Expected behavior
If connect using Swing Client then should connect without some error

- Your OS: macOS
- Java version: 17 
- ADempiere version: develop
- DB Provider: PG
- DB version: 14

e-Evolution:
…d for Swing Client https://github.com/adempiere/adempiere/issues/3980

fixed #3980

### DIFF
--- a/base/src/org/compiere/db/DB_MariaDB.java
+++ b/base/src/org/compiere/db/DB_MariaDB.java
@@ -514,7 +514,6 @@ public class DB_MariaDB implements AdempiereDatabase {
 				config.setPassword(connection.getDbPwd());
 				config.setConnectionTestQuery(DEFAULT_CONN_TEST_SQL);
 				config.setIdleTimeout(0);
-				config.setKeepaliveTime(30000);
 				config.setMinimumIdle(15);
 				config.setMaximumPoolSize(150);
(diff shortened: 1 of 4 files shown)


marcalwestf:
@e-Evolution  Víctor, is there an error to be reproduced, or is it only that in principle _setKeepaliveTime_ should not be set?
I notice no difference in Postgres. I don't know the behavior in the other databases.
Thanks

e-Evolution:
Hi @marcalwestf This parameter should not be set for the Swing connection as the idea is that the connection pool is maintained, if the code changes it to all database implementations.

The other error it shows you is because Tomcat and Jetty don't support RMI with EJB on port 1099, so you need to connect to the database manually.

marcalwestf (approved):

Issue #3983, opened 2022-09-05
[Bug Report] Is reconciled flag is not correctly set when a payment is in an account statement with draft document status


e-Evolution:
## Bug report

1. Create a Payment as "Is Receipt" in Garden World and complete the document
2. Create a Bank Statement and add the payment into Bank Statement Line, but not complete the Bank Statement
3. Revert Payment.
4. Complete the Bank Statement and Void it after

#### Link to minimal reproduction

Using standard Garden World demo

#### Expected behavior
The reconciled should have as result the same that reversed document

#### Other relevant information
- Your OS: MaxOS
- Java version: 17
- ADempiere version: last develop for 3.9.4
- DB Provider: PG 
- DB version: 14

PR #3984, opened 2022-09-05
 #3983 [Bug Report] Is reconciled flag is not correctly set when a payment is in an account statement with draft document status

e-Evolution:
Is reconciled flag is not correctly set when a payment is in an account statement with draft document status https://github.com/adempiere/adempiere/issues/3983

fixed 3983


SusanneCalderon (changes requested):
@e-Evolution;
When the bank statement is already voided this bank statment lines should not be considered.
```

	private int getC_BankStatementLine_ID() {
		String sql = "SELECT bsl.C_BankStatementLine_ID FROM C_BankStatementLine bsl WHERE bsl.C_Payment_ID=? "
				+ "AND EXISTS(SELECT 1 FROM C_BankStatement bs "
				+ "					WHERE bs.C_BankStatement_ID = bsl.C_BankStatement_ID "
				+ "AND bs.DocStatus NOT IN('VO', 'RE'))";
		return DB.getSQLValue(get_TrxName(), sql, getC_Payment_ID());
	}	//	getC_BankStatementLine_ID
```

e-Evolution:
@SusanneCalderon Hi Susanne, Currently, there is no business logic for "Reverse Correct" in the bank statement, in the case of Void, the lines are emptied and the invoice and payment references are cleared, so there would be no point in validating why they have no effect on the logic.

niclopup:
tested OK

https://user-images.githubusercontent.com/31936815/190438035-f177132d-b424-416d-9f10-060bbbf4f699.mp4

SusanneCalderon (approved):
@e-Evolution , ok, as there is at this moment only a  business logic for void , it is ok not to filter for voided bank statements.

e-Evolution:
@marcalwestf please Can you merge this bugfix?

PR #3986, opened 2022-09-13
#3985 [Feature Request] Adding URL Map to a location map to Business …

### LINKED ISSUE #3985: [Feature Request] Adding URL Map to a location map to Business Partner Location
## Feature request
Adding field with URL Map to a location to Business Partner Location

It is quite useful to use an URL Map to get the exact location of an address.

This functionality allows you to open the map with the address in embedded format or in a new tab.

This new feature looks for longitude and latitude coordinates in the URL Map to synchronize it with the address data.

image

#### Describe alternatives you've considered
Currently, the system does not have an option, but it tries to build the URL dynamically but it may not be exact because it does not use longitude or latitude.

e-Evolution:
…Partner Location  https://github.com/adempiere/adempiere/issues/3985

fixed #3985


marcalwestf (changes requested):
- Load and apply _09850_3985_Adding_URL_Map_To_A_Location_ Map.xml_
- Open GoogleMaps and select a location:
imagen

- On Swing: Open a Business Partner and save the URL.
An error is produced:
imagen

- Additional information: the Address has latitude. longitude and altitude.

imagen

Perhaps it is only implemented for ZK?

marcalwestf:
@e-Evolution I noticed **after** testing, that there is a conflict that must be resolved.
Maybe because of the last merge.

e-Evolution:
> @e-Evolution I noticed **after** testing, that there is a conflict that must be resolved.
> Maybe because of the last merge.

@marcalwestf the conflict was solve

marcalwestf (changes requested):
Repeated the test:
- Build new binary -> OK
- Run manually RUN_Import.sh -> OK
- Run manually RUN_MigrateXML -> OK
- Make sure all XMLs were imported -> OK
- Install Adempiere on Tomcat -> OK
- Run Tomcat server -> OK
- Open BP, create a location entry **without** MAP URL  -> OK
- Open Google Maps -> OK
- Select a random address -> OK
  (Following instructions from e.g. https://help.webriti.com/common-docs/tutorials/how-to-get-google-map-url/)
- Try this address on an empty browser -> OK
   (The URL finds the address on Google Maps)
- Copy link address into field _Map URL_ -> OK
- Save record -> ERROR

I selected the address as specified on the Google Maps help. It delivers addresses like https://goo.gl/maps/CvHJ6TiL3q82t37X6 for the Brandenburger Tor in Berlin, Germany:
imagen

 This very address (https://goo.gl/maps/CvHJ6TiL3q82t37X6) delivers an error on Adempiere:
20220925_Google_Maps_01


I tried several other addresses with the same results.

Perhaps the program is expecting the address in another format. The example you put in the issue has names in the URL, whereas all the URL  addresses I get from Google maps are alphanumeric combinations.

e-Evolution:
@marcalwestf the error is because the url map check that the url format is based on this format  https://maps.google.com/?q=41.0530382,-73.5318390  , the reason is because the goal is get the latitude and longitude  to use some other GPS system, can testing with a Map URL with this format.


This format is same that IOS can get when share any location i.e. with Whatapps

marcalwestf:
@e-Evolution  Víctor, FYI: if I put the correct coordinates on the latitude and longitude of the address, and then press the **Map** button, the address is displayed correctly on Google Maps!!
imagen

Same happens when you select the world icon on the right side of the Address field:
20220925_Google_Maps_02

e-Evolution:
@marcalwestf other import subject is that the latitude and longitude are save on or syncronize into the table  C_Location

marcalwestf:
Now I got it: the field **Map URL** accepts only URL of the following format:
`https://maps.google.com/?q=<latitude>,<longitude>`

For example, for the Brandenburger tor in Berlin, Germany, the address would be:
`https://maps.google.com/?q=52.516266,13.377775`

As the code in this PR, the other way to find the Brandenburger Tor  (which is the one Google Maps offers the link to share) is not allowed:
`https://goo.gl/maps/NpvKMxFeSBuyNUxe6`

This may be interesting to the users who like me didn't know that there is only one format allowed.

e-Evolution:
> Now I got it: the field **Map URL** accepts only URL of the following format: `https://maps.google.com/?q=<latitude>,<longitude>`
> 
> For example, for the Brandenburger tor in Berlin, Germany, the address would be: `https://maps.google.com/?q=52.516266,13.377775`
> 
> As the code in this PR, the other way to find the Brandenburger Tor (which is the one Google Maps offers the link to share) is not allowed: `https://goo.gl/maps/NpvKMxFeSBuyNUxe6`
> 
> This may be interesting to the users who like me didn't know that there is only one format allowed.

@marcalwestf  I modify the description and help and other hand now allow any url map , please test if is ok then merge.

kind regards

marcalwestf (approved):
With the last changes, it is now possible to save the Google Maps URL in either format.

Tests passed.

Issue #3987, opened 2022-09-20
[Feature Request] Synchonize Entity Type


e-Evolution:
## Feature request
Synchronize entity type based on the main men

![image](https://user-images.githubusercontent.com/786968/191151617-cc3baa82-bfc0-459c-b80b-d86ea10b16cf.png)

#### Describe the solution you'd like
This process allows for synchronizing the Entity Type for the elements defined in the menu, processes, windows, forms, etc.

#### Describe alternatives you've considered
Not is posible change now

marcalwestf:
@e-Evolution  Victor, is the last commit of this pull request only refactoring that has not too much to do with the functionality of this PR?

I got confused at first, and only saw the last commit and missed the changes for the new process.

Perhaps it can be separated: one pull request for the functionality, another for refactoring. Probably in this case, first the refactoring, and later the functionality.

I will test it right now.

e-Evolution:
Hello @marcalwestf Mario!

the functionality is related to the refactory final, the process changes the entity type within the application dictionary components, it also includes the missing refactory from @yamelsenih 

These changes are the result of the review that was made of all the changes made by @Yamel and had not been made since it was agreed that all classes X_ and I_ should all be here org.adempiere.core.domains.models.

On the other hand I fix the sequences of the XML Script , not if another PR is required just to change the sequence and the name of the XML file.

PR #3988, opened 2022-09-20
#3987 [Feature Request] Synchronize Entity Type https://github.com/ad…

e-Evolution:
…empiere/adempiere/issues/3987

#3987 fixed


marcalwestf (changes requested):
- git pull
- open Adempiere swing as to load and apply the XML files
- an error is thrown:
imagen

- looking at the DB, I notice that for the class **FixedAsset** the code defines the package **org.compiere.asset.model.validator** whereas the database expects a model validator **org.compiere.FA.model.ModelValidator** .
imagen

Perhaps the refactoring was done after testing the functionality, but it was not tested after the last commit.

e-Evolution:
@marcalwestf Hi Mario 

The Fixed Asset Package was moved from org.compiere.FA to org.compiere.asset was one of the things I had requested in PR from @yamelsenih  which I did as it was not included.

So if you want to test this functionality you must generate the database seed again , since the fixed assets functionality refactory XML was changed , if you generate your seed again you should not have any problems , use RUN_MigrateXML.sh to perform this task

marcalwestf:
I find the error here:
imagen

marcalwestf:
ahhh ok Victor, thanks.
I will create the new seed and test on it.

marcalwestf (changes requested):
Swing does not work either from the installation or from the IDE:
imagen

imagen

ZK works.

**This error is not caused by the changes in this pull request**.
Rather, it is caused by changes done before: I tested this and other pull requests and they also do not work on Swing.

@e-Evolution , @yamelsenih , @niclopup  et al, please confirm if you can reproduce the error, and if you can reproduce it, perhaps you help find out the pull request which causes the error.

I will continue to try out and document it on this thread.

e-Evolution:
> Swing does not work either from the installation or from the IDE: imagen
> 
> imagen
> 
> ZK works.
> 
> **This error is not caused by the changes in this pull request**. Rather, it is caused by changes done before: I tested this and other pull requests and they also do not work on Swing.
> 
> @e-Evolution , @yamelsenih , @niclopup et al, please confirm if you can reproduce the error, and if you can reproduce it, perhaps you help find out the pull request which causes the error.
> 
> I will continue to try out and document it on this thread.

@marcalwestf the image show that not exists the AD_System table , this only happen when the database have not ADempiere a valid connection or RUN_ImportAdempiere.sh not was executed , so check first that the connection si valida and the the AD_System table exist.

marcalwestf:
@e-Evolution  Víctor, thanks for your hinh. I found the error.
For some weird reason I haven't investigated yet, after executing RUN_ImportAdempiere.sh and RUN_Migrate_XML.sh, the database had the schema "postgres", instead of "adempiere.

After running the command
`ALTER SCHEMA postgres RENAME TO adempiere`
I can run Swing...

Does it happen to you or can you imagine why It happened.
It run it twice and both times, the schema was "postgres".
Perhaps I must drop the database first... I'll check out and document it here....

marcalwestf:
The process  "Synchronize Entity Type" seems not to have defined the Data Access  Level:
imagen

e-Evolution:
> The process "Synchronize Entity Type" seems not to have defined the Data Access Level:
> imagen

@marcalwestf Than you , Mario , I fixed to Only System

kind regards
Victor

marcalwestf (approved):
- Build -> OK
- XML load and apply -> OK
- Check process and parameters -> OK
- Check code -> OK
- Check window, process etc:

imagen

Check results -> OK
The entity type of the window, process etc. was changed accordingly to the parameter selected.

Notes:
- id changes recursively menu objects
- it doesn't change recursively inside an object. It only changes the selected object.
  This means for the window, the process changes the window, tab, and fields, but not the table and column.

Tests passed

Issue #3991, opened 2022-10-06
Error in jasperreport because JASPER_STARTER_CLASS  in ProcessInfo is not updated


SusanneCalderon:
## Bug report
The static variable JASPER_STARTER_CLASS in the class ProcessInfo has the old path. As the class was moved, this pathe mus be updated to the new path.

## Steps to reproduce:
Try any process with an attached jasperreport, where the field classname is empty.

```

public final class ProcessUtil {

	public static final String JASPER_STARTER_CLASS = "org.compiere.report.ReportStarter";
```

Correct variable:
```

public final class ProcessUtil {

	public static final String JASPER_STARTER_CLASS = "org.compiere.jr.report.ReportStarter";
```

marcalwestf:
this issue was solved by accepting the pull request

PR #3992, opened 2022-10-06
#3991/bug/jasper starter class

SusanneCalderon:
#3991
The static variable JASPER_STARTER_CLASS has still the old path. So, when in the definition of a jasper process the classname is not set to the new path _org.compiere.jr.report.ReportStarter_, a class not found exception occurs. 
```

public final class ProcessUtil {

	public static final String JASPER_STARTER_CLASS = "org.compiere.report.ReportStarter";

	/**	Logger				*/
	private static CLogger log = CLogger.getCLogger(ProcessUtil.class);
	
	private ProcessUtil() {}

```

With the updated variable jasper reports can be loaded.


e-Evolution (approved):
@SusanneCalderon Thank you a lot  for your bug fix

SusanneCalderon:
@e-Evolution thank you Victor.

PR #3993, opened 2022-10-10
Adding URL Map to a location map to Business Partner Location 

e-Evolution:
#3985 fixed


marcalwestf (approved):
- Check Code  -> OK
  (some refactoring and change setting from beforeSave() to afterSave()
- Build -> OK
- Test with different locations -> OK

Issue #3989, opened 2022-09-26
[Bug Report] Error when re-creating a view by altering a table with the t_alter_column function causing nested casting making recreated views slower by applying nested casting


e-Evolution:
## Bug report
Error when re-creating a view by altering a table with the t_alter_column function causing nested casting making recreated views slower by applying nested casting

problem was introduced in PR #3198

#### Steps to reproduce
1.- Execute this sentence multiples times 
`INSERT INTO t_alter_column values('c_invoice','Description','TEXT',null,'NULL');`
2.- Review the dependences views 
NOTICE:  drop view rv_c_invoice_customerprodqtr
NOTICE:  drop view rv_c_invoice_customervendqtr
NOTICE:  drop view rv_c_invoice_day
NOTICE:  drop view rv_c_invoice_month
NOTICE:  drop view rv_c_invoice_prodmonth
NOTICE:  drop view rv_c_invoice_productmonth
NOTICE:  drop view rv_c_invoice_productqtr
NOTICE:  drop view rv_c_invoice_prodweek
NOTICE:  drop view rv_c_invoice_vendormonth
NOTICE:  drop view rv_c_invoice_week
NOTICE:  drop view c_invoiceline_v
NOTICE:  drop view rv_bpartneropen
NOTICE:  drop view rv_c_invoiceline
NOTICE:  drop view rv_combinedopenitem
NOTICE:  drop view rv_fm_openloan
NOTICE:  drop view rv_openitem
NOTICE:  drop view rv_openitemtodate
NOTICE:  drop view c_dunning_line_v
NOTICE:  drop view c_dunning_line_vt
NOTICE:  drop view c_invoice_header_v
NOTICE:  drop view c_invoice_header_vt
NOTICE:  drop view c_invoice_linetax_v
NOTICE:  drop view c_invoice_linetax_vt
NOTICE:  drop view c_invoice_v
NOTICE:  drop view c_invoice_v1
NOTICE:  drop view c_payselection_remittance_v
NOTICE:  drop view c_payselection_remittance_vt
NOTICE:  drop view rv_c_bankstatement
NOTICE:  drop view rv_c_invoice
NOTICE:  drop view rv_c_invoicetax
NOTICE:  drop view rv_commissionrundetail
NOTICE:  drop view rv_fm_loanamortization
NOTICE:  drop view rv_inout_createfrom
NOTICE:  drop view rv_invoice_createfrom
NOTICE:  drop view rv_unposted
[...]

PR #3997, opened 2022-10-11
#3989 [Bug Report] Error when re-creating a view by altering a table

e-Evolution:
…with the t_alter_column function causing nested casting making recreated views slower by applying nested casting https://github.com/adempiere/adempiere/issues/3989

### DIFF
--- a/db/ddlutils/postgresql/functions/altercolumn.sql
+++ b/db/ddlutils/postgresql/functions/altercolumn.sql
@@ -1,4 +1,6 @@
-create or replace function altercolumn(tablename name, columnname name, datatype name,
+DROP TABLE t_alter_column;
+DROP FUNCTION altercolumn(tablename name, columnname name, datatype name,nullclause varchar, defaultclause varchar);
+CREATE OR REPLACE FUNCTION altercolumn(tablename name, columnname name, datatype name,
 nullclause varchar, defaultclause varchar) returns void as $$
 declare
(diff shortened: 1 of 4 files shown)


piracio (approved):
I don't see anything strange at code level. looks good.

marcalwestf:
@piracio  could you run the XMLs?
Thanks!

marcalwestf:
@e-Evolution  hi Victor, what should be exactly the steps to reproduce the error?
I did not understand well.

Besides, the correction would be to run the XML files and then do the steps again, right?
Thank you.

e-Evolution:
@marcalwestf Hi Mario!

The test is simple to change some column and synchronizes the column.

i.e. change the length of POReference to 255. for the tables C_Order or C_Invoice, this should allow the synchronization and rebuild the dependences views.

i.e. change the length of Descripcion to 4024 into the C_OrderLine or C_InvoiceLine, this should allow the synchronization and rebuild the dependences views.

You can review some views where the field POReference or Description is used and compare between the original view and reapply views.

kind regards
Victor

piracio (approved):
Feature tested with OL7/ postgreSQL 12
Deployment OK
Migration OK

piracio (approved):
Test OK

piracio (approved):
Test OK

piracio (approved):
Test OK

Issue #3990, opened 2022-10-02
Rules do not work in Adempiere 394


SusanneCalderon:
## Bug report
I tested 394 in a database from a running installation which uses rules for callouts and table script validators. The database was updated until xml 9,880. Every rules produced the same error: "Script1.groovy: 1: unable to resolve class org.compiere.model.xxx". 

To confirm this error I tested  a new rule in the newest seed database from github and got the same error.

#### Steps to reproduce
**1. Create a rule**
![image](https://user-images.githubusercontent.com/15349639/193435052-1bd34c9f-38d3-4e94-aad0-ebea117bd70f.png)

**2. Assign this rule to table c_Invoice**

![image](https://user-images.githubusercontent.com/15349639/193435082-b208973b-7f9c-412e-8ab0-143b6dce2590.png)

**3. Stop the Server and start the server**

**4. Generate a new invoice and try to save**

![image](https://user-images.githubusercontent.com/15349639/193435151-c48cfa65-f80d-448c-b2c7-a9f52fb06db1.png)

Complete error message: Error: _javax.script.ScriptException: org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed: - Script1.groovy: 1: unable to resolve class org.compiere.model.MInvoice @ line 1, column 1. import_

#### Additional context
The problem only occurs in webUI, in Swing the invoice is saved and the rule is executed.
![image](https://user-images.githubusercontent.com/15349639/193435226-ff396777-b3a8-4272-8804-7b3ef66ddc7d.png)

marcalwestf:
Further findings: one first thought was that possibly there are different groovy versions in Swing and ZK. 
Actually, there is only one groovy version, and the libraries for Swing and ZK are identical.

Here the images that support it:
1.- There is only one place in the repository where the groovy libraries are kept:
![imagen](https://user-images.githubusercontent.com/1789408/194415185-f4e2c90a-716e-417c-99f8-fd3937025129.png)

2.-  in tools/build.xml both libraries groovy-all-2.4.15.jar and groovy-engine.jar are defined for **CSTools.jar**
![imagen](https://user-images.githubusercontent.com/1789408/194413763-074e3d98-6765-4449-8861-ca0579f29d22.png)

3.-  in tools/build.xml both libraries groovy-all-2.4.15.jar and groovy-engine.jar are defined for **CCTools.jar**
[...]

PR #4001, opened 2022-10-13
#3990 Rules do not work in Adempiere 394  https://github.com/adempier…

e-Evolution:
…e/adempiere/issues/3990

fixed ##3990


SusanneCalderon (approved):
I tested a tablescriptvalidator in table c_Incoive with a groovy rule, instalation with tomcat.
image
The reported error did not occur in webui, the rule was executed correctly. 
image

Issue #4018, opened 2022-11-24
Re-number XML Migration Files Previous to Release


marcalwestf:
Previous to the next release, it is necessary that the XML files get a sequence numbering.

PR #4019, opened 2022-11-25
Re-number XML Migration Files Previous to 394 Release

marcalwestf:
Implements #4018 

Previous to the next 3.9.4 release, it is necessary that the XML migration files get individual sequence numbering.


e-Evolution (approved):
I ran the complete migration without errors

PR #4025, opened 2022-12-07
Fixed error with run process without transaction

yamelsenih:
# The Problem
When a process is running without transaction using the `ProcessBuilder`, this execution make a new transaction for it.

All is ok but when a workflow is processed the transaction is openend and never ending.

# The solution

Just let the process handle the own transaction if the `ProcessBuilder` is called without transaction

# Step for Step
A example for run

```Java
ProcessBuilder builder = ProcessBuilder.create(Env.getCtx())
			.process(111)	//	C_Invoice Process
			.withRecordId(318, 100)	//
			.withoutPrintPreview()
			.withWindowNo(0)
			.withTitle("Process Invoice")
			.withParameter("C_Invoice_ID", recordId)
			.withParameter("DocAction", "CO");
		//	Execute Process
builder.execute(null);
```

### DIFF
--- a/base/src/org/eevolution/services/dsl/ProcessBuilder.java
+++ b/base/src/org/eevolution/services/dsl/ProcessBuilder.java
@@ -32,7 +32,6 @@ import org.compiere.process.ProcessInfoUtil;
 import org.compiere.util.ASyncProcess;
 import org.compiere.util.Env;
 import org.compiere.util.Trx;
-import org.compiere.util.TrxRunnable;
 import org.compiere.util.Util;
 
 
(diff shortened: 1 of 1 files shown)


rcapecce:
Hi @yamelsenih , could you provide an example to test this modification in a seed?

e-Evolution:
Hi @yamelsenih I consider that you should never call an execute without a valid transaction, the intention of passing the transaction is so that a process is executed with a parent transaction and that the process that is executed does not close it when completing the process, so the correct way to call the API is as follows.


```
Trx.run(trxName -> {
            //Create partial return
            commandReceiver.setProcessInfo(org.eevolution.services.dsl.ProcessBuilder
                    .create(commandReceiver.getCtx())
                    .process(commandReceiver.getProcessId())
                    .withTitle(commandReceiver.getName())
                    .withParameter(org.adempiere.pos.process.CreateOrderBasedOnAnother.C_OrderSource_ID, commandReceiver.getOrderId())
                    .withParameter(org.adempiere.pos.process.CreateOrderBasedOnAnother.Bill_BPartner_ID, commandReceiver.getPartnerId())
                    .withParameter(org.adempiere.pos.process.CreateOrderBasedOnAnother.DocSubTypeSO, org.compiere.model.MDocType.DOCSUBTYPESO_ReturnMaterial)
                    .withParameter(org.adempiere.pos.process.CreateOrderBasedOnAnother.DocAction, org.compiere.process.DocAction.ACTION_None)
                    .withParameter(org.adempiere.pos.process.CreateOrderBasedOnAnother.IsIncludePayments, false)
                    .withParameter(org.adempiere.pos.process.CreateOrderBasedOnAnother.IsAllocated, false)
                    .withoutTransactionClose()
                    .execute(trxName)
            );
        });```

this way you can call a process inside another making everything atomic.

So the validation should be if a null value is passed, the execution should not be allowed since the execute() method exists for this.

yamelsenih:
This is very hard to replicate

Issue #4038, opened 2022-12-30
[Bug Report] Error when create the Client with last develop 


e-Evolution:
## Bug report
===========> InitialClientSetup.process: Error while creating the Accounting Setup. 
Check the error log and the format of the Accounting*.cvs file - No duplicate Accounts and all all defaut accounts must be included (even if you will not use them). [143]
org.adempiere.exceptions.AdempiereException: Error while creating the Accounting Setup. 
Check the error log and the format of the Accounting*.cvs file - No duplicate Accounts and all all defaut accounts must be included (even if you will not use them).
        at org.adempiere.process.InitialClientSetup.doIt(InitialClientSetup.java:187)
        at org.compiere.process.SvrProcess.process(SvrProcess.java:176)

#### Steps to reproduce
1. Build the last code 
2. Execute RUN_setup.sh 
3. Execute RUN_ImportAdempiere.sh 
4. Execute RUN_MigrationXML.sh 
5. Try to Create a new Client

#### Expected behavior
The client should be create as previous versions.

#### Other relevant information
- Your OS: MacOs
- Java version: 11
- ADempiere version: 3.9.4 night build
- DB Provider: PG 
- DB version: 15

#### Additional context
- I compare the class and I can review that this class was changed
InitialClientSetup
MSetup.importChart(File chart)

May be here was introducing the problems , I not sure is necessary review with detail

marcalwestf:
Here the steps to reproduce the first exception:

1.- The BP Group _Standard_ is being created in _MSetup.createEntities()_ :
![imagen](https://user-images.githubusercontent.com/1789408/210115850-573db35a-b99c-43f9-b0df-138e6ce87a4c.png)

2.- The call stack takes us up to _MAcctSchema.getClientAcctSchema()_, where `MClientInfo.get(ctx, AD_Client_ID);` is called:
![imagen](https://user-images.githubusercontent.com/1789408/210115948-1f45899b-7bc4-4c19-91ed-9d75fc6a870a.png)

3.- _MClientInfo.get()_ returns **null**:
![imagen](https://user-images.githubusercontent.com/1789408/210116104-66189a47-3641-4726-af90-0b7572e121ae.png)

4.- Back to  _MAcctSchema.getClientAcctSchema()_, trying to de-reference _info.getC_AcctSchema1_ID()_ will lead to an exception:
[...]

PR #4040, opened 2023-01-02
adempiere Solving #4038 Error when create the Client with last develop

SusanneCalderon:
Solves https://github.com/adempiere/adempiere/issues/4038

**1. Error:`** Null Pointer Exception, when creating _c_BpGroup_Acct_.
Creating C_BP_Group_Acct uses MClientInfo. MClientInfo is not found by the client_ID.

**2. Error**: Null Pointer Exception when creating _c_AcctSchema_GL_  and _C_Acctschema_Default_
_c_AcctSchema_ID_  is not present in the database when saving theese objects.

### Solution
**Class _MSetup_, methods _createClient()_ and _createAccounting()_**
Initializing the class MSetup, a transaction _Setup_ is created and started.
When at the end of createClient() and createAccounting() a commit is included, the process _InitialClientSetup_ is completed without error.
```

public boolean createClient (String clientName, String orgValue, String orgName,
		String userClient, String userOrg, String phone, String phone2, String fax, String eMail, String taxID,
		String DUNS, String logoFile, int Country_ID)
	{...

		log.info("fini");
		m_trx.commit();
		return true;
	}   //  createClient
```
image

### Problem
The class _InitialClientSetup_ has it own transaction. In case of an error the database contains object from the commit in class _MSetup_

### DIFF
--- a/base/src/org/compiere/model/MSetup.java
+++ b/base/src/org/compiere/model/MSetup.java
@@ -424,6 +424,7 @@ public class MSetup
 		rp.saveEx();
 		
 		log.info("fini");
+		m_trx.commit();
 		return true;
 	}   //  createClient
 
@@ -523,6 +524,7 @@ public class MSetup
 			m_trx.close();
 			return false;
 		}
+		m_trx.commit();
 		//  Info
 		m_info.append(Msg.translate(m_lang, "C_AcctSchema_ID")).append("=").append(m_as.getName()).append("\n");
 


marcalwestf:
**Recapitulating the situation**

- An exception occurs when reading AD_Client directly from the database (via Query): it yields null.
- AD_Client_ID is needed for creating + saving other objects.
- Same with C_AcctSchema.
- The AD_Client was saved before.
- Saving AD_Client apparently doesn't get through the database.
- Saving AD_Client goes up to PO.saveNew() without errors, but it is not visible on the database.
- InitialClientSetup has a transaction "null" and MSetup a transaction "Setup".
- InitialClientSetup hasn't been changed since 393.
- MSetup was only modified in the class import since 393.
- So it is not a matter that either class was modified incorrectly.
- Neither was the class moved to another package.
- When committing the MSetup transaction at  Client or Account Schema creation, the changes can be seen in the database. This is what this pull request does.
- Also, the reading of AD_Client and C_AcctSchema don't yield null anymore.
- This solution has a possible side-effect: if (when) there is an error, the Client and the Account Schema stay in the database.
- This can be avoided by deleting both when catching an error.
- The deletion can be easily done when this modification has been approved.
- Otherwise it would be futile work.

e-Evolution:
@SusanneCalderon Hi Susanne, @marcalwestf  Mario While this approach solves the problem, it does introduce a potential problem, it makes the company creation process non-atomic as it forces a transaction to commit.

The problem is caused by incorrect transaction handling.

the fixed is here #4041

Issue #4056, opened 2023-01-27
[Bug Report] Error when try print any Jasper Report from Swing client


e-Evolution:
## Bug report
Error when try print any Jasper Report from Swing client
introducing by @yamelsenih and @marcalwestf  with #4026

#### Steps to reproduce
1.- Try print any jasper report with is Print Preview Yes

15:05:03.865 ADialog.error: null - ProcessError null [19]
===========> ProcessUtil.startJavaProcess: org.compiere.jr.report.ReportStarter [75]
java.lang.NullPointerException
	at org.compiere.jr.report.ReportStarter.startProcess(ReportStarter.java:666)
	at org.adempiere.util.ProcessUtil.startJavaProcess(ProcessUtil.java:171)
	at org.adempiere.util.ProcessUtil.startJavaProcess(ProcessUtil.java:116)
	at org.compiere.apps.ProcessCtl.startProcess(ProcessCtl.java:610)
	at org.compiere.apps.ProcessCtl.run(ProcessCtl.java:392)
	at java.base/java.lang.Thread.run(Thread.java:829)

#### Link to minimal reproduction

Last 394 release

#### Expected behavior
the jasper report should work as before.

#### Other relevant information
- Your OS: macOS
- Java version: 11
- ADempiere version: 3.9.4
- DB Provider: PG
- DB version: 15.1

#### Additional context
Testing Swing Client

marcalwestf:
Tested on Swing with the same configuration as #4026.
Result:
![imagen](https://user-images.githubusercontent.com/1789408/215620394-b89325c8-8e55-4680-a2d5-4278e57f4598.png)

marcalwestf:
The problem: in Swing _ReportStarter.startProcess()_ line 664 `getReportViewerProvider()` returns **null**:
![imagen](https://user-images.githubusercontent.com/1789408/215621962-2a9ff58d-a870-4919-a0fa-ebf7e1f34ba5.png)
..if ReportStarter.startProcess() is changed in line 664 to
`JRViewerProvider viewerLauncher = getReportViewerProvider()==null? new SwingJRViewerProvider():getReportViewerProvider();`

..then the Jasper Report is displayed in Swing:
![imagen](https://user-images.githubusercontent.com/1789408/215622718-9ac92c3f-dd15-4aac-a82f-aebd558f1980.png)

[...]

PR #4026, opened 2022-12-08
Remove JDK8 compatibility gradle

yamelsenih:
Just a minor change for all gradle subproject and remove JDK8 compatibility flag

PR #4059, opened 2023-01-31
#4056 [Bug Report] Error when try print any Jasper Report from Swing …

e-Evolution:
…client https://github.com/adempiere/adempiere/issues/4056

Error when try print any Jasper Report from Swing client introduced with #4026

### DIFF
--- a/client/src/org/compiere/apps/ProcessCtl.java
+++ b/client/src/org/compiere/apps/ProcessCtl.java
@@ -28,6 +28,8 @@ import org.adempiere.core.domains.models.I_AD_Process;
 import org.adempiere.util.ProcessUtil;
 import org.compiere.db.CConnection;
 import org.compiere.interfaces.Server;
+import org.compiere.jr.report.ReportStarter;
+import org.compiere.jr.report.SwingJRViewerProvider;
 import org.compiere.model.MPInstance;
 import org.compiere.model.MProcess;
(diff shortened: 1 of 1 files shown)


marcalwestf:
@e-Evolution  hi Victor,
- I didn't find the branch feature/4056 on your repository, but on the ADempiere repository, which I suppose is incorrect, unless you have a compelling reason to do it:
imagen
- According to the descrption of milestones (https://github.com/adempiere/adempiere/milestones), a pure bug fix (like this) should not be done against the **_develop_** branch. Instead, there should be an additional branch (e.g. **_develop-fix_**, **_develop-bugfixes_** or a name according to the developer community) that would collect pure fixes, to be later published for example in version 3.9.4a.
- The branch **_develop_** would be for features and complex bug fixes.

I will test the pull request, but I'd expect you to delete this branch (so the ADempiere repository doesn't get cluttered), to create a branch in your repository and to create a pull request that points to the bug fixes branch on the ADempiere repo.

Thanks!

marcalwestf:
just tested this PR: it runs Jasper Reports on both Swing and ZK.

marcalwestf:
... the changes seem ok to me, so if we agree on repository/branches and @yamelsenih also agrees on the changes, it can be merged.

e-Evolution:
Replace by this PR #4175

Issue #4057, opened 2023-01-30
Run_Adempiere is showing error.


amirsaifi:
## Bug report
Run_Adempiere.bat is showing error "
'functions.bat' is not recognized as an internal or external command,
operable program or batch file.
'""' is not recognized as an internal or external command,
operable program or batch file.

#### Steps to reproduce
1. Right Click 'Run_Adempiere.bat' and Run as Administrator.

#### Expected behavior
Start ADempiere client and show login screen.

#### Other relevant information
- Your OS: Windows 2012 R2
- Java version: Jdk 11
- ADempiere version: 394
- Browser Version [e.g. 22]
- DB Provider: PostgreSQL 15
- DB version:

marcalwestf:
@amirsaifi could you please describe step by step, starting from zero, better if with additional screenshots, how you installed ADempiere and how you are starting it.

With the information you supplied, I believe hardly anybody will be able to figure out what you did, much less find the cause of the error.

An example of an explanation step by step and with images can be found here: https://github.com/adempiere/adempiere/discussions/4058 .

amirsaifi:
@marcalwestf Given below are the steps.

1. Downloaded ADempiere394 LTS Zip.
2. Unzipped ADempiere and place in C:\ root (C:\Adempiere).
3. downloaded and installed jdk11 on path C:\jdk11
4. Installed PostgreSQL 15.
5. Downloaded tomcat 9 and unzipped in C:\ root (C:\tomcat9)
6. Set environment variables, ADEMPIERE_HOME, JAVA_HOME.
7. initiated setup with 'Run_Setup'. Test returned ok with all values set.
8. Configuration setup with 'Save'
9. After Configuration completed, started server with 'Start_Server2'
10. ADempiere is accessible in browser through tomcat.
11. All set and working fine.
12. Tried to open ADempiere Client with 'Run_ADempiere' which resulted in below mentioned error

Run_Adempiere.bat is showing error "
'functions.bat' is not recognized as an internal or external command,
operable program or batch file.
'""' is not recognized as an internal or external command,
operable program or batch file.

13. Tried 'Web Start' which resulted in below mentioned error on console.

Closing DownloadServiceListener
Reading Shortcuts with UTF-8
Using WindowsShortCutManager C:\Users\AmirSaifi\.cache\icedtea-web\cache\shortcutList.txt
Desktop Link Path : D:\Desktop/Adempiere Client 3.9.4LTS.lnk
[...]

PR #4064, opened 2023-02-07
set java-version: '17' to check the effect

homebeaver:
Hi @marcalwestf 

in my env: WIN10 + `temurin 17` I cannot build ADempiere with

`javac target="11"` in build.xml

This PR is a test what happens on ubuntu-latest with temurin 17

regards EUGen

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,7 +51,7 @@ jobs:
         with:
             # For available options: see https://github.com/actions/setup-java#supported-distributions
             distribution: 'temurin'
-            java-version: '11.0'
+            java-version: '17'
             
       # Deploy Tomcat-Server 
       - name: Deploy Tomcat ...


homebeaver:
OK - that is the same result I got on WIN10:

```
...
init:
     [echo] =========== Build Base
    [mkdir] Created dir: /home/runner/work/adempiere/adempiere/base/build

makedir:

compile:
    [javac] Compiling 3429 source files to /home/runner/work/adempiere/adempiere/base/build
    [javac] warning: target release 11 conflicts with default source release 17

BUILD FAILED
/home/runner/work/adempiere/adempiere/build.xml:29: The following error occurred while executing this line:
/home/runner/work/adempiere/adempiere/utils_dev/build.xml:21: The following error occurred while executing this line:
/home/runner/work/adempiere/adempiere/base/build.xml:69: Compile failed; see the compiler error output for details.
```

Usually this problem can be solved adding source attrib:

```xml 
...
<javac source="11" target="11"  ...
```

It works for all subprojects but in "base" you trap in faild tests ... :-(

marcalwestf:
@homebeaver  hi Eugen!
I must first understand the issue to answer correctly.

First, I'd like to assert that ADempiere 3.9.4 can only be built with JDK 11; whereas it can be installed with either JDK 11 or JDK 17. It is stated on the release notes. This works on Ubuntu and on Windows. See also https://github.com/adempiere/adempiere/issues/4057.

And I understand, this PR is not to be accepted but to make a point.

That said, I would like to know what you would like to achieve, what you have done, and what is missing.
Thanks!

e-Evolution:
Hi @homebeaver ADempiere version 3.9.4 is not yet compatible to compile 100% for JDK 17, the changes made only make it compilable for JDK 11, but you can to run 3.9.4 with JDK 17 in production.

e-Evolution:
@homebeaver If you are thinking of compiling develop for JDK 17 you should make sure that the whole project can be built with JDK 17 and that the code is compatible with JDK 17, I think that for a while it is good to live with JDK 11 and JDK 17.

So you can propose these changes and if the tests pass we can merge the develop branch

PR #4066, opened 2023-02-08
Fix path to class for Import Table

### LINKED ISSUE #4065: [Bug Report] Outdated path of class for import table
## Bug report

When executing a data import process with import tables, due to the change of packages, when the import class is searched, it cannot be found because it belongs to the EntityType Dictionary and does not have a package, so when setting the default , it takes an outdated package `org.compiere.model` instead of `org.adempiere.core.domains.models`.

This causes a cast error because it loads GenericPO instead of the required class.

https://github.com/adempiere/adempiere/blob/5fc84c4f87de754c6b2cc715e4d5c779c387ad5f/base/src/org/compiere/model/MTable.java#L231

https://github.com/adempiere/adempiere/blob/5fc84c4f87de754c6b2cc715e4d5c779c387ad5f/base/src/org/compiere/model/MTable.java#L226-L239

#### Expected behavior

when you need to use the default package, it should be `org.adempiere.core.domains.models` instead of `org.compiere.model`

- Your OS: Xubutnu 22.04
- Java version: 11
- ADempiere version: 394 develop
- Browser [e.g. chrome, safari]: Debug Swing
- DB Provider: PostgreSQL
- DB version: 12.4

rcapecce:
Fix #4065

### DIFF
--- a/base/src/org/compiere/model/MTable.java
+++ b/base/src/org/compiere/model/MTable.java
@@ -228,7 +228,7 @@ public class MTable extends X_AD_Table
 			MEntityType et = MEntityType.get(Env.getCtx(), entityType);
 			String etmodelpackage = et.getModelPackage();
 			if (etmodelpackage == null || MEntityType.ENTITYTYPE_Dictionary.equals(entityType))
-				etmodelpackage = "org.compiere.model"; // fallback for dictionary or empty model package on entity type
+				etmodelpackage = "org.adempiere.core.domains.models"; // fallback for dictionary or empty model package on entity type
 			Class<?> clazz = getPOclass(etmodelpackage + ".X_" + tableName, tableName);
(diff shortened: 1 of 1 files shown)


OFB-Review1:
se acordó que las correcciones estarán en la rama de hotfix, no la de develop. Por favor modificar.

e-Evolution:
@OFB-Review1 , @yamelsenih , the code is right , @yamelsenih can please resend the PR with hotfix branch please

rcapecce:
> @OFB-Review1 ,the code is right , @rcapecce Raul can please resend the PR with hotfix branch please

Ok, I pushed the PR to the hotfix branch

OFB-Review1 (approved):
the error is replicated and checks the correct operation of the new code

PR #4067, opened 2023-02-09
Minor Change: Dashboard

yamelsenih:
A minor chage that allows return a default graph when not exists a type for him.

just see the code

```Java
//  Change this
else
{
  throw new IllegalArgumentException("unknown chart type=" + type);
}


//  For it
return createBarChart();
```


e-Evolution:
Replace PR for #4147

PR #4070, opened 2023-02-09
Change modifier to public to TimeUtil methods

yamelsenih:
Just add the follow methods as public

- `public static int getDaysBetweenWithCalendar (Timestamp start, Timestamp end, boolean onlyMatchWithCalendar, int... includeDay)`

- `public static Timestamp addDays(Timestamp day, int offset, boolean onlyMatchWithCalendar, int... includeDay)`


e-Evolution (approved):
Is good feature adding as public method

PR #4073, opened 2023-02-09
Add support to UOM conversion for Physical Inventory

yamelsenih:
Just add support to UOM conversion to Physical Inventory like Sales Order
Screenshot from 2023-02-09 18-09-03

### DIFF
--- a/base/src/org/spin/model/CalloutInventory.java
+++ b/base/src/org/spin/model/CalloutInventory.java
@@ -1,0 +1,231 @@
+/******************************************************************************
+ * Product: Adempiere ERP & CRM Smart Business Solution                       *
+ * Copyright (C) 1999-2006 ComPiere, Inc. All Rights Reserved.                *
+ * This program is free software; you can redistribute it and/or modify it    *
+ * under the terms version 2 of the GNU General Public License as published   *
+ * by the Free Software Foundation. This program is distributed in the hope   *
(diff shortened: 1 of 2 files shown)


e-Evolution:
@yamelsenih Hi Yamel ,  Inventory management in another unit of measure is a good feature but it should be implemented in the same way as it now works in the sales order, purchase order, receipt or delivery of material.

The quantity is always entered in the QtyEntered field and the result of the conversion would be in this case in the quantity counted.

I can see in the attached image that it is not implemented in the same way.

The quantity count or calculate not should be updatable as in other window that implement the same functionality

image

yamelsenih:
> @yamelsenih Hi Yamel , Inventory management in another unit of measure is a good feature but it should be implemented in the same way as it now works in the sales order, purchase order, receipt or delivery of material.
> 
> The quantity is always entered in the QtyEntered field and the result of the conversion would be in this case in the quantity counted.
> 
> I can see in the attached image that it is not implemented in the same way.
> 
> The quantity count or calculate not should be updatable as in other window that implement the same functionality
> 
> image

Hello @e-Evolution I change the field distribution and visualization:
Screenshot from 2023-11-01 18-19-49

Here a little video with the functionality

https://github.com/adempiere/adempiere/assets/2333092/325731c7-9154-4628-8943-075a2eb3a3b6

yamelsenih:
This pull request is waiting for improve this functionality: https://github.com/adempiere/adempiere/issues/4193

PR #4074, opened 2023-02-09
Add Import Requisition functionality

yamelsenih:
This process allows import requisitions based on standard importer
Screenshot from 2023-02-09 18-35-34
Screenshot from 2023-02-09 18-35-51

## Testing
All need files are here:
- `data/import/Example_Requisition_Format.xlsx`: Complete excel format
- `data/import/Example_Requisitions.tsv` Tab separed Values format for import only requisitions and products
- `data/import/Example_Requisitions_And_Products.tsv`: Tab separed Values format for import only requisitions
- [Online File](https://docs.google.com/spreadsheets/d/1_mxnUNK7glh1LvQsWqzQ1PUE55OEUfNc/edit?usp=sharing&ouid=112500014010442521967&rtpof=true&sd=true)

## Some Data
Note that this pull request have two XML's:
- `09970_Add_Import_Requisition.xml`
- `09980_Import_Requisitions_Add_Import_Formats.xml`

After import, you should see two import formats:
- `Requisitions`: This should match with `data/import/Example_Requisitions.tsv` format
- `Requisitions - Create Products`: This should match with  `data/import/Example_Requisitions_And_Products.tsv`


e-Evolution:
@yamelsenih  The import processes are usually quite slow, since they are processed sequentially.

For this reason it is recommended to implement the import in parallel by document number, making the process take minutes instead of hours.

You can change this import process to be done in parallel

Here is an example of how I implemented it in the import of Orders.

https://github.com/adempiere/adempiere/blob/5fc84c4f87de754c6b2cc715e4d5c779c387ad5f/base/src/org/compiere/process/ImportOrder.java#L228

yamelsenih:
> @yamelsenih The import processes are usually quite slow, since they are processed sequentially.
> 
> For this reason it is recommended to implement the import in parallel by document number, making the process take minutes instead of hours.
> 
> You can change this import process to be done in parallel
> 
> Here is an example of how I implemented it in the import of Orders.
> 
> https://github.com/adempiere/adempiere/blob/5fc84c4f87de754c6b2cc715e4d5c779c387ad5f/base/src/org/compiere/process/ImportOrder.java#L228

Hi @e-Evolution now this import process have `parallelStream` processing

Please review it

e-Evolution:
my testing 

image

image

Import window

image

The import process , I Think that should be entable the Draft status 

image

image

Adding Location to Organization Info

image

Error when try calculate the Tax

image


Error 
image

Reviewing the code I realize that to define the tax of the line it requires that the organization have a location, this is solved by adding a location in the organization's information, on the other hand the tax machine looks for the location of the business partner to determine the tax, in the case of these tests there is no business partner and therefore there is no address of the business partner and the tax cannot be calculated.

This behavior makes the business partner mandatory

I will carry out a test by placing a business partner with an address

e-Evolution:
I fixed the import by placing these two lines of code:

`
int taxId = Env.getContextAsInt(getCtx(),"#C_Tax_ID");
requisitionLine.setC_Tax_ID(taxId);
`

I think the most appropriate thing is to add the location of the partner and with this the tax machine can calculate the tax based on the category of the product and business partner if it exists.

e-Evolution:
@yamelsenih Hello Yamel, reviewing when I implemented the tax within the requisition I realized that the location of the business partner is crucial for the tax calculation machine, I decided to provide support and add this to the requisition line, this should facilitate the implementation of the tax in importing requisitions, try it and I hope this helps your purpose

https://github.com/adempiere/adempiere/pull/4180

e-Evolution (approved):
The changes mas implemented and now also is compatible with Max

PR #4075, opened 2023-02-09
RFQ Create PO Fixed error with unused parameter

yamelsenih:
Currently exists a document type as parameter when a Purchase Order is
created from a Request for Quotation but it is unused inside process.
image

### DIFF
--- a/base/src/org/compiere/process/RfQCreatePO.java
+++ b/base/src/org/compiere/process/RfQCreatePO.java
@@ -169,7 +169,10 @@ public class RfQCreatePO extends SvrProcess
 				{
 					order = new MOrder (getCtx(), 0, get_TrxName());
 					order.setIsSOTrx(false);
-					order.setC_DocTypeTarget_ID();
+					if (p_C_DocType_ID != 0)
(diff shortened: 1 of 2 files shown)


OFB-Review1 (commented):
When analyzing the code, it is seen that this validation is found in the FOR with the comment 'Winner for whole RfQ' line 100, but not in the FOR 'Selected Winner on Line Level' line 172.

When carrying out tests, it is evident that the document type parameter is not used. But adding the modification solves the problem and the order is created with the selected document type.

e-Evolution:
Hi @yamelsenih The code is ok , but I think is a good moment to refactor , create the Abstract process and use the getParameterName in generar the old process wrote old wat is good reactory.

Can please to make the Refactor?

ScolomaM:
@e-Evolution podría explicar un poco mas el requerimiento solicitado

yamelsenih:
Hello @e-Evolution the pull request was updated

e-Evolution:
@yamelsenih thank you to refactor

PR #4076, opened 2023-02-09
Add Columns to Import Payment

yamelsenih:
Just add **Conversion Type** and **Description** columns for import payment

Screenshot from 2023-02-09 19-20-13

## Testing
All need files are here:
- `data/import/Example_Payment.xlsx`: Complete excel format
- `data/import/Example_Payment.tsv` Tab separed Values format for import payments
- [Online File](https://docs.google.com/spreadsheets/d/1GeXVyuiwTbJGC2wSDTBqvwX5dM0ae-mF/edit?usp=sharing&ouid=112500014010442521967&rtpof=true&sd=true)

## Some Data
Note that this pull request have two XML's:
- `10320_Add_Payment_Import_Format.xml`

After import, you should see one import format:
- `Payments - Currency Type`: This should match with `data/import/Example_Payment.tsv` format


yamelsenih:
This is a simple change

e-Evolution:
@yamelsenih , please can change the WHERE constants ie I_IsImported='N' To I_IsImported=? and use the parameters, it is important to leave that practice behind since using ' in SQL leaves it open to SQL injection and executing a query more slowly, the reason is that the query optimizer cannot make a preparatestatement and search the appropriate index, in a few records there is no difference but when talking about a lot of data the seconds matter.

kind regards
Victor

yamelsenih:
> @yamelsenih , please can change the WHERE constants ie I_IsImported='N' To I_IsImported=? and use the parameters, it is important to leave that practice behind since using ' in SQL leaves it open to SQL injection and executing a query more slowly, the reason is that the query optimizer cannot make a preparatestatement and search the appropriate index, in a few records there is no difference but when talking about a lot of data the seconds matter.
> 
> kind regards Victor

Hello @e-Evolution please check it

e-Evolution:
Testing Import Payment 

image

Import Process

image

Import working

image

e-Evolution (approved):
The import Payment working

PR #4077, opened 2023-02-10
Add validation for maximum 20 digits of account

yamelsenih:
Add validation for maximum 20 digits of account

Currently the bank account number has a size of 20 characters, this is
created by database.

This pull request prevent the maximum characters when a bank account is
created based on model class


e-Evolution:
@yamelsenih Why not just leave the constraint , to the length you require , in this way the DBA defines it.

In the case of some country it may be that the length is greater, and even when you change the size in the database it will not allow saving it.

It seems to me that this is more of a customization for a specific customer

PR #4078, opened 2023-02-10
Remove firstOnly method for search account

yamelsenih:
Just remove a method that throw a exception when exists more than one account.

This is very molest and should not throw a exception to end user without a reason know

### DIFF
--- a/base/src/org/compiere/model/MAccount.java
+++ b/base/src/org/compiere/model/MAccount.java
@@ -304,7 +304,7 @@ public class MAccount extends X_C_ValidCombination
 										.setClient_ID()
 										.setParameters(params)
 										.setOnlyActiveRecords(true)
-										.firstOnly();
+										.first();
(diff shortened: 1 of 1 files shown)


e-Evolution:
@yamelsenih Yamel This problem is caused by some document that is generating duplicate combinations for the same parameters, which should never happen, so removing the validation does not solve the problem, since this can cause other problems in accounting or report generation. since the system has the premise that there is only one combination for the same parameters

This happened to me when the code is handling the transaction incorrectly, creating a combination outside of the transaction when accounting is immediate.

I recommend that you create a unique index for the combinations table and solve the problem with the bad code that handles transactions, which is where the real problem is.

I do not recommend removing this control that validates that the combinations are unique. If something is duplicating them, your accounting is trying to be done outside of a transaction, correct your code that causes it and do not remove the control.

PR #4079, opened 2023-02-10
Fixed error with validate payment amount

yamelsenih:
This pull request resolve a problem when a payment is in progress or invalid and it is included in a validation of payment selection.

### DIFF
--- a/base/src/org/compiere/model/MPaySelection.java
+++ b/base/src/org/compiere/model/MPaySelection.java
@@ -685,7 +685,7 @@ public class MPaySelection extends X_C_PaySelection implements DocAction, DocOpt
 				+ "(CASE "
 				+ "		WHEN COALESCE("
 				+ "					SUM(CASE "
-				+ "							WHEN p.DocStatus NOT IN('VO', 'RE') "
+				+ "							WHEN p.DocStatus IN('CO', 'CL') "
 				+ "							THEN 1 "
 				+ "							ELSE 0 "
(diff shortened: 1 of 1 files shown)


OFB-Review1:
Please provide more information on the error that is being solved, to replicate and check that it is correct.

e-Evolution:
@yamelsenih , Hola Yamel I have reviewed the code and it seems to me that it is correctly implemented, but on the other hand, when you proposed the template to raise issues, the idea was to be clear about how to replicate the problem by a functional consultant. With this pull request format, there is no way for a consultant to functional test, I think we should follow the practice of creating the issue before the pull request so that the community functional can test correctly

RamiroV13:
Hi Yamel, reviewing the code, it accepted any payment that was not in reference 'VO' or 'RE' and it was modified to only accept payments in references 'CO' and 'CL', however we do not have more information to validate functionally

yamelsenih:
Hello everybody, the main problem can be decribed with the follow example:

I have three payments:

Document No | Amount | Document Status
-- | --: | --
P-001 | 100 | Completed
P-002 | 200 | Draft
P-003 | 300 | Invalid

The validation for complete payment should not be based on documents (P-002 and P-003) because the document status is invalid.


This pull request fixed the validation and allows that the document to use are **complete** or **closed**

e-Evolution:
@yamelsenih , I review the code and approach is ok

PR #4080, opened 2023-02-12
Allows overwrite overpayment amount a invoice from payment selection

yamelsenih:
Currently a payment allows overwrite a invoice, but, a payment selection don't allow it.

This change just allows overwrite a invoice from payment selection like is allowed from payment

### DIFF
--- a/base/src/org/compiere/model/MPaySelectionLine.java
+++ b/base/src/org/compiere/model/MPaySelectionLine.java
@@ -370,8 +370,8 @@ public class MPaySelectionLine extends X_C_PaySelectionLine
 		if(getC_BPartner_ID() == 0)
 			throw new AdempiereException("@C_BPartner_ID@ @NotFound@");
 
-		if (getC_Invoice_ID() > 0 && getOpenAmt().subtract(getPayAmt()).subtract(getDiscountAmt()).signum() < 0 )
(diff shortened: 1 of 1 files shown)


e-Evolution:
@yamelsenih Yamel allowing a provider to overpay is a terrible practice, since it opens the opportunity for possible fraud.

I think if anyone has this requirement it should be enabled from a configuration variable using system config , and it should be disabled by default.

In particular, I do not recommend opening doors that add risks, especially when it comes to money.

e-Evolution:
> > @yamelsenih Yamel allowing a provider to overpay is a terrible practice, since it opens the opportunity for possible fraud.
> > I think if anyone has this requirement it should be enabled from a configuration variable using system config , and it should be disabled by default.
> > In particular, I do not recommend opening doors that add risks, especially when it comes to money.
> 
> Hi @e-Evolution I don't agree with it, you can see that the **Payment** window allows it.
> 
> The payment selection allows define a approval workflow for it if you want take more control about it.

@yamelsenih Yamel exactly, if you want, then to do that you can make a direct payment, in the payment it allows under payment or overpayment. but the payment selection was never  allowed any  overpayment, changing this unilaterally does not consider customers who already know that you cannot issue a payment for an amount greater than the amount of the invoice.

In addition, this allow the possibility of any fraud since it allows you to make a payment greater than necessary. if this feature is enable then is simple empty your bank account

In any case, I don't see a use case, where you need to make an overpayment is not natural, can you explain your use case?

PR #4082, opened 2023-02-12
Fixed error with element value insert for tree

yamelsenih:
When you have many element values like User1, User2 and account schema for a same client, the tree node used for insert it is wrong because is not search with a correct criteria.

This pull request resolve it


yamelsenih:
Hello, maybe someone can review this error

e-Evolution (approved):
Create new Tree for User 01 and User 02

image

image

Create new User 01 Dimension 

image

Create new User 02 Dimension 

image

Setting new dimension into account schema

image


image


Adding Element to new dimension 

image


image

The account element not is modify

the mix is working

PR #4085, opened 2023-03-02
Fixed error with missing property file for apache ecs packages

yamelsenih:
Just add the file `ecs.properties` is missing after publish packages, this throw a error when you try run a report with html format

The error is showed in terminal like this:

```Shell
java.lang.NoClassDefFoundError: Could not initialize class
org.adempiere.legacy.apache.ecs.ECSDefaults

Cannot find org.adempiere.legacy.apache.ecs.ecs.properties.
```

### DIFF
--- a/tools/build.gradle
+++ b/tools/build.gradle
@@ -12,8 +12,12 @@ dependencies {
 sourceSets {
     main {
          java {
-            srcDirs = ['src']
+            srcDirs 'src'
          }
+         resources {
+    		srcDirs 'src'
+        	include 'org/adempiere/legacy/apache/ecs/ecs.properties'
+    	 }
     }
 }
 


marcalwestf:
@yamelsenih  can you please document step-wise how to reproduce this bug and the correction?

I tested in https://adqa.prolinux.cl/webui/index.zul , the site sponsored by Horacio.

I could run a report in HTML format :
imagen

Maybe I am testing in the wrong place...

yamelsenih:
Hello @marcalwestf the problem is not in the main adempiere/adempiere project, the problem is that the tools library after publish don't have these files. This error affect to project like:
- https://github.com/adempiere/zk-ui
- https://github.com/adempiere/swing-ui

The change is very very little, you can see that is the gradle definition

PR #4098, opened 2023-04-12
fix: getProductPrice with where clause.

EdwinBetanc0urt:
When using the `getProductPrice` method of the `MPriceListVersion` class:

https://github.com/adempiere/adempiere/blob/0070473977179e538835f46b9b486084307a5d30/base/src/org/compiere/model/MPriceListVersion.java#L116-L127

And passing a where clause that does not start with spaces, for example `AND 1=1` or any other filter:

```sql
AND EXISTS (
  SELECT 1 FROM C_OrderLine ol ol
  WHERE ol.C_Order_ID = 1000035 AND ol.M_Product_ID = M_ProductPrice.M_Product_ID
)
```

 it generates an error:

``` log
org.postgresql.util.PSQLException: ERROR: trailing junk after parameter at or near "$1A"
  Position: 215, SQL=SELECT AD_Client_ID,AD_Org_ID,Created,CreatedBy,IsActive,M_PriceList_Version_ID,M_ProductPrice_ID,M_Product_ID,PriceLimit,PriceList,PriceStd,UUID,Updated,UpdatedBy FROM M_ProductPrice WHERE (M_PriceList_Version_ID=?AND EXISTS(SELECT 1 FROM C_OrderLine ol WHERE ol.C_Order_ID = 1000035 AND ol.M_Product_ID = M_ProductPrice.M_Product_ID) AND EXISTS(SELECT 1 FROM C_OrderLine ol WHERE ol.C_Order_ID = 1000035 AND ol.M_Product_ID = M_ProductPrice.M_Product_ID)) [38]
```
```sql
WHERE M_PriceList_Version_ID=?AND 
```

To prevent this error, you can send the parameter with the where clause with the string starting with a blank space, but this may be impractical, and it also duplicates the where clause sent as a parameter:

```sql
SELECT AD_Client_ID,AD_Org_ID,Created,CreatedBy,IsActive,M_PriceList_Version_ID,M_ProductPrice_ID,M_Product_ID,PriceLimit,PriceList,PriceStd,UUID,Updated,UpdatedBy 
FROM M_ProductPrice WHERE (M_PriceList_Version_ID=?
AND EXISTS(SELECT 1 FROM C_OrderLine ol WHERE ol.C_Order_ID = 1000035 AND ol.M_Product_ID = M_ProductPrice.M_Product_ID) 
AND EXISTS(SELECT 1 FROM C_OrderLine ol WHERE ol.C_Order_ID = 1000035 AND ol.M_Product_ID = M_ProductPrice.M_Product_ID))
```

### DIFF
--- a/base/src/org/compiere/model/MPriceListVersion.java
+++ b/base/src/org/compiere/model/MPriceListVersion.java
@@ -24,6 +24,7 @@ import org.adempiere.core.domains.models.I_M_ProductPrice;
 import org.adempiere.core.domains.models.X_M_PriceList_Version;
 import org.compiere.util.DisplayType;
 import org.compiere.util.TimeUtil;
+import org.compiere.util.Util;
 
 /**
  *	Price List Version Model
@@ -115,9 +116,10 @@ public class MPriceListVersion extends X_M_PriceList_Version
 	 */
 	public MProductPrice[] getProductPrice (String whereClause)
 	{
-		String localWhereClause = I_M_ProductPrice.COLUMNNAME_M_PriceList_Version_ID+"=?"+whereClause;
-		if (whereClause != null)
+		String localWhereClause = I_M_ProductPrice.COLUMNNAME_M_PriceList_Version_ID + "=?";
+		if (!Util.isEmpty(whereClause, true)) {
 			localWhereClause += " " + whereClause;
+		}
 		List<MProductPrice> list = new Query(getCtx(),I_M_ProductPrice.Table_Name,localWhereClause,get_TrxName())
 			.setParameters(getM_PriceList_Version_ID())
 			.list();
--- a/base/src/org/compiere/model/MPriceListVersion.java
+++ b/base/src/org/compiere/model/MPriceListVersion.java
@@ -118,6 +118,9 @@ public class MPriceListVersion extends X_M_PriceList_Version
 	{
 		String localWhereClause = I_M_ProductPrice.COLUMNNAME_M_PriceList_Version_ID + "=?";
 		if (!Util.isEmpty(whereClause, true)) {
+			if (!whereClause.trim().startsWith("AND")) {
+				localWhereClause += " AND ";
+			}
 			localWhereClause += " " + whereClause;
 		}
 		List<MProductPrice> list = new Query(getCtx(),I_M_ProductPrice.Table_Name,localWhereClause,get_TrxName())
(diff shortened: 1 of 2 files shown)


EdwinBetanc0urt:
Unit tests were added for the `MPriceListVersion.getProductPrice` method, however it seems that the continuous integration with gradle is not considering the tests.


https://github.com/EdwinBetanc0urt/adempiere/actions?query=branch%3Afeature%2Fadd-unit-test+

imagen

rcapecce:
Testing this functionality, it is found that this method is effectively duplicating the subquery, affecting the performance of the method.

Additionally you can see that for this to work correctly the call has to be done with a space before it when the additional query is added.

image

This funcionality was tested with the "Copiar Precio a Costo Estandar" process.

With this PR the query does not repeat the query

image

This PR was also tested with the vue backend functionality in POS noting that it applies the same solution.

rcapecce (approved):
Testing this functionality, it is found that this method is effectively duplicating the subquery, affecting the performance of the method.

Additionally you can see that for this to work correctly the call has to be done with a space before it when the additional query is added.

image

This funcionality was tested with the "Copiar Precio a Costo Estandar" process.

With this PR the query does not repeat the query

image

This PR was also tested with the vue backend functionality in POS noting that it applies the same solution.

marcalwestf:
@rcapecce  did you test and accept this PR?
I didn't understand well your answer.

marcalwestf (commented):
The code changes seem OK.
Now a detailed description to test it would be helpful

marcalwestf:
After conversation with @EdwinBetanc0urt , he explained that in cases like in
https://github.com/adempiere/adempiere/blob/032c953e491782603120c06e372e69b8f2d98ee7/org.eevolution.manufacturing/src/main/java/base/org/eevolution/manufacturing/process/CopyPriceToStandard.java#L58

Where in a process code for example the following is found:
`Arrays.stream(priceListVersion.getProductPrice(" AND " + MProductPrice.COLUMNNAME_PriceStd + " <> 0")).forEach(productPrice -> {`

if the code is changed to (no blank before **AND**)
`...."AND " + MProductPrice.COLUMNNAME_PriceStd + " <> 0")).forEach(productPrice -> {...`

then with the actual implementation, an error is produced (besides the duplication of the query as exposed by @rcapecce ).

marcalwestf (approved):
Code accepted after explanations.
Also, a complete ANT build was successful.

PR #4107, opened 2023-06-12
Default charge on bank statement

### LINKED ISSUE #4106: [Bug Report] Default charge on bank statement
When performing the account statement reconciliation, the process automatically assigns a default charge to the import lines that do not have a payment/receipt.

In addition, this charge is assigned by obtaining the MAX (C_Charge_ID), instead of asking for some check or parameterization.

However, it is not correct to automatically assign a charge to lines without payment.

#### Screenshot or Gif

Example

#### Expected behavior

The reconciliation process should not assign a charge automatically

nsarlabos:
Fixes #4106

### DIFF
--- a/client/src/org/spin/apps/form/BankStatementMatchController.java
+++ b/client/src/org/spin/apps/form/BankStatementMatchController.java
@@ -882,10 +882,7 @@ public class BankStatementMatchController {
 	public String saveData(int m_WindowNo, String trxName) {
 		AtomicInteger processed = new AtomicInteger();
 		AtomicInteger lineNo = new AtomicInteger(10);
-		int defaultChargeId = DB.getSQLValue(null, "SELECT MAX(C_Charge_ID) FROM C_Charge WHERE AD_Client_ID = ?", Env.getAD_Client_ID(Env.getCtx()));
-		if(defaultChargeId <= 0) {
-			return Msg.parseTranslation(Env.getCtx(), "@C_Charge_ID@ @NotFound@");
-		}
+
 		importedPaymentHashMap
 			.entrySet()
 			.stream()
@@ -898,10 +895,6 @@ public class BankStatementMatchController {
 				//	For Bank Statement
 				if(bankStatement != null
 						&& !bankStatement.isProcessed()) {
-					if(currentBankStatementImport.getC_Payment_ID() <= 0
-							&& currentBankStatementImport.getC_Charge_ID() <= 0) {
-						currentBankStatementImport.setC_Charge_ID(defaultChargeId);
-					}
 					if(currentBankStatementImport.getC_BankStatementLine_ID() == 0) {
 						importMatched(currentBankStatementImport, lineNo.get());
 						lineNo.addAndGet(10);


marcalwestf:
@nsarlabos  a question: looking at the code, it searches for the Charge with the highest ID.
Mostly, there are charges in every client, so one charge will be at least there.
I wonder if this is correct. You would get **any** charge, not the one you may wish.
```
int defaultChargeId = DB.getSQLValue(null, "SELECT MAX(C_Charge_ID) FROM C_Charge WHERE AD_Client_ID = ?", Env.getAD_Client_ID(Env.getCtx()));
if(defaultChargeId <= 0) {
  return Msg.parseTranslation(Env.getCtx(), "@C_Charge_ID@ @NotFound@");
}
```

Perhaps there is a reason.
Thanks!

nsarlabos:
Hi @marcalwestf 

Sure, that's how it's done right now, but we understand that it's wrong for a charge to be assigned this way, especially if it's the first charge found.

yamelsenih:
I don't sure that this solution is the right. The bank statement line model have a validation for charge if the payment is not setted here.

A option is add the charge as form parameter

yamelsenih:

yamelsenih:
Hello @nsarlabos Already exists a Smart browser to create Payment from Bank Statement and these process allows define a charge. The bank statement match just make a match between payment and bank statement line.

I think that this pull request not apply

niclopup:
> Hello @nsarlabos Already exists a Smart browser to create Payment from Bank Statement and these process allows define a charge. The bank statement match just make a match between payment and bank statement line.
> 
>  Create_Payment_From_BankStatement.mp4 
> I think that this pull request not apply

@yamelsenih I think that the process that you refer in the SB "create Payment from Bank Statement" is after the BankStatementLines are created from the "Bank Statement Match".

The problem thaw we found is when the lines of the Bank Statement are created from the "Bank Statement Match". They are created with a Charge that it is wrong. First of all because it is no more needed there and second, the charge must be defined in the payment that would be created after, for example in the process of the SB that you refer "create Payment from Bank Statement".

The only thing that we made here is to remove the definition of a Charge in the "Bank Statement Line" that it has no porpuse

yamelsenih:
> > Hello @nsarlabos Already exists a Smart browser to create Payment from Bank Statement and these process allows define a charge. The bank statement match just make a match between payment and bank statement line.
> > Create_Payment_From_BankStatement.mp4
> > I think that this pull request not apply
> 
> @yamelsenih I think that the process that you refer in the SB "create Payment from Bank Statement" is after the BankStatementLines are created from the "Bank Statement Match".
> 
> The problem thaw we found is when the lines of the Bank Statement are created from the "Bank Statement Match". They are created with a Charge that it is wrong. First of all because it is no more needed there and second, the charge must be defined in the payment that would be created after, for example in the process of the SB that you refer "create Payment from Bank Statement".
> 
> The only thing that we made here is to remove the definition of a Charge in the "Bank Statement Line" that it has no porpuse

Thanks @nsarlabos let me test it and send you any update

Issue #4000, opened 2022-10-12
[Bug Report] Allows deletion of System client records with the GardenWorld client.


EdwinBetanc0urt:
#### Steps to reproduce

1. Login with the role `GardenWorld Admin` into `GardenWorld` client.
2. Open the window `Country, Region and City`.
3. Delete any records

#### Screenshot or Gif

https://user-images.githubusercontent.com/20288327/195465414-9995df08-10c6-4083-b3e3-8fc9a7d11f2e.mp4

#### Expected behavior
Note that the record is created with the `System` client, and is logged into the `GardenWorld` client, so it should not allow deleting records from another client.

#### Other relevant information
- Your OS: Kubuntu 20.04
- Java version: OpenJDK Temurin-11.0.15+10
- ADempiere version: Develop with commit [34a95f5](https://github.com/adempiere/adempiere/commit/34a95f5fdb7b832689ae0bf9d583fc702732fc45)
- Browser [e.g. chrome, safari]: Mozilla Firefox 
- Browser Version [e.g. 22]: 101.0
- DB Provider: PostgreSQL
- DB version: 14.2

PR #4128, opened 2023-08-13
fix: #4000 Deletion of records from another `Client`.

EdwinBetanc0urt:
https://github.com/adempiere/adempiere/assets/20288327/fef0147e-1f13-4567-a041-66ad06946ff9

fixes https://github.com/adempiere/adempiere/issues/4000

### DIFF
--- a/base/src/org/compiere/model/PO.java
+++ b/base/src/org/compiere/model/PO.java
@@ -118,7 +118,7 @@ import org.w3c.dom.Element;
  *			@see FR [ 729 ] Add Support to Parent Column And Search Column for Tree </a>
  * @author Edwin Betancourt, EdwinBetanc0urt@outlook.com, https://github.com/EdwinBetanc0urt
  *			<a href="https://github.com/adempiere/adempiere/issues/4000">
- *			@see BF [ 729 ] Allows deletion of System client records with the GardenWorld client</a>
+ *			@see BR [ 4000 ] Allows deletion of System client records with the GardenWorld client</a>
  */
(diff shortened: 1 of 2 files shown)


marcalwestf (commented):
Error Replicated: I could delete the country Zimbabwe as GardenAdmin:
image

image

marcalwestf (commented):
Now I tried to delete the same country (Zimbabwe) with the changes of this PR:

image


image

It works as expected.

marcalwestf:
@EdwinBetanc0urt  before accepting this PR, I will check the code and test if the changes don't affect other situations.

For this, I will try to delete normal records and perhaps records by Smart Browsers and processes?

EdwinBetanc0urt:
Hi @marcalwestf I did not test with processes/smart browse, in those cases you should set the `force` parameter to `true`.


https://github.com/adempiere/adempiere/pull/4128/files#diff-1038ae52d64ad525f8ac5aa0f9ac555d754c724fb9e141db5d437d10c96bc618R3050

marcalwestf:
thanks @EdwinBetanc0urt  everything seems to be OK: I just tested deleting several records of the same AD_Client and it works well. The positive test thus works!!

Only to be sure, I am thinking loud if there are situations where Env would not contain the AD_Client_ID.....

marcalwestf (approved):
Functional tests passed. See my posts before.

Code review: variable _Env_ will alway have the AD_Client_ID, since it is set at session begin.

Issue #4138, opened 2023-08-25
Translate Contributions in the Discussions Section into English


marcalwestf:
## Feature request
Translate comments of Discussions into English.
For example, in the Discussion section of this repository, the discussion was started in Spanish, but some participants wish to get it in English (see https://github.com/adempiere/adempiere/discussions/4123).

#### Describe the solution you'd like
The solution would be to include discussions in the responsible GitHub Event Action. 
The file concerned would be the following: https://github.com/adempiere/adempiere/blob/develop/.github/workflows/issue-and-pr-translator.yml .

According to GitHub official documentation about events (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#discussion), the solution would be to add the following code:

```
on:
  ...
  discussion:
    types: [created, edited, answered]
  discussion_comment:
    types: [created, deleted]
  ...
```

PR #4140, opened 2023-08-25
Add translation for Discussions in GitHub

marcalwestf:
Solves #4138

@e-Evolution  Victor, @yamelsenih Yamel: either of you can accept this PR.

Unluckily, being this Pull Request a GitHub Action, this Pull Request can only be tested **after** the Pull Request has been accepted by creating for example a comment to a Discussion in Spanish; it should be translated as it happens now to issues and pull requests.

So, the acceptance has to be done "blindly" and if everything goes well, OK.
Otherwise, the Pull Request has to be reverted; then find the reason for the PR not functioning, fixing it and creating a new PR.

### DIFF
--- a/.github/workflows/issue-and-pr-translator.yml
+++ b/.github/workflows/issue-and-pr-translator.yml
@@ -9,7 +9,11 @@ on:
     types: [opened]
   # Valid for issues and pull requests
   issue_comment: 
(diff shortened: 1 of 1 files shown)


e-Evolution (approved):
good feature, thank you

PR #4157, opened 2023-09-22
fix: Remove deprecated warning with `new Float` constructor.

EdwinBetanc0urt:
`The constructor Float(double) is deprecated since version 9`

`The constructor Float(float) is deprecated since version 9`

imagen


marcalwestf (approved):
Review:
- Check code: no compiler errors.
- Check code: use of _valueOf()_ instead of _New()_
- Compiler: warnings reduced.
- Build: ANT build run without errors.
- Usage: random window and reports calling; creating, modifying and deleting records.

Result: no errors -> **PR accepted**.

PR #4164, opened 2023-09-25
fix: Remove deprecated warning with `DB.prepareStatement` method.

EdwinBetanc0urt:
```log
The method prepareStatement(String, int, int) from the type DB is deprecated
```

```log
The method prepareStatement(String) from the type DB is deprecated
```
prepareStatement deprecated

### DIFF
--- a/base/src/org/compiere/util/DB.java
+++ b/base/src/org/compiere/util/DB.java
@@ -733,21 +733,6 @@ public final class DB
 	}	//	prepareCall
 
 
-	/**************************************************************************
-	 *	Prepare Statement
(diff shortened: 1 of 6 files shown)


marcalwestf (approved):
Review:
- Check code: no compiler errors.
- Check code: use of _DB.prepareStatement(..., null)_ instead of _DB.prepareStatement(...)_
- Compiler: warnings reduced.
- Build: ANT build run without errors.
- Usage: creating, modifying and deleting records.

Result: no errors -> **PR accepted**.

PR #4181, opened 2023-11-03
Fixed error with sequence when a PO is generated from requisition

yamelsenih:
## What is the problem?
When a Purchase order is generated from a requisition using the smart browser `Requisition-to-Invoice` -> `Browser Requisition to Create PO from Requisition Lines`, the document type selected is setted but the sequence is wrong.

## What is the reason?
The smart browser never send the document type to process and the standard process for generate Purchases don't have support to document type.

## What is the change?
This pull request add a new parameter to process `Requisition-to-Invoice` -> `Create PO from Requisition` to add support to document type.

## Step for Step

- Create a new Requisition
- Go to Smart Browser `Requisition-to-Invoice` -> `Browser Requisition to Create PO from Requisition Lines`
- Select a Document Type distinct to default document type of PO
- Process it

Note that after process, the document type inside order is the selected but the **Document No** for Purchase Order is the sequence of default document type instead the sequence of selected document type


yamelsenih:
Replaced by: https://github.com/adempiere/adempiere/pull/4182

PR #4190, opened 2023-11-13
Bugfix #3933

### LINKED ISSUE #3933: Payment rule in sales order with sub type "Invoice Order"
When a sales order with sub document type "Invoice Order" is created, the payment rule is not automatically set to "Cash", as it happens with the sub type "POS Order".

PaymentRuleOrder

#### Expected behavior
When selecting a type of sales order that has the sub type "Invoice Order" the payment rule must also be set to "Cash".

nsarlabos:
Fixes #3933


nsarlabos:
Fixes #3933

e-Evolution:
@nsarlabos thank a lot , Chat with @rcapecce Raul can see the problem about the mix end Lines that exist in some java code , we Think that should apply the change for all files using Linux end code line

PR #4198, opened 2023-11-20
Bugfix/3924 notification created in scheduler run

### LINKED ISSUE #3924: Notification created of each process can't be deactivate

rcapecce:
Fix #3924

### DIFF
--- a/base/src/org/compiere/model/MScheduler.java
+++ b/base/src/org/compiere/model/MScheduler.java
@@ -41,6 +41,9 @@ import org.compiere.util.Msg;
  *  
  *  Contributors:
  *    Carlos Ruiz - globalqss - FR [3135351] - Enable Scheduler for buttons
+ *  @author Raul Capecce, raul.capecce@openupsolutions.com, OpenupSolutions https://openupsolutions.com
+ *			<a href="https://github.com/adempiere/adempiere/issues/3924">
+ * 			@see FR [ 3924 ] Notification created of each process can't be deactivate</a>
  */
 public class MScheduler extends X_AD_Scheduler
 	implements AdempiereProcessor, AdempiereProcessor2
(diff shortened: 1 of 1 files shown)


yamelsenih:
Hello @rcapecce maybe you can resolve this build error: https://github.com/adempiere/adempiere/actions/runs/6935572892/job/18865975321#step:4:37

rcapecce:
> Hello @rcapecce maybe you can resolve this build error: https://github.com/adempiere/adempiere/actions/runs/6935572892/job/18865975321#step:4:37

Sure, this was solved @yamelsenih

base/src/org/compiere/model/MProduction.java
```diff
@@ -44,7 +44,6 @@
 import org.compiere.util.Env;
 import org.compiere.util.Msg;
 import org.compiere.util.Util;
-import org.eevolution.manufacturing.model.MPPProductBOMLine;
```
yamelsenih, line 47: Please, this pull request should be atomic. A production class is unnecessary for this scope.

Can you use a fresh branch?

yamelsenih (changes requested):

base/src/org/compiere/model/MProduction.java
rcapecce: Sure, I checkout master and pick this commits again without this change

yamelsenih (approved):

PR #4215, opened 2023-11-21
Correct XML sequences before 3.9.4.001

marcalwestf:
The XML numbers must follow the last sequence from 3.9.4


marcalwestf:
aborted because it pointed to wrong branch!

PR #4216, opened 2023-11-22
Feature/3.9.4.001 correct xml sequences

marcalwestf:
Previous to Release 3.9.4.001


e-Evolution:
@marcalwestf I Think that this PR is incorrect because should start from branch hotfix/3.9.4.001 vs develop

marcalwestf:
right @e-Evolution 
aborted because it pointed to wrong branch!

PR #4220, opened 2023-11-24
The Generate Outbound (Sales Order/Distribution) the Weight and…

e-Evolution:
… Volumen show incorrect info https://github.com/adempiere/adempiere/issues/2921

Add Delivery Rule and Delivery Via
Add the Attribute Set Instance Id
Add Calculate Delivery to to multiples Attribute Set Instance Id

replace this PR  #3338


e-Evolution:
Close incorrect PR

PR #4229, opened 2023-11-27
 fix: Build including the latest version of `xml` migration files. 

EdwinBetanc0urt:
Currently the extending branches of `hotfix` are not including the xml migration files from the `394lts-394.0001` folder, only the `392lts-393lts` and `393lts-394lts` folders are being included.

imagen

So if the continuous integration requires importing the migration xml it will not include the other `394lts-394.0001` files.

imagen

https://github.com/EdwinBetanc0urt/adempiere/actions/runs/7010632974

### Before this changes.

imagen

imagen

https://github.com/EdwinBetanc0urt/adempiere/actions/runs/7010384589


EdwinBetanc0urt:
The build fails in the step of importing the migration xml files because of this file https://github.com/adempiere/adempiere/blob/hotfix/3.9.4.001/migration/394lts-395lts/09990_Fix_Lenght_for_Employee_Validation_Code.xml without this file builds everything correctly

Issue #3884, opened 2022-05-21
[Bug Report] Shipment Detail report does not finish generating


EdwinBetanc0urt:
#### Steps to reproduce
When executing the Shipment Detail report it stays in a never ending execution.

1. Open the Shipment Detail report.
2. Execute without modifying any parameters.

#### Screenshot or Gif

https://user-images.githubusercontent.com/20288327/169669658-b9b284ce-d257-477d-9f8d-b118dea801d3.mp4

#### Expected behavior
It is expected to generate the output of the report, or otherwise display information about the error to the user.

#### Other relevant information
- Your OS: Kubuntu 20.04
- Java version: OpenJDK Temurin-11.0.15+10
- ADempiere version: Develop with commit [436fd0c](https://github.com/adempiere/adempiere/commit/436fd0c2e90e6fc37e89a136e64583389b9a44fe)
- Browser [e.g. chrome, safari]: Mozilla Firefox 
- Browser Version [e.g. 22]: 100.0
- DB Provider: PostgreSQL
- DB version: 14.2

PR #4246, opened 2023-12-01
Fixed error with many executions for report flagged as `IsForm` inside

yamelsenih:
## Bug
The problem only affect to reports inside windows using button **Print Preview** for non standard reports.

## Step by Step
- Go to **Open Items** -> **Payment Selection** window
- Print Report inside window
- Note that the report show all records

## Base setup
The Print format must be flagged as `Form` and must has **Header** and **Content** bands

## Some History
The problem is that a `OR` operator was add [here](https://github.com/EdwinBetanc0urt/adempiere/commit/5d58da08e190e7f4074dc9618f8bb4b32eb15561#diff-e6e4fea8b4088210a4e30432da4d3b4ebee36723ece363e4d96572d119ca5c7eR874) instead a compose `AND` operator.

Note: This behavior not apply for legacy formats like Invoice, Orders, Shipments and other old hardcoded formats

https://github.com/adempiere/adempiere/assets/2333092/4c03a5a1-0cde-4305-b00b-df8fbf5c3dfa

A example of multiple record:
Screenshot from 2023-11-30 20-12-23

image

Related to: https://github.com/adempiere/adempiere/issues/3884

### DIFF
--- a/base/src/org/compiere/print/ReportEngine.java
+++ b/base/src/org/compiere/print/ReportEngine.java
@@ -869,10 +869,7 @@ queued-job-count = 0  (class javax.print.attribute.standard.QueuedJobCount)
 				columnKey = tableName + "_ID";
 
 			query = MQuery.getEqualQuery(columnKey, pi.getRecord_ID());
-		}
-		if(tableName.startsWith("T_")
-				|| (Optional.ofNullable(pi.getAD_PInstance_ID()).isPresent() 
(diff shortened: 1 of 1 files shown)


carlosaparadam:
Hi @yamelsenih, with this change its this problem again https://github.com/adempiere/adempiere/issues/3884

marcalwestf (commented):
Working on Swing:

I could replicate the error with my own, makeshift print format:
Selection_368

PR #4248, opened 2023-12-01
Fixed error with print report from window with parameters

### LINKED ISSUE #4247: [Bug Report] Loading Report from window with parameter
## Bug report
If you try to print a report from any window, it gives an error.

#### Steps to play

- Go to Physical Inventory
- Try to print from the window on the printer button
- See the bug

#### Screenshot or Gif
image

yamelsenih:
Fixes: #4247

### DIFF
--- a/zkwebui/WEB-INF/src/org/adempiere/webui/apps/ProcessPanel.java
+++ b/zkwebui/WEB-INF/src/org/adempiere/webui/apps/ProcessPanel.java
@@ -632,10 +632,7 @@ public class ProcessPanel extends ProcessController implements SmallViewEditable
 					FDialog.error(getWindowNo(), getPanel(), "Error", saveError);
 				}
 			}
-		} else {
-			saveError = saveParameters();
 		}
-			
 		return saveError;
 	}
 	


marcalwestf (commented):
Test on Swing run successfully:
image

I see now, the error was generated on ZK.
The test will be run on ZK.

marcalwestf (commented):
When trying to reproduce the error, executed on
- Installed bare metal ADempiere or launched from Eclipse 2023-12/Tomcat
- Ubuntu 22.04
- ZK
- Firefox or Chrome

This is the result when trying to print a Physical Inventory:
image

Then, when clicking on Print:
image


I cannot reproduce the error in the context I specified.

e-Evolution:
@yamelsenih Hi Yamel 

I think that instead of removing the logic of saving the parameters, the logic of saving the **Last Run** parameters should be revised, which is how it should work.
 
On the other hand @marcalwestf  Mario, did you prove that this change does not affect the execution of normal processes? Because I understand that the execution of normal processes the parameters are saved with the name ** Last Run **, removing this can cause a malfunction for the processes or normal reports.

i.e.

image

e-Evolution:
@yamelsenih Hi Yamel , @marcalwestf Mario 

By applying the steps described in the issue, I was not able to replicate the problem. In my case, it does not give me an error.

When running before the patch I get the following:

On the other hand, it is very curious in which case the fSavedName is null, since in that case I understand that this line of code would be executed

https://github.com/adempiere/adempiere/blob/master/zkwebui/WEB-INF/src/org/adempiere/webui/apps/ProcessPanel.java#L636

@yamelsenih  can please set all steps to replicase the problem?

Issue #4253, opened 2023-12-01
[Bug Report] Duplicated Session from ZK


yamelsenih:
## Bug report
When trying to load the same session twice, an annoying message appears that does not allow the session to be loaded again.

`Please report conditions to your system administrator or in sf tracker 2832968`

#### Steps by Step
1. Load the same user twice in two browsers
2. Login

PR #4254, opened 2023-12-01
Fixed error with duplicated session

yamelsenih:
Fixes: #4253


marcalwestf (commented):
Followed the instructions on 
- Ubuntu 22.04
- ZK
- Adempiere 394

Results when opening two sessions on Firefox, on Chrome or on either:
- The first session is opened as usual
- The first session is opened as usual
- When the first session is selected again, the following is displayed,
image

 but when clicking on the link, the session seems to be a duplicate from the first.

I cannot reproduce the error on my operating system.
I will try it out on Windows.

Maybe @yamelsenih  you can specify operating system and browser name?

e-Evolution:
This bufix is a good approach

e-Evolution:
@marcalwestf This behavior is very complicated to reproduce, on some occasions ZK lost the context of the session and showed that error on the ZK screen, after the changes I made in ZK for the management of context and sessions it has not happened in any of my clients, so I can assume that it was fixed, @yamelsenih Yamel's approach is correct, that is, if the context is lost or it is inconsistent with the session, then it executes a logout to avoid strange behavior.

PR #4265, opened 2023-12-06
Fixed error with commission run detail report

yamelsenih:
## Setup
For this error you should create:
- Two commission definitions

Name | Commission Group | Calculation Basis | Frequency Type | Currency | List Details | Days due from Payment Term | Commission only specified Orders | Is Percentage | Multiplier Amount
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
Commission 1 | Test Commision Group | Invoice | Monthly | USD | Yes | No | Yes | Yes | 7
Commission 2 | Test Commision Group | Invoice | Monthly | USD | Yes | No | Yes | Yes | 3

- One commission group

Value | Name
-- | --
TCG | Test Commision Group

## Step to Reproduce

After it you should create two commision run:

Document No | Commission | Commission Group
-- | -- | --
TC-1 | Commission 1 |  
TC-2 |   | Test Commision Group

Now just run the **Quote-to-Invoice** -> **Sales and Marketing** -> **Commission Run Detail** report
Screenshot from 2023-12-06 11-13-45

You can run it without parameters
Screenshot from 2023-12-06 11-13-57

Note that the result is that the commission `TC-01` has data but the `TC-2` is not showed even though it should be shown.

## What is the problem? 
Is very easy, the problem is that the view make a mandatory join with `C_Commission` using the `C_CommissionRun.C_Commission_ID` link column instead (`C_CommissionAmt.C_CommissionLine_ID` -> `C_CommissionLine.C_Commission_ID`)

This pull request just add these changes to:

- `db/ddlutils/postgresql/views/RV_COMMISSIONRUNDETAIL.sql`
- `db/ddlutils/oracle/views/RV_COMMISSIONRUNDETAIL.sql`

Also add a XML to apply inside ADempiere dictionary
- `migration/394lts-3.9.4.002/10330_Fixed_Commission_Run_Detail_View.xml`

### DIFF
--- a/db/ddlutils/oracle/views/RV_COMMISSIONRUNDETAIL.sql
+++ b/db/ddlutils/oracle/views/RV_COMMISSIONRUNDETAIL.sql
@@ -34,8 +34,9 @@ SELECT cr.AD_Client_ID, cr.AD_Org_ID, cr.IsActive, cr.Created,cr.CreatedBy, cr.U
     COALESCE (i.AD_User_ID,o.AD_User_ID) AS AD_User_ID,
     COALESCE (i.C_DocType_ID,o.C_DocType_ID) AS C_DocType_ID
 FROM C_CommissionRun cr
-    INNER JOIN C_Commission c ON (cr.C_Commission_ID=c.C_Commission_ID)
     INNER JOIN C_CommissionAmt ca ON (cr.C_CommissionRun_ID=ca.C_CommissionRun_ID)
+    INNER JOIN C_CommissionLine cl ON (cl.C_CommissionLine_ID=ca.C_CommissionLine_ID)
+    INNER JOIN C_Commission c ON (c.C_Commission_ID=cl.C_Commission_ID)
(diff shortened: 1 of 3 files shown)


marcalwestf (commented):
**First Step: reproduce the error**
I made sure there were some invoices completed for the period to be tested.
Also, I made sure the invoiced referenced to the same SalesRep (GardenAdmin).

Followed the instructions and created 
- Two Commission Definitions for the same SalesRep (GardenAdmin) which differ only on the commission %:
image
image


- A _Commission Run_ for the group and executed the action _Prepare_:
image
image

- A _Commission Run_ for the commission and executed the action _Prepare_:
image
image

- Run report _Commission Run Detail_:
image


Result: the report is missing the other commission altogether.

**The error could be reproduced.**

marcalwestf (approved):
**Reproducing the Bug Fix**

1. Successfully apply XML _10330_Fixed_Commission_Run_Detail_View.xml_
2. Run report _Commission Run Detail_ without parameters:
image

Now the report displays both commission definitions; the error was solved.

Tests passed.

marcalwestf:
This pull request has been tested and accepted; it can be merged, but it is waiting for older pull requests to be merged.

PR #4267, opened 2023-12-07
Bug Fix Missing trx in workflow query outdated PO data

### LINKED ISSUE #4266: [Bug Report] PO outdated in workflow conditions due to missing transaction
## Bug report

In some cases there is data from the PO that arrives outdated to the evaluation of workflow conditions due to having an outdated trx.

In this case, when passing from the node (DocPrepare) to an Approval node which must meet a certain condition, and the fields involved in the condition were modified in the `prepareIt()` method, this data is not updated when evaluating the conditions, and determine which node to continue to.

(You can put a breakpoint in each code block)

The field in question is `C_DocType_ID`, at the beginning the field has the value "New", but in the `prepareIt()` changes to the value of `C_DocTypeTarget_ID`.

https://github.com/adempiere/adempiere/blob/5fc84c4f87de754c6b2cc715e4d5c779c387ad5f/base/src/org/compiere/model/MOrder.java#L1304

For the example, we have a node whose condition is to evaluate that the C_DocType has a certain value, as in the `prepareIt()` was updated it should consult the correct value (e.g. "Invoice Order" instead of "New")

The PO is saved here
https://github.com/adempiere/adempiere/blob/5fc84c4f87de754c6b2cc715e4d5c779c387ad5f/base/src/org/compiere/wf/MWFActivity.java#L885

And the save of the transaction is here
https://github.com/adempiere/adempiere/blob/5fc84c4f87de754c6b2cc715e4d5c779c387ad5f/base/src/org/compiere/wf/MWFActivity.java#L782

The problem is in this line when you get the PO to evaluate the data already updated in the `prepareIt()`, because you got it without trx and it caused you to have the data that you had when your PO were in Draft

https://github.com/adempiere/adempiere/blob/5fc84c4f87de754c6b2cc715e4d5c779c387ad5f/base/src/org/compiere/wf/MWFNodeNext.java#L158

#### Steps to reproduce

1. Start in a seed database
2. Modify the Process_Order Workflow
2.1. Create an approval node as show in the pic
image
2.2. With the approval user as show in the pic
image
2.3. And transitioning to the Complete node
image
2.4. In the `(DocPrepare)` Node, point at `ApproveCO` node.
2.5 Add the next condition -> `C_DocType_ID = 50038` the "Invoice Order" Document Type
image
3. With the user GardenAdmin create a Sales Order with the "Invoice Order" Document Type
4. Observe how the sales order is completed despite having a flow that should prevent it.

#### Expected behavior

When GardenAdmin complete the order with "Invoice Order" Document Type, the document should be locked for approval in this case by GardenUser.

rcapecce:
Fix #4266

### DIFF
--- a/base/src/org/compiere/wf/MWFProcess.java
+++ b/base/src/org/compiere/wf/MWFProcess.java
@@ -241,6 +241,7 @@ public class MWFProcess extends X_AD_WF_Process
 			m_po = lastPO;
 		
 		//
+		set_TrxName(m_po.get_TrxName());
 		MWFActivity[] activities = getActivities (true, true);	//	requery active
 		String closedState = null;
 		boolean suspended = false;
--- a/base/src/org/compiere/wf/MWFProcess.java
+++ b/base/src/org/compiere/wf/MWFProcess.java
@@ -47,6 +47,9 @@ import org.compiere.util.Util;
  * 				<li>[Bug Report] The workflow engine is not correctly handling transactions when processing documents #3170
  * 				<a href="https://github.com/adempiere/adempiere/issues/3170">
  *  @version $Id: MWFProcess.java,v 1.2 2006/07/30 00:51:05 jjanke Exp $
+ *  @author Raul Capecce, raul.capecce@solopsoftware.com, Solop https://solopsoftware.com/
+ *		<a href="https://github.com/adempiere/adempiere/issues/4266">
+ * 		@see FR [ 4266 ] PO outdated in workflow conditions due to missing transaction</a>
  */
 public class MWFProcess extends X_AD_WF_Process
 {


marcalwestf:
@rcapecce this is an insignificant matter, but I suppose you wanted the commit c4aa1502014951ca90f0eeb67ad8deefbe77d3a7 to be made on class MWFProcess, instead of MOrder ;)

Starting to test your PR.

marcalwestf (approved):
**To reproduce the error**, the steps described in the issue were completed.
1.- An Order was created and it was indeed completed without running through the approval workflow defined:
image


**To reproduce the bug fix**, the steps described in the issue were completed.
1.- After executing the _Complete_ action, the order is now not in the status "Completed", but  "In Progress"
image


2.- The Order has to be accepted by User "GardenUser"
image

3.- Result of Approval:
image

4.- Result at Order: document status became _Approved_ after _In Progress_ and _Completed_ after _Approved:_
image


All tests passed; pull request accepted

marcalwestf:
This pull request has been tested and accepted; it can be merged, but it is waiting for older pull requests to be merged.

base/src/org/compiere/model/MOrder.java
```diff
@@ -80,6 +80,9 @@
  * @author Yamel Senih, ysenih@erpya.com, ERPCyA http://www.erpya.com
  *		<a href="https://github.com/adempiere/adempiere/issues/1455">
  * 		@see FR [ 1455 ] Add Sales Region to Order and Invoice</a>
+ * @author Raul Capecce, raul.capecce@solopsoftware.com, Solop https://solopsoftware.com/
+ *		<a href="https://github.com/adempiere/adempiere/issues/4266">
```
EdwinBetanc0urt: Hi @rcapecce, is there a reason to modify the comment in the `MOrder` class? I think the comment should be moved to the `MWFProcess` class.

EdwinBetanc0urt (changes requested):

e-Evolution:
@rcapecce Hi Raul

I have reviewed the logic and it is not clear to me the sequence of your nodes in your workflow. Given that the transaction is in progress, it is correct that you cannot evaluate the C_DocType_ID field because that value has not yet been committed since it is part of the transaction not completed.

Can set here you flow the your workflow as :

image

Every time a step is processed, the PO of the document should be uploaded before processing, so I consider that the problem is that it is being evaluated against a stale PO, when you mention that the PO was uploaded for a null transaction I do not understand how it can happen if The doc action that executes the process to complete the document adds a transaction so it should not be null and if it is I believe from the MWFActivity call it assigns a transaction, so again it should not be null, if it is that is the problem, we must identify why the document instance is loaded as null.

In any case, your validation in the workflow should be with C_DocTypeTarget_ID which has a value before the node (DocPrepare) is called. I think it is not valid to use C_DocType_ID when the prepareIt() method that sets that value is just being processed.

kind regards
Victor

base/src/org/compiere/model/MOrder.java
rcapecce: oh my bad, I have fixed this

rcapecce:
> @rcapecce this is an insignificant matter, but I suppose you wanted the commit [c4aa150](https://github.com/adempiere/adempiere/commit/c4aa1502014951ca90f0eeb67ad8deefbe77d3a7) to be made on class MWFProcess, instead of MOrder ;)
> 
> Starting to test your PR.

Yes, fixed! Thank you @marcalwestf

rcapecce:
> @rcapecce Hi Raul
> 
> I have reviewed the logic and it is not clear to me the sequence of your nodes in your workflow. Given that the transaction is in progress, it is correct that you cannot evaluate the C_DocType_ID field because that value has not yet been committed since it is part of the transaction not completed.
> 
> Can set here you flow the your workflow as :
> 
> image
> 
> Every time a step is processed, the PO of the document should be uploaded before processing, so I consider that the problem is that it is being evaluated against a stale PO, when you mention that the PO was uploaded for a null transaction I do not understand how it can happen if The doc action that executes the process to complete the document adds a transaction so it should not be null and if it is I believe from the MWFActivity call it assigns a transaction, so again it should not be null, if it is that is the problem, we must identify why the document instance is loaded as null.
> 
> In any case, your validation in the workflow should be with C_DocTypeTarget_ID which has a value before the node (DocPrepare) is called. I think it is not valid to use C_DocType_ID when the prepareIt() method that sets that value is just being processed.
> 
> kind regards Victor

Hi @e-Evolution 

I am not saying that the PO information is not being loaded in each step, what I am commenting here is that only in the section where the conditions of the node to advance are evaluated, that is where the transaction reaches null, only at the time of evaluating the conditions.

Here https://github.com/adempiere/adempiere/blob/5fc84c4f87de754c6b2cc715e4d5c779c387ad5f/base/src/org/compiere/wf/MWFNodeNext.java#L158 when obtains the PO to evaluate, the trx is missing and obtanins a data out of the current transaction.

Then, the action of the prepareIt() node ends and saves the savepoint, but the conditions when loaded the PO from the database without a transaction do not see the data reflected

marcalwestf:
a modification like this should have included a whole workflow test after the actual PR test.

To make sure the workflows weren't damaged, I also tested the following cases.

Prepare, complete and void of the following documents:
- Requisition
- Purchase Order
- Purchase Invoice
- Material Receipt
- Sales Order
- Sales Invoice
- Material Issue

rcapecce:
I understand that the workflow does not apply to processes but to make sure I carried out the following test with the base code and with the PR, obtaining that all the documents were Completed, those that had a flow and those that did not.
The process used was SB Process Batch Orders.

PR #4274, opened 2023-12-12
Improve performance at running payroll process

yamelsenih:
## Improve perfomance to Payroll Process
This pull request just separate each movement saving in distints transactions, this is a little change but very useful to improve database long transaction by payroll process

### DIFF
--- a/org.eevolution.hr_and_payroll/src/main/java/base/org/eevolution/hr/model/MHRProcess.java
+++ b/org.eevolution.hr_and_payroll/src/main/java/base/org/eevolution/hr/model/MHRProcess.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 
 import org.adempiere.core.domains.models.X_HR_Process;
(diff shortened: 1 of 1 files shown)


e-Evolution:
@yamelsenih Hello Yamel!

In a good data model, the smaller the transactions, the better, but on the other hand in ADempiere all documents when completed are atomic, this change would make the payroll processes not atomic, which could cause inconsistency, since the payroll It could fail or not be completed correctly, leaving movements calculated in the middle of the calculation, which can be very risky.

Since this change breaks atomicity, I think we should validate a better implementation, based on the processing flags, although this may be more complicated because it would work based on semaphores and if something fails we have to guarantee that the exception is handled correctly to eliminate calculated moves if something goes wrong.

I think @piracio Horacio could help us in this case to implement the best approach.

I think we should discuss in more depth, I propose tomorrow in the meeting to discuss this important topic.

kind regards
Victor

piracio:
Depends on the case. Internet is full of articles good and bad about the approach.

Here is an example why NOT to use sub-transactions. 
https://about.gitlab.com/blog/2021/09/29/why-we-spent-the-last-month-eliminating-postgresql-subtransactions/

yamelsenih:
Maybe this solution is very little. The real problem is the ADempiere Workflow management.

PR #4275, opened 2023-12-12
Improve In-Memory performance to get `lastConcept` method

yamelsenih:
This pull request just improve the use of `getLastConcept` method to payroll process. Just read the source core and test it with the same use case of payroll.


yamelsenih:
Hello @marcalwestf can you test it?

marcalwestf:
@yamelsenih  yes, we'll test it

SusanneCalderon (approved):
I tested in a running configuration of payroll and included the modified processes in different concept calculations. The performance was improved considerably.

PR #4278, opened 2023-12-12
Fixed error with Import Product definition for `Volume` column

yamelsenih:
Just fixed error with Product Import.

## What is the problem?
The problem is the the `Volume` column was defined as `Integer` instead `Amount`. The problem is that a volume unit is not a integer

### DIFF
--- a/base/src/org/adempiere/core/domains/models/I_I_Product.java
+++ b/base/src/org/adempiere/core/domains/models/I_I_Product.java
@@ -679,12 +679,12 @@ public interface I_I_Product
 	/** Set Shelf Depth.
 	  * Shelf depth required
 	  */
-	public void setShelfDepth (int ShelfDepth);
+	public void setShelfDepth (BigDecimal ShelfDepth);
 
 	/** Get Shelf Depth.
 	  * Shelf depth required
 	  */
-	public int getShelfDepth();
+	public BigDecimal getShelfDepth();
 
     /** Column name ShelfHeight */
     public static final String COLUMNNAME_ShelfHeight = "ShelfHeight";
@@ -692,12 +692,12 @@ public interface I_I_Product
 	/** Set Shelf Height.
 	  * Shelf height required
 	  */
-	public void setShelfHeight (int ShelfHeight);
+	public void setShelfHeight (BigDecimal ShelfHeight);
 
 	/** Get Shelf Height.
 	  * Shelf height required
 	  */
-	public int getShelfHeight();
+	public BigDecimal getShelfHeight();
 
     /** Column name ShelfWidth */
     public static final String COLUMNNAME_ShelfWidth = "ShelfWidth";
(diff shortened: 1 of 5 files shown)


e-Evolution:
@yamelsenih Hello

The problem is that there is no way to define the unit of measurement for the fields Weight, Volume, Shelf Width, Shelf Depth. This would allow establishing complete units for meters, centimeters or pounds. kilos, so the correct thing to do is to add the units of measurement as was done in the vehicle window.

Changing the amount from Integer to Amount does not solve the problem because you will not know if the amount is meters or centimeters. It is better to define the unit of measurement and use the Unit of Measurement conversion tools to obtain more precise data.

The correct approach would be to add the units of measure for each measurement item as in the vehicle window .

* Volume Unit of Measure
* Weight Unit of Measure
* Dimension Unit of Measure

May be

Shelf Unit of Measure

image

kind regards
Victor

yamelsenih:
> @yamelsenih Hello
> 
> The problem is that there is no way to define the unit of measurement for the fields Weight, Volume, Shelf Width, Shelf Depth. This would allow establishing complete units for meters, centimeters or pounds. kilos, so the correct thing to do is to add the units of measurement as was done in the vehicle window.
> 
> Changing the amount from Integer to Amount does not solve the problem because you will not know if the amount is meters or centimeters. It is better to define the unit of measurement and use the Unit of Measurement conversion tools to obtain more precise data.
> 
> The correct approach would be to add the units of measure for each measurement item as in the vehicle window .
> 
> * Volume Unit of Measure
> * Weight Unit of Measure
> * Dimension Unit of Measure
> 
> May be
> 
> Shelf Unit of Measure
> 
> image
> 
> kind regards Victor

Hi @e-Evolution the unit of measurement is not the problem here. Note that the `Volume` column inside `M_Product` was created as `Amout` but the `I_Product` was created as `Integer` then if you want import a product with Volume reference you can't set decimals.

Note: The volume is defined as real number and is a wrong definition assume that these values are integers.

e-Evolution:
@yamelsenih Here I also see another problem, all quantity units are defined with the Quantity reference and not with Amount, Amount is used for monetary units, this can affect why the rounding in Quantity has more precision than that of Amount since this is defined for the currency, so your change should be with the reference Quantity vs Amount

yamelsenih:
> @yamelsenih Here I also see another problem, all quantity units are defined with the Quantity reference and not with Amount, Amount is used for monetary units, this can affect why the rounding in Quantity has more precision than that of Amount since this is defined for the currency, so your change should be with the reference Quantity vs Amount

Hello @e-Evolution I was changed it

yamelsenih:
> > @yamelsenih Here I also see another problem, all quantity units are defined with the Quantity reference and not with Amount, Amount is used for monetary units, this can affect why the rounding in Quantity has more precision than that of Amount since this is defined for the currency, so your change should be with the reference Quantity vs Amount
> 
> Hello @e-Evolution I was changed it

@marcalwestf please test it

e-Evolution:
The change changes from a reference with an integer value to a numerical value that allows decimals, but the solution is incomplete because the unit of measurement of the dimensions cannot be defined.

PR #4285, opened 2024-01-09
fix: SB Process with empty `Record_ID`.

EdwinBetanc0urt:
When `Record_ID` is empty the process runs successfully without sending an exception message in some cases, for example:

https://github.com/adempiere/adempiere/blob/59557cc2ee85ac938cd4f31a246d891bc2b15b8f/base/src/org/compiere/process/InvoiceCreateFrom.java#L61-L63

Some are validated but if the value is -1 it is considered a valid identifier.

More detail is also added to the message to more correctly identify which id is missing.

### DIFF
--- a/base/src/org/adempiere/pipo/PackOut.java
+++ b/base/src/org/adempiere/pipo/PackOut.java
@@ -159,7 +159,7 @@ public class PackOut extends PackOutAbstract {
 		log.info("doIt - AD_PACKAGE_EXP_ID=" + getRecord_ID());
 		// Valid Record Identifier
 		if (getRecord_ID() <= 0) {
-			throw new AdempiereException("@AD_Package_Exp_ID@ (@Record_ID@) @NotFound@");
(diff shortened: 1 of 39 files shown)


marcalwestf:
@EdwinBetanc0urt  one question , Edwin: is this the case for any Smart Browser?

I saw the commit; it is for every Smart Browser; this means every Smart Browser should be tested for approval.
As it maybe not be practical, we can distribute the tests.

EdwinBetanc0urt:
Hi @marcalwestf It is not possible to replicate in `zk`/`swing`, it always sends the record id of the source window to the smart browse process so that the process adds it in the `ProcessBuilder`, it was already corrected in `vue` (from where the error could be identified) that sent the record id in the smart browse, but not to the process, and when the process specifically `base/src/org/compiere/process/InOutCreateFrom.java` was executed it did not give an error, but it did not do anything either, however the real problem was that the user was not informed that a value (`Record_ID`) that is necessary for the execution of the process was missing.

PR #4293, opened 2024-02-15
fix: Resource leak: 'zipFile' is never closed.

EdwinBetanc0urt:
The IDE generates the warning `Resource leak: 'zipFile' is never closed`.

imagen

#### Additional context
To test this method, a packing must be created.

### DIFF
--- a/base/src/org/adempiere/pipo/CreateZipFile.java
+++ b/base/src/org/adempiere/pipo/CreateZipFile.java
@@ -99,22 +99,29 @@ public class CreateZipFile {
             Unzipper.setOwningTarget(new Target());	
             Unzipper.execute();
      }
-	static public String getParentDir(File zipFilepath)
-    {
+
+	static public String getParentDir(File zipFilepath) throws IOException
+	{
+		String fileName = "";
+		ZipFile zipFile = null;
 		try {
-		ZipFile zipFile = new ZipFile(zipFilepath);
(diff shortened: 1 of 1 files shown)


e-Evolution:
Hello Edwin @EdwinBetanc0urt !

Thank you very much for detecting these problems, this makes the project of better quality.

You could use the vavr function library, to ensure the closure of resources, unfortunately is not best approach  so that finally statement does not guarantee a successful closure 


Check here : 
https://docs.vavr.io/#_try

Here an example that make fastly :
```
		Try<String> result = Try.withResources(() -> new ZipFile(zipFilepath))
				.of(zipFile -> {
					Enumeration<? extends ZipEntry> entries = zipFile.entries();
					ZipEntry entry = entries.nextElement();
					File tempfile = new File(entry.getName());
					while (tempfile.getParent()!=null)
						tempfile = tempfile.getParentFile();
					return tempfile.getName();
				});
		result.onSuccess(content -> System.out.println("Contenido del archivo: " + content))
			  .onFailure(error -> System.out.println("Error al leer el archivo: " + error.getMessage()));
```

here you can see how I implemented the other way 
https://github.com/adempiere/adempiere/blob/6ef05ce1d620e95db31c0cb21b4a3b77c247e409/base/src/org/compiere/util/DB.java#L2583

kind regards
Victor

EdwinBetanc0urt:
Hi @e-Evolution the changes you suggested have already been applied.

PR #4294, opened 2024-02-15
fix: Resource leak: 'workbook' is never closed.

EdwinBetanc0urt:
The IDE generates the warning `Resource leak: 'workbook' is never closed`.

imagen

#### Additional context
To test this method, you must export in excel from the `Trial Balance` form in the zk ui.

### DIFF
--- a/zkwebui/WEB-INF/src/org/adempiere/webui/apps/form/WTrialBalance.java
+++ b/zkwebui/WEB-INF/src/org/adempiere/webui/apps/form/WTrialBalance.java
@@ -364,7 +364,7 @@ public class WTrialBalance extends TrialBalanceDrill implements IFormController,
 		getData(data, false);
 		Vector<String> columnNames = getColumnNames();
 
-		ListModelTable modelI = new ListModelTable(data);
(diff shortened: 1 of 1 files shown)


e-Evolution:
Hi @EdwinBetanc0urt 

Thank you very much for detecting these problems, this makes the project of better quality.

You could use the vavr function library, to ensure the closure of resources, unfortunately is not best approach  so that finally statement does not guarantee a successful closure 


Check here : 
https://docs.vavr.io/#_try

Here an example that make fastly :
```
		Try<String> result = Try.withResources(() -> new ZipFile(zipFilepath))
				.of(zipFile -> {
					Enumeration<? extends ZipEntry> entries = zipFile.entries();
					ZipEntry entry = entries.nextElement();
					File tempfile = new File(entry.getName());
					while (tempfile.getParent()!=null)
						tempfile = tempfile.getParentFile();
					return tempfile.getName();
				});
		result.onSuccess(content -> System.out.println("Contenido del archivo: " + content))
			  .onFailure(error -> System.out.println("Error al leer el archivo: " + error.getMessage()));
```

Issue #4332, opened 2024-07-31
Adempiere report that is running when pid file exist


piracio:
## Bug report
After a server crash or java kill -9, the pid file will remain and the current logic is checking if the file exist.
With the current bug Adempiere will NOT start when pid files exist.
#### Steps to reproduce
1. Start Adempiere : ./RUN_Server2.sh
2. kill the java process with -9 :  $kill -9 $(pidof java)
3. Start the Adempiere : ./RUN_Server2.sh 
4. Check the process id reported as Adempiere with : ps -p $PID -o comm=

#### Screenshot or Gif

![image](https://github.com/user-attachments/assets/215eb0a7-58b3-471b-b714-3bb31cef685e)

#### Expected behavior
It is expected to detect that the Adempiere is NOT running and the starting script should start the Adempiere.

#### Other relevant information
- Your OS: OL8
- Java version: "11.0.12" 2021-07-20 LTS
- ADempiere version: AD394_001 with professional patches from EE Support.
- Browser : NA
- Browser Version NA
- DB Provider: PostgreSQL
- DB version: 13.13

#### Additional context
I will work on the patch as soon as possible.

PR #4345, opened 2024-09-05
#4332 Windows Fix JAVA version detected properly, Service FIX tomcat pid when is detected tomcat was not starting

piracio:
FIXED: RUN_setup.bat failed to detect the Java Version.
FIXED: RUN_Server2.sh failed to start Adempiere when Tomcat is used and PID file exist.

### DIFF
--- a/install/Adempiere/RUN_setup.bat
+++ b/install/Adempiere/RUN_setup.bat
@@ -1,7 +1,28 @@
 @Title Install ADempiere Server
 @Echo off
 
-@CALL utils\functions.bat
+@CALL %CD%\utils\functions.bat
+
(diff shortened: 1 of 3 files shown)


EdwinBetanc0urt:
Greetings @piracio, can you describe the steps to replicate the error before this change?

piracio:
Tomcat will not start when the PID file exists. This situation occurs when the tomcat is killed by kill -9 or when the machine is rebooted hard or failed (power off, NOT shutdown), usually when the machine crashes or OOM is triggered (out of memory).

The java version is not detected by run_setup.sh; that is the other improvement.

Issue #4350, opened 2024-09-26
[Bug Report] Default Connections for HikariCP


yamelsenih:
## Bug report
Currently each intance of ADempiere using **HikariCP** use around of 15 connections. 

I finding if exist a way for make a setup for **HikariCP** I look many docs about this pool manager, the follow is a list of some parameters proposed:

 * `IDLE_TIMEOUT`: It sets the maximum time a connection can sit around without being used before it gets closed to free up resources. Default: `300`.
 * `MINIMUM_IDLE`: It sets the minimum number of connections that should be kept open and ready to use, even if they're not currently being used. This helps improve performance by reducing the time it takes to get a connection. Default: `1`.
 * `MAXIMUM_POOL_SIZE`: It sets the maximum number of connections that can be open at the same time. This helps prevent the pool from getting too big and using up too much memory. Default: `10`.
 * `CONNECTION_TIMEOUT`: it sets the maximum time HikariCP will wait to get a connection from the pool before giving up and throwing an error. Default: `5000`.
 * `MAXIMUM_LIFETIME`: It sets the maximum amount of time a connection can stay open in the pool before it's automatically closed. This helps keep the pool clean and prevents problems. Default: `6000`.
 * `KEEPALIVE_TIME`: It sets a test query that HikariCP will run on connections to make sure they're still working properly. Default: `360000`.
 * `CONNECTION_TEST_QUERY`: It sets how often HikariCP will check if a connection is still working properly. This helps prevent problems with connections that might become inactive. Default: `SELECT 1`

#### Steps to reproduce
1. Make a query to database like this: `SELECT COUNT(*) FROM pg_stat_activity;` Note that is nothing
2. Start any ADempiere service using **HikariCP** like Swing or Backend server
3. Make a query to database like this: `SELECT COUNT(*) FROM pg_stat_activity;` Note that is 15 conections

**Before Start**
```SQL
SELECT COUNT(*) FROM pg_stat_activity;
 count 
-------
     6
(1 row)
```
**After Start**
```SQL
SELECT COUNT(*) FROM pg_stat_activity;
 count 
-------
     21
(1 row)
```

#### Additional context
The **HikariCP** is completely configurable but the [DB_PostgreSQL.java](https://github.com/adempiere/adempiere/blob/develop/base/src/org/compiere/db/DB_PostgreSQL.java#L589) is hardcode the values

e-Evolution:
@yamelsenih Hi Yamel 

[...]

PR #4351, opened 2024-09-26
Add parameters from Ini to Database pool connections

yamelsenih:
References: https://github.com/adempiere/adempiere/issues/4350

### DIFF
--- a/base/src/org/compiere/db/DB_PostgreSQL.java
+++ b/base/src/org/compiere/db/DB_PostgreSQL.java
@@ -565,7 +565,70 @@ public class DB_PostgreSQL implements AdempiereDatabase
 		return localConnection;
 	}
 	
-
+	private static final String CONFIG_DB_IDLE_TIMEOUT = "DB|CONFIG_DB_IDLE_TIMEOUT";
+	private static final String CONFIG_DB_MINIMUM_IDLE = "DB|CONFIG_DB_MINIMUM_IDLE";
+	private static final String CONFIG_DB_MAXIMUM_POOL_SIZE = "DB|CONFIG_DB_MAXIMUM_POOL_SIZE";
+	private static final String CONFIG_DB_CONNECTION_TIMEOUT = "DB|CONFIG_DB_CONNECTION_TIMEOUT";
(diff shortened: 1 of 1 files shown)


e-Evolution:
@yamelsenih Hi Yamel 

This change is not necessary since the connections obtained from the container through the Java Naming and Directory Interface™ (JNDI) protocol for the java:/AdempiereDS or java:/AdempiereSRDS connection pools are configured directly in the container in the following file [postgresql-ds.xml](https://github.com/adempiere/adempiere/blob/13bdd8dbe981300cea8ebbc0174568736d300178/lib/postgresql-ds.xml) , here you can do fine tuning with the parameters that are displayed when you run RUN_setup.sh , from ADempiere 3.9.4 the connections are not managed by ADempiere.

Another important point is that the system can now manage two connection pools, one for writing and one for reading. This is because in a fine tuning the writing parameters can be different from the reading ones.

Now HikariCP and the container manage the connections. It is important to know how each parameter works. Consulting a DBA or reading the HikariCP documentation is important.

[HikariCP NDI-DataSource-Factory](https://github.com/brettwooldridge/HikariCP/wiki/JNDI-DataSource-Factory-(Tomcat,-etc.))

So this change would not make sense. In the case of the connections for the Swing client, they are limited and were defined with the minimum parameters previously handled, since the ADempiere Swing client works as a server client. Each client opens its own connection pools.

kind regards
Victor

piracio:
Actually, will be great to have two type of connections extra to the ones that currently exist.
Or perhaps a flag, to add work_mem changes to the heavy grid process, anything that move massive numbers will be better to have a large work_mem parameters, example:
A database that have, 60G of RAM, put 1G of RAM for a heavy process at session level will be great, as all the sorts will be done in memory. 

Perhaps anything that have over 10K lines with a count can trigger change this parameter.

e-Evolution:
One thing I forgot to mention is that the way to apply the parameters depends on the container, so you should check how it is used for Tomcat, Jetty or WildFly. Here you can see more clearly the one for [Jetty](https://github.com/adempiere/adempiere/blob/develop/lib/jetty-ds.xml) , which is the one I use because it is the fastest and lightest.

Kind regards
Vic

yamelsenih:
Hello, thanks.

This pr just implement a way for remove a hardcode connection DB by parameters.

